PR #50, opened 2016-10-02
Updates to build on Xcode 8/iOS 10.x, test in simulator over wi-fi, add Fabric:

gracion:
@import Quartz on macOS of @QuickLook because QLPreviewItem.h is in QuickLookUI Framework on macOS.
Add new ^ argument to iOS app uses of addMessageHistoryFrom:...
Add en1 Wi-FI definition for testing in simulator. Added warning when missing def results in IP of 0.
Add $SRCROOT-based framework search path for Monal-iOS/External
objc_retainedObject() is no longer supported, replaced with equivalent CFBridgingRelease()
Added now-required Microphone use text to info.plist
Added missing Fabric.framework files


anurodhp:
thanks, my build server was complaining about this :)

PR #75, opened 2017-03-18
Push support (XEP-0357)

tmolitor-stud-tu:
The current implementation compiles and is tested in the iOS simulator, but I don't own a developer account or an iPhone so I could not test the actual push.
Registering for push in the emulator works fine, though.

Currently the pushserver is hardcoded to `https://push.eightysoft.de/v1/register` (http api endpoint for registering the device) and `push.eightysoft.de` for xmpp pushes (IQ stanzas of type "set" as in XEP-0357).

The push server implementation running at push.eightysoft.de can be found here:
https://github.com/tmolitor-stud-tu/mod_push_appserver

If you provide me with a test voip push certificate I can configure the appserver at push.eightysoft.de to correctly do pushes to your test device.

The app settings and push requrements needed to make VOIP push work are explained in detail in this good tutorial (see Prerequisite settings): http://www.nikola-breznjak.com/blog/ios/create-native-ios-app-can-receive-voip-push-notifications/

I think I configured the app to be compiled with the iOS 10.2 SDK, but you should compile it for iOS 9 to enable background voip mode together with voip push (to be on the safe side and deliver a better user experience).

Currently all important log output I added is prefixed with several asterisks to be easily spotted in log output, I'll clean up my code before the final merge as soon as you have tested it (and everything works as expected) :)

### DIFF
--- a/Monal/Classes/MLXMPPManager.h
+++ b/Monal/Classes/MLXMPPManager.h
@@ -208,5 +208,6 @@ Attempts to upload a file to the  HTTP upload service
  */
 -(void) cleanArrayOfConnectedAccounts:(NSMutableArray *) dirtySet;
 
+-(void) setPushNode:(NSString *)node andSecret:(NSString *)secret;
 
 @end
--- a/Monal/Classes/MLXMPPManager.m
+++ b/Monal/Classes/MLXMPPManager.m
@@ -38,6 +38,9 @@ An array of Dics what have timers to make sure everything was sent
 
 @implementation MLXMPPManager
 
+NSString *pushNode;
+NSString *pushSecret;
+
 -(void) defaultSettings
 {
     BOOL setDefaults =[[NSUserDefaults standardUserDefaults] boolForKey:@"SetDefaults"];
@@ -54,6 +57,8 @@ An array of Dics what have timers to make sure everything was sent
         
         [[NSUserDefaults standardUserDefaults] setBool:YES forKey:@"OfflineContact"];
         [[NSUserDefaults standardUserDefaults] setBool:NO forKey:@"SortContacts"];
+        
+        [[NSUserDefaults standardUserDefaults] setObject:[[NSUUID UUID] UUIDString] forKey:@"DeviceUUID"];
         
         [[NSUserDefaults standardUserDefaults] setBool:YES forKey:@"SetDefaults"];
         [[NSUserDefaults standardUserDefaults] synchronize];
@@ -270,6 +275,8 @@ An array of Dics what have timers to make sure everything was sent
     
     xmpp* xmppAccount=[[xmpp alloc] init];
     xmppAccount.explicitLogout=NO;
+    xmppAccount.pushNode=pushNode;
+    xmppAccount.pushSecret=pushSecret;
     
     xmppAccount.username=[account objectForKey:kUsername];
     xmppAccount.domain=[account objectForKey:kDomain];
@@ -814,7 +821,17 @@ withCompletionHandler:(void (^)(BOOL success, NSString *messageId)) completion
     [dirtySet removeObjectsAtIndexes:indexSet];
 }
 
-
+-(void) setPushNode:(NSString *)node andSecret:(NSString *)secret
+{
+    pushNode=node;
+    pushSecret=secret;
+    for(NSDictionary* row in _connectedXMPP)
+    {
+        xmpp* xmppAccount=[row objectForKey:@"xmppAccount"];
+        xmppAccount.pushNode=node;
+        xmppAccount.pushSecret=secret;
+        [xmppAccount enablePush];
+    }
+}
 
 @end
-
--- a/Monal/Classes/MonalAppDelegate.h
+++ b/Monal/Classes/MonalAppDelegate.h
@@ -7,7 +7,9 @@
 //
 
 @import UIKit;
+#if TARGET_OS_IPHONE
 @import PushKit;
+#endif
 #import "Appirater.h"
 #import "DataLayer.h"
 #import "MLTabBarController.h"
@@ -19,14 +21,17 @@
 
 
 
-
+#if TARGET_OS_IPHONE
 @interface MonalAppDelegate : UIResponder <UIApplicationDelegate, UISplitViewControllerDelegate, PKPushRegistryDelegate >
+#else
+@interface MonalAppDelegate : UIResponder <UIApplicationDelegate, UISplitViewControllerDelegate >
+#endif
 
 @property (nonatomic, strong) UIWindow* window;
 @property (nonatomic, strong) UINavigationController* chatNav;
 @property (nonatomic, strong) MLTabBarController* tabBarController;
 @property (nonatomic, strong) UISplitViewController* splitViewController;
-@property (nonatomic, strong)  DDFileLogger *fileLogger;
+@property (nonatomic, strong) DDFileLogger *fileLogger;
 
 -(void) updateUnread;
 
--- a/Monal/Classes/MonalAppDelegate.h
+++ b/Monal/Classes/MonalAppDelegate.h
@@ -6,7 +6,8 @@
 //  Copyright __MyCompanyName__ 2008. All rights reserved.
 //
 
-@import UIKit; 
+@import UIKit;
+@import PushKit;
 #import "Appirater.h"
 #import "DataLayer.h"
 #import "MLTabBarController.h"
@@ -19,7 +20,7 @@
 
 
 
-@interface MonalAppDelegate : UIResponder <UIApplicationDelegate, UISplitViewControllerDelegate >
+@interface MonalAppDelegate : UIResponder <UIApplicationDelegate, UISplitViewControllerDelegate, PKPushRegistryDelegate >
 
 @property (nonatomic, strong) UIWindow* window;
 @property (nonatomic, strong) UINavigationController* chatNav;
--- a/Monal/Classes/MonalAppDelegate.m
+++ b/Monal/Classes/MonalAppDelegate.m
@@ -29,7 +29,6 @@
 @import Fabric;
 #import <DropboxSDK/DropboxSDK.h>
 
-
 //xmpp
 #import "MLXMPPManager.h"
 
@@ -171,6 +170,37 @@ static const int ddLogLevel = LOG_LEVEL_VERBOSE;
 
 }
 
+// Register for VoIP notifications
+-(void) voipRegistration
+{
+    DDLogInfo(@"******************************************************************* registering for voip push...");
+    dispatch_queue_t mainQueue = dispatch_get_main_queue();
+    // Create a push registry object
+    PKPushRegistry * voipRegistry = [[PKPushRegistry alloc] initWithQueue: mainQueue];
+    // Set the registry's delegate to self
+    voipRegistry.delegate = self;
+    // Set the push type to VoIP
+    voipRegistry.desiredPushTypes = [NSSet setWithObject:PKPushTypeVoIP];
+}
+
+// Handle updated push credentials
+-(void)pushRegistry:(PKPushRegistry *)registry didUpdatePushCredentials: (PKPushCredentials *)credentials forType:(NSString *)type
+{
+    //tmolitor: Register VoIP push token (a property of PKPushCredentials) with server
+    DDLogInfo(@"******************************************************************* voip push token: %s", [credentials token]);
+}
+
+-(void)pushRegistry:(PKPushRegistry *)registry didInvalidatePushTokenForType:(NSString *)type
+{
+    DDLogInfo(@"******************************************************************* didInvalidatePushTokenForType called...");
+}
+
+// Handle incoming pushes
+-(void)pushRegistry:(PKPushRegistry *)registry didReceiveIncomingPushWithPayload:(PKPushPayload *)payload forType:(NSString *)type
+{
+    //tmolitor: Process the received push
+    DDLogInfo(@"******************************************************************* incoming voip notfication: %@", [payload dictionaryPayload]);
+}
 
 #pragma mark notification actions
 -(void) showCallScreen:(NSNotification*) userInfo
@@ -257,6 +287,9 @@ static const int ddLogLevel = LOG_LEVEL_VERBOSE;
         UIUserNotificationSettings *settings = [UIUserNotificationSettings settingsForTypes:UIUserNotificationTypeAlert|UIUserNotificationTypeSound|UIUserNotificationTypeBadge categories:categories];
         [[UIApplication sharedApplication] registerUserNotificationSettings:settings];
     }
+    
+    //voip push
+    [self voipRegistration];
     
     [self createRootInterface];
 
--- a/Monal/Classes/MonalAppDelegate.m
+++ b/Monal/Classes/MonalAppDelegate.m
@@ -199,10 +199,13 @@ NSMutableData *pushAPIData;
     NSString *response = [[NSString alloc] initWithData: pushAPIData encoding:NSUTF8StringEncoding];
     DDLogInfo(@"************************ push api returned: %@", response);
     NSArray *responseParts=[response componentsSeparatedByString:@"\n"];
-    if([responseParts count]==2)
-        [[MLXMPPManager sharedInstance] setPushNode:responseParts[0] andSecret:responseParts[1]];
+    if([responseParts[0] isEqualToString:@"OK"] && [responseParts count]==3)
+    {
+        DDLogInfo(@"************************ push api: node='%@', secret='%@'", responseParts[1], responseParts[2]);
+        [[MLXMPPManager sharedInstance] setPushNode:responseParts[1] andSecret:responseParts[2]];
+    }
     else
-        DDLogInfo(@"************************ push api returned data count which is not equal to 2: %d", [responseParts count]);
+        DDLogInfo(@"************************ push api returned invalid data: %@", [responseParts componentsJoinedByString: @" | "]);
 }
 
 // Register for VoIP notifications
@@ -218,29 +221,33 @@ NSMutableData *pushAPIData;
     voipRegistry.desiredPushTypes = [NSSet setWithObject:PKPushTypeVoIP];
     
     //tmolitor: dummy call for iOS simulator
-    PKPushCredentials * credentials = [[PKPushCredentials alloc] init];
-    [self pushRegistry:voipRegistry didUpdatePushCredentials:credentials forType:@"voip"];
+    //PKPushCredentials * credentials = [[PKPushCredentials alloc] init];
+    //[self pushRegistry:voipRegistry didUpdatePushCredentials:credentials forType:@"voip"];
 }
 
 // Handle updated push credentials
 -(void)pushRegistry:(PKPushRegistry *)registry didUpdatePushCredentials: (PKPushCredentials *)credentials forType:(NSString *)type
 {
-    //tmolitor: Register VoIP push token (a property of PKPushCredentials) with server
     DDLogInfo(@"************************ voip push token: %@", credentials.token);
     NSString *token = [[NSString alloc] initWithData:credentials.token encoding:NSUTF8StringEncoding];
-    NSString *device = [[NSUserDefaults standardUserDefaults] stringForKey:@"DeviceUUID"];
+    NSString *node = [[NSUserDefaults standardUserDefaults] stringForKey:@"DeviceUUID"];
+    if(![node length])
+    {
+        [[NSUserDefaults standardUserDefaults] setObject:[[NSUUID UUID] UUIDString] forKey:@"DeviceUUID"];
+        [[NSUserDefaults standardUserDefaults] synchronize];
+    }
     
     //send http api request to register this client-token combination and get back a node-secret combination
     //see http://stackoverflow.com/a/15749527 for help on sending out http requests
     
-    NSString *post = [NSString stringWithFormat:@"device=%@&token=%@", [device stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding],
+    NSString *post = [NSString stringWithFormat:@"type=apns&node=%@&token=%@", [node stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding],
     [token stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding]];
     NSData *postData = [post dataUsingEncoding:NSASCIIStringEncoding allowLossyConversion:YES];
     NSString *postLength = [NSString stringWithFormat:@"%d",[postData length]];
     
     NSMutableURLRequest *request = [[NSMutableURLRequest alloc] init];
-    //this is the hardcoded push api endpoint and should match what is set in XMPPIQ.m
-    [request setURL:[NSURL URLWithString:@"https://push.eightysoft.de/monal/v1/register"]];
+    //this is the hardcoded push api endpoint
+    [request setURL:[NSURL URLWithString:@"https://push.eightysoft.de/v1/register"]];
     [request setHTTPMethod:@"POST"];
     [request setValue:postLength forHTTPHeaderField:@"Content-Length"];
     [request setValue:@"application/x-www-form-urlencoded" forHTTPHeaderField:@"Content-Type"];
@@ -264,7 +271,6 @@ NSMutableData *pushAPIData;
 // Handle incoming pushes
 -(void)pushRegistry:(PKPushRegistry *)registry didReceiveIncomingPushWithPayload:(PKPushPayload *)payload forType:(NSString *)type
 {
-    //tmolitor: Process the received push
     DDLogInfo(@"************************ incoming voip notfication: %@", [payload dictionaryPayload]);
     
     // should any accounts connect?
--- a/Monal/Classes/MonalAppDelegate.m
+++ b/Monal/Classes/MonalAppDelegate.m
@@ -42,6 +42,8 @@ static const int ddLogLevel = LOG_LEVEL_VERBOSE;
 
 @implementation MonalAppDelegate
 
+NSMutableData *pushAPIData;
+
 -(void) createRootInterface
 {
     self.window=[[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
@@ -170,10 +172,43 @@ static const int ddLogLevel = LOG_LEVEL_VERBOSE;
 
 }
 
+#if TARGET_OS_IPHONE
+
+// handle push api incoming response
+-(void) connection:(NSURLConnection *) connection didReceiveResponse:(NSURLResponse *)response
+{
+    [pushAPIData setLength:0];
+}
+
+// handle push api incoming data
+- (void)connection:(NSURLConnection *)connection didReceiveData:(NSData*)data
+{
+    [pushAPIData appendData:data];
+}
+
+// handle push api error
+- (void)connection:(NSURLConnection *)connection didFailWithError:(NSError *)error
+{
+    DDLogInfo(@"************************ push api returned error: %@", error);
+    [pushAPIData setLength:0];
+}
+
+// handle push api response
+- (void)connectionDidFinishLoading:(NSURLConnection *)connection
+{
+    NSString *response = [[NSString alloc] initWithData: pushAPIData encoding:NSUTF8StringEncoding];
+    DDLogInfo(@"************************ push api returned: %@", response);
+    NSArray *responseParts=[response componentsSeparatedByString:@"\n"];
+    if([responseParts count]==2)
+        [[MLXMPPManager sharedInstance] setPushNode:responseParts[0] andSecret:responseParts[1]];
+    else
+        DDLogInfo(@"************************ push api returned data count which is not equal to 2: %d", [responseParts count]);
+}
+
 // Register for VoIP notifications
 -(void) voipRegistration
 {
-    DDLogInfo(@"******************************************************************* registering for voip push...");
+    DDLogInfo(@"************************ registering for voip push...");
     dispatch_queue_t mainQueue = dispatch_get_main_queue();
     // Create a push registry object
     PKPushRegistry * voipRegistry = [[PKPushRegistry alloc] initWithQueue: mainQueue];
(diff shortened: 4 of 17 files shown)


anurodhp:
Very cool. Thanks for doing this and the server component. I'll try to review and merge this  soon.  Im currently very busy  so it might not be as immediate as I would prefer

tmolitor-stud-tu:
Great, let me know if I can help you :)

tmolitor-stud-tu:
Is there anything I can do to help you getting this merged?

I can test it on a real device, if you want to merge this and publish a beta.

anurodhp:
I just set up a prosody server on a vm. I'll get to it this week probably. 

--
Anurodh Pokharel

tmolitor-stud-tu:
Sounds good :)

Let me know when you need any help! :)

anurodhp:
I've pulled your changes into a branch and am making changes to it. Largely things like not breaking the keychain, updating the APIs used etc.   Looking at the code, assuming the pubsub and the push server work, i don't see any reason why this shouldn't work.  There are some UI/UX issues i would like to test though. In particular if we can close the stream and resume with with XEP-0198, which monal does support, then the voip push should be able to work in a variety of scenarios. However, we need to make sure that the user knows how to log out short of deleting the app.  the push should continue to work even if the app is closed, so the usual approach of swiping up to close the app does not stop notifications as you would expect.

anurodhp:
something else you may want to consider, looking at your lua script i think you are using the old APNS end point and not the new one that uses HTTP2
https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html

the new protocol resolves almost all of the issues with the old one. in particular feedback comes back on the same channel  rather than have to query out of band. the old APNS isn't really reliable in providing feedback and unless you regularly poll its hard to know if there was a success or not.

anurodhp:
the APNS part of your code is working ok. I have the module installed prosody and am testing how it works with APNS directly. 
2017-04-17 12:48:42:231 Monal[608:536893] ************************ incoming voip notfication: {
    aps =     {
        alert = "Testing.. (1)";
        badge = 1;
        sound = default;
    };
}

tmolitor-stud-tu:
Very cool :)

I know that there is an HTTP/2 APNS enpoint.
The thing is: I wanted to keep the appserver as simple as possible and using HTTP/2 would add an extra dependency (it seems there is no standard lua library providing HTTP/2).
But maybe I'll add a second APNS module (using the HTTP/2 API) so that user can decide which one to use :)

anurodhp:
Im testing the module right now. I think it is trying to use SSL instead of TLS "sslv3 alert handshake failure. " when connecting to APNS.  I see you have specified tls1.2 but i suspect it is not using it.  Any thoughts?

tmolitor-stud-tu:
That's strange....could you add me to your roster to discuss this further?
tmolitor@datenknoten.me

Did you configure the path to your voip client certificate and key correctly?

tmolitor-stud-tu:
The options are coming from lua-sec: https://github.com/brunoos/luasec/wiki/LuaSec-0.5#ssl_newcontext

Could you try `sslv23` as `protocol` option. This should allow every TLS version.

anurodhp:
i double checked the luasec version i was using. I am on ubuntu yakkety so it .5 and has tls1_2.  ill try that. Added as a contact.

tmolitor-stud-tu:
It could be an issue with the ciphers list, too.
Try the string `HIGH` as value for the config option `push_appserver_apns_ciphers`.

No incoming contact on my end...what's your xmpp id?

anurodhp:
it should be from gmail which is an insecure domain on many servers so its possibly blocked.

tmolitor-stud-tu:
Oh...well...do you know any xmpp server allowing insecure communications with gmail?

tmolitor-stud-tu:
Do you know which TLS version openssl is using?
Could you pass the output of `openssl s_client ...` here?

anurodhp:
this should be the relevant section of the output: 

```
subject=/C=US/ST=California/L=Cupertino/O=Apple Inc./CN=gateway.push.apple.com
issuer=/C=US/O=Entrust, Inc./OU=See www.entrust.net/legal-terms/OU=(c) 2012 Entrust, Inc. - for authorized use only/CN=Entrust Certification Authority - L1K
---
Acceptable client certificate CA names
/C=US/O=Apple Inc./OU=Apple Certification Authority/CN=Apple Root CA
/C=US/O=Apple Inc./OU=Apple Certification Authority/CN=Apple Application Integration Certification Authority
/C=US/O=Apple Inc./OU=Apple Worldwide Developer Relations/CN=Apple Worldwide Developer Relations Certification Authority
/CN=Apple Application Integration 2 Certification Authority/OU=Apple Certification Authority/O=Apple Inc./C=US
Client Certificate Types: RSA sign, ECDSA sign
Requested Signature Algorithms: RSA+SHA512:ECDSA+SHA512:RSA+SHA384:ECDSA+SHA384:RSA+SHA256:ECDSA+SHA256:RSA+SHA1:ECDSA+SHA1
Shared Requested Signature Algorithms: RSA+SHA512:ECDSA+SHA512:RSA+SHA384:ECDSA+SHA384:RSA+SHA256:ECDSA+SHA256:RSA+SHA1:ECDSA+SHA1
---
SSL handshake has read 3364 bytes and written 2546 bytes
---
New, TLSv1/SSLv3, Cipher is AES256-SHA
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : AES256-SHA
```

tmolitor-stud-tu:
The used cipher seems to be `AES256-SHA`, try to add this to the cipher list option.
TLS 1.2 is used, though.

anurodhp:
that did it, odd i thought i saw RSA-AES256-GCM-SHA384 in an earlier output but must have imagined that.

tmolitor-stud-tu:
It's odd enough Apple's server don't allow any GCM ciphers, because Apple is forcing TLS (1.2) for apps and such.
GCM ciphers are more safe than non-GCM variants (especially those using SHA1), because non-GCM variants are more difficult to do right and implementation flaws can cause security holes.

chrisballinger:
I've had confusing issues with that error where the cert simply needed to be regenerated.

anurodhp:
ah i see the problem. We are using the older stuff.  if you connect to 443 
subject=/CN=api.development.push.apple.com/OU=management:idms.group.533599/O=Apple Inc./ST=California/C=US
issuer=/CN=Apple IST CA 2 - G1/OU=Certification Authority/O=Apple Inc./C=US
---
No client certificate CA names sent
Client Certificate Types: RSA sign, ECDSA sign
Requested Signature Algorithms: RSA+SHA512:ECDSA+SHA512:RSA+SHA384:ECDSA+SHA384:RSA+SHA256:ECDSA+SHA256:RSA+SHA1:ECDSA+SHA1
Shared Requested Signature Algorithms: RSA+SHA512:ECDSA+SHA512:RSA+SHA384:ECDSA+SHA384:RSA+SHA256:ECDSA+SHA256:RSA+SHA1:ECDSA+SHA1
Peer signing digest: SHA256
Server Temp Key: ECDH, P-256, 256 bits
---
SSL handshake has read 3103 bytes and written 2330 bytes
---
New, TLSv1/SSLv3, Cipher is ECDHE-RSA-AES256-GCM-SHA384
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : ECDHE-RSA-AES256-GCM-SHA384
    Session-ID: 5AB52A6F344CD06D8F24F6C8A9261CE3570F9ECA25E2BEB1C2D2FA52917BA565

tmolitor-stud-tu:
Ah, that explains it :)

anurodhp:
I've forked your code but do you just want to add the correct cipher in your code?  no need for me to make pull request that way.

tmolitor-stud-tu:
DONE :)

tmolitor-stud-tu:
Regarding the XEP-0198 thing:
I think it would be best, if Monal still tried to keep an active voip connection in the background, but still resume the session via XEP-0198 on incoming pushes, if the app got closed or otherwise disconnected.

That would give us more flexibility, because XEP-0198 resumes have a timeout and keeping the connection open for as long as possible will extend that timeout.

tmolitor-stud-tu:
Keeping the XEP-0198 session active for as long as possible will make sure that MUC scenarios (invites, etc.) behave as expected.

anurodhp:
good call. 
 another bug you might want resolve the path for the setting output is missing /push_appserver/

tmolitor-stud-tu:
> another bug you might want resolve the path for the setting output is missing /push_appserver/
What do you mean exactly?

anurodhp:
the path of the settings call is 
http://192.168.2.3:5280/push_appserver/v1/settings
but all of the links on the page lack 
push_appserver in their path

The device page's Back to List also lacks push_appserver in the path.

tmolitor-stud-tu:
Ah I understand.
The thing is: I used nginx as ssl terminator in front of my prosody http server and removed the `/mod_push_appserver/` prefix this way.

tmolitor-stud-tu:
A side note: Prosody provides HTTPS on port 5281, you should use this instead of the plain HTTP on port 5280.

I'll update the readme and back button :)

anurodhp:
since this was a local machine with an IP, i have ATS and SSL  turned off for testing :D

tmolitor-stud-tu:
okay...sounds reasonable :D

tmolitor-stud-tu:
I made the links to `/v1/settings` relative, they should work now when using prosody directly and when using nginx in front of prosody :)

anurodhp:
Quick question, I've added the module to the prosody folder and the HTTP components are working. However I don't see push listed as a feature: 
`<iq id='6606C4B2-6301-4405-8190-5976E1E4EFE5' type='result' to='london@monal.im/Monal-iOS' from='monal.im'><query xmlns='http://jabber.org/protocol/disco#info'><identity type='pep' name='Prosody' category='pubsub'/><identity type='im' name='Prosody' category='server'/><feature var='urn:xmpp:carbons:2'/><feature var='urn:xmpp:carbons:1'/><feature var='jabber:iq:register'/><feature var='vcard-temp'/><feature var='http://jabber.org/protocol/commands'/><feature var='jabber:iq:roster'/><feature var='jabber:iq:private'/><feature var='urn:xmpp:ping'/><feature var='http://jabber.org/protocol/pubsub#publish'/><feature var='jabber:iq:last'/><feature var='msgoffline'/><feature var='http://jabber.org/protocol/disco#info'/><feature var='http://jabber.org/protocol/disco#items'/><feature var='jabber:iq:version'/><feature var='urn:xmpp:time'/><feature var='jabber:iq:time'/></query></iq> 
`

tmolitor-stud-tu:
My modules are the appserver part of push.
You need mod_cloud_notify (get it here: https://modules.prosody.im/mod_cloud_notify.html) on the server where the user is connected to, to make use of push.

tmolitor-stud-tu:
Of course you can use the same prosody server for both components (mod_cloud_notify **AND** my appserver modules).

anurodhp:
Thanks.  I think i've got it up and running. Things are triggering where i expect them to. There is an error talking to APNS still though. I see a similar error for the feedback channel call. 
```
Apr 17 17:53:14 192.168.2.3:push_appserver	info	Firing event 'incoming-push-to-apns' (node = '9F6B427E-F850-460C-8A80-DE48365F968A', secret = '48a049fa961e2dfe456f28a26155bd7b3f495b17cf88f12407e4070012956e5b')
Apr 17 17:53:14 192.168.2.3:push_appserver_apns	error	Could not receive data from APNS socket: closed
Apr 17 17:53:14 192.168.2.3:push_appserver	error	Push handler for type 'apns' not executed successfully: Error communicating with APNS (receive)
```

chrisballinger:
@anurodhp That will happen if the APNS certificate is invalid. Try generating one of the new-style production+sandbox certs.

anurodhp:
@chrisballinger that is what this is.  It is a a VOIP cert and works for both prod and sandbox. I have used this cert to send APNS pushes from my local  machine already so i know the cert and key  pair itself it fine.

anurodhp:
This might be where I sign off for the day.   Let me know if there is anything i can do to help with debugging the APNS connection.

tmolitor-stud-tu:
oh...uhm...well...interesting :D
I'll see what I can do...

tmolitor-stud-tu:
Does the push message reach the device?

anurodhp:
In my test when i run the push locally it will reach the device.  The push from the prosody module does not.

anurodhp:
yeah the token copied  from  /push_appserver/v1/settings/<deviceid>, when put into an APNS tool will push correctly to the device.

tmolitor-stud-tu:
Can you activate debug output in prosody and send me the complete output when trying to push?

tmolitor-stud-tu:
And also the debug output when the system tries to connect to the feedback server after start.

anurodhp:
> Apr 18 09:17:50 192.168.2.3:http	debug	Serving 'push_appserver' at https://192.168.2.3:5281/push_appserver
Apr 18 09:17:50 192.168.2.3:push_appserver	warn	Debugging is activated, you should turn this off on production servers for security reasons!!!
Apr 18 09:17:50 192.168.2.3:push_appserver	warn	Setting: 'push_appserver_debugging'.
Apr 18 09:17:50 192.168.2.3:push_appserver	info	Appserver started at URL: <https://192.168.2.3:5281/push_appserver/>

> Apr 18 09:17:50 192.168.2.3:push_appserver_apns	info	Appserver APNS submodule loaded
Apr 18 09:17:50 192.168.2.3:push_appserver_apns	info	Connecting to APNS feedback service
Apr 18 09:17:50 192.168.2.3:push_appserver_apns	debug	connecting to feedback.sandbox.push.apple.com port 2196
Apr 18 09:17:50 192.168.2.3:push_appserver_apns	debug	connection established successfully
Apr 18 09:17:50 192.168.2.3:push_appserver_apns	error	Could not receive data from APNS feedback socket (receive 1): closed
Apr 18 09:17:50 hostmanager	debug	Activated host: localhost



And

> Apr 18 09:26:25 c2s5573ba9a5ed0	debug	Destroying session for boston@192.168.2.3/Monal-iOS (boston@192.168.2.3): closed
Apr 18 09:26:25 c2s5573ba9a5ed0	debug	mod_smacks hibernating session for up to 300 seconds
Apr 18 09:26:25 socket	debug	server.lua: closed client handler and removed socket from list
Apr 18 09:26:26 c2s5573baa92fa0	debug	Handled 10 incoming stanzas
Apr 18 09:26:26 c2s5573baa92fa0	debug	Received[c2s]: <message id='5E30BBFD-52EB-456F-B1DE-3DDB74478D48' type='chat' to='boston@192.168.2.3'>
Apr 18 09:26:26 c2s5573ba9a5ed0	debug	Invoking cloud handle_notify_request for smacks queued stanza
Apr 18 09:26:26 192.168.2.3:cloud_notify	debug	Sending push notification for boston@192.168.2.3 to 192.168.2.3 (9F6B427E-F850-460C-8A80-DE48365F968A)
Apr 18 09:26:26 192.168.2.3:push_appserver	debug	node: 9F6B427E-F850-460C-8A80-DE48365F968A, secret: 48a049fa961e2dfe456f28a26155bd7b3f495b17cf88f12407e4070012956e5b
Apr 18 09:26:26 192.168.2.3:push_appserver	info	Firing event 'incoming-push-to-apns' (node = '9F6B427E-F850-460C-8A80-DE48365F968A', secret = '48a049fa961e2dfe456f28a26155bd7b3f495b17cf88f12407e4070012956e5b')
Apr 18 09:26:26 192.168.2.3:push_appserver_apns	debug	Frame data: 0200500100209785D4719C5FCD2130D838E43A05EE7EF1C4678F9EC29F38092077E5E08E88EC02001F7B22617073223A7B22636F6E74656E742D617661696C61626C65223A317D7D030004637FD69805000105
Apr 18 09:26:26 192.168.2.3:push_appserver_apns	debug	Frame ID: 637fd698
Apr 18 09:26:26 192.168.2.3:push_appserver_apns	debug	connecting to gateway.sandbox.push.apple.com port 2195
Apr 18 09:26:26 192.168.2.3:push_appserver_apns	debug	connection established successfully
Apr 18 09:26:26 192.168.2.3:push_appserver_apns	error	Could not receive data from APNS socket: closed
Apr 18 09:26:26 192.168.2.3:push_appserver	error	Push handler for type 'apns' not executed successfully: Error communicating with APNS (receive)
Apr 18 09:26:26 hostmanager	warn	Unhandled response sent to local host 192.168.2.3: <iq id='d86b07af5023fe78d69fac59dcae8280a2bf2401a43f5ce8a76d9abe348eac15' type='error' to='boston@192.168.2.3' from='192.168.2.3'><error type='wait'><internal-server-error xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/><text xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'>Error communicating with APNS (receive)</text></error></iq>

tmolitor-stud-tu:
Strange.
The closed feedback service connection is expected. After transmitting the tokens which should be deleted, the APNS server closes the connection. In our case nothing has to be deleted, so the connection is closed immediately after establishment.

But in the push case a closed connection is strange.
Normally the APNS server should send back an error frame before closing the connection.

The data sent to APNS is this: `0200500100209785D4719C5FCD2130D838E43A05EE7EF1C4678F9EC29F38092077E5E08E88EC02001F7B22617073223A7B22636F6E74656E742D617661696C61626C65223A317D7D030004637FD69805000105` (sent as binary).

The format is documented here: https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/BinaryProviderAPI.html#//apple_ref/doc/uid/TP40008194-CH13-SW1
As far as I can tell, the above hex string adheres to the format described there, but I'm going to double check this.

Does your APNS push tool allow to specify the push protocol (HTTP/2 or the older binary one)?

tmolitor-stud-tu:
What is your configuration? Are you using `silent` pushes (default) or `high` priority pushes?
You should use high priority for VoIP pushes.

What happens if you specify the wrong key for the certificate? I suspect the connection would not be established in that case, but it would be nice if you could double check this.

tmolitor-stud-tu:
I found the error.
The frame length should be a long (4 bytes), but I'm sending a short (2 bytes).

I commited a fix, could you try again? :)

anurodhp:
hmm,  calls to http://192.168.2.3:5280/push_appserver/v1/register no longer work. I get a 404. What should i use as the path now?

tmolitor-stud-tu:
That's strange. The only change I made is this: https://github.com/tmolitor-stud-tu/mod_push_appserver/commits/master/mod_push_appserver

What do you see in the prosody debug output? Could it be that it fails to load the module and you get a backtrace of some sort?

anurodhp:
yup. hmm let me make sure the push module is there. 

`Apr 18 13:34:05 modulemanager	error	Error initializing module 'push_appserver_apns' on 'monal.im': /usr/lib/prosody/core/moduleapi.lua:151: Unable to load required module, mod_push_appserver: /usr/lib/prosody/core/moduleapi.lua:121: Failed to load plugin library 'throttle', error: /usr/lib/prosody/modules/throttle.lib.lua: No such file or directory
stack traceback:
	/usr/lib/prosody/core/modulemanager.lua:29: in function </usr/lib/prosody/core/modulemanager.lua:29>
	[C]: in function 'error'
	/usr/lib/prosody/core/moduleapi.lua:121: in function 'require'
	...dy/modules/mod_push_appserver/mod_push_appserver.lua:18: in main chunk
	(tail call): ?
	[C]: in function 'xpcall'
	/usr/lib/prosody/core/modulemanager.lua:29: in function 'pcall'
	/usr/lib/prosody/core/modulemanager.lua:171: in function 'do_load_module'
	/usr/lib/prosody/core/modulemanager.lua:249: in function 'load'
	/usr/lib/prosody/core/moduleapi.lua:149: in function 'depends'
	.../mod_push_appserver_apns/mod_push_appserver_apns.lua:18: in main chunk
	...
	/usr/lib/prosody/core/modulemanager.lua:171: in function 'do_load_module'
	/usr/lib/prosody/core/modulemanager.lua:249: in function 'load'
	/usr/lib/prosody/core/modulemanager.lua:74: in function '?'
	/usr/lib/prosody/util/events.lua:67: in function 'fire_event'
	/usr/lib/prosody/core/hostmanager.lua:98: in function 'activate'
	/usr/lib/prosody/core/hostmanager.lua:44: in function '?'
	/usr/lib/prosody/util/events.lua:67: in function 'fire_event'
	/usr/lib/prosody/../../bin/prosody:337: in function 'prepare_to_start'
	/usr/lib/prosody/../../bin/prosody:403: in main chunk
	[C]: ?
stack traceback:
	/usr/lib/prosody/core/modulemanager.lua:29: in function </usr/lib/prosody/core/modulemanager.lua:29>
	[C]: in function 'error'
	/usr/lib/prosody/core/moduleapi.lua:151: in function 'depends'
	.../mod_push_appserver_apns/mod_push_appserver_apns.lua:18: in main chunk
	(tail call): ?
	[C]: in function 'xpcall'
	/usr/lib/prosody/core/modulemanager.lua:29: in function 'pcall'
	/usr/lib/prosody/core/modulemanager.lua:171: in function 'do_load_module'
	/usr/lib/prosody/core/modulemanager.lua:249: in function 'load'
	/usr/lib/prosody/core/modulemanager.lua:74: in function '?'
	/usr/lib/prosody/util/events.lua:67: in function 'fire_event'
	/usr/lib/prosody/core/hostmanager.lua:98: in function 'activate'
	/usr/lib/prosody/core/hostmanager.lua:44: in function '?'
	/usr/lib/prosody/util/events.lua:67: in function 'fire_event'
	/usr/lib/prosody/../../bin/prosody:337: in function 'prepare_to_start'
	/usr/lib/prosody/../../bin/prosody:403: in main chunk
	[C]: ?
`

tmolitor-stud-tu:
You need this file
https://github.com/tmolitor-stud-tu/mod_push_appserver/blob/master/mod_push_appserver/throttle.lib.lua
inside the mod_push_appserver directory.

tmolitor-stud-tu:
The best is to clone the entire repository to some directory (`/opt/mod_push_appserver` for example) and point to this directory inside the main prosody conf file like this:

`plugin_paths = { "/opt/mod_push_appserver" };`

If you also clone the `https://hg.prosody.im/prosody-modules/` repo (for `mod_cloud_notify` for example) to `/usr/local/lib/prosody-modules/` the directive could look like this:
`plugin_paths = { "/usr/local/lib/prosody-modules/", "/opt/mod_push_appserver" };`

anurodhp:
yup i accidentally cloned the repo in a sub folder again.  

Good news! its working. The high priority push behavior is also working.  A forced closed app will be reopened an the message retrieved.  As expected, this means we need to add a UI element to allow people to deregister.  I will initially do this on disable. However, that may not always work (if you aren't connected when you disable).  It might need its own UI control per account. 

That said, i do think there might be a bug somewhere. Either in the client or server.  In the following flow:
message 1
message 2
<app closed>
message 3 -> push restarts app 
message 4 

message 3 seems to be lost. I'll see if the server actually provides it.

tmolitor-stud-tu:
You can use prosody's `rawdebug` module to display the raw stanzas going over the wire to debug this further.

Look if the smacks session is resumed properly.
MAM is not implemented in Monal, is that right?

tmolitor-stud-tu:
If you don't want to debug this on your own I can provide you an account on my server to debug this for you :)

anurodhp:
mam0 is supported and looking at the XEP i see that it is now at mam2 :).  Ill take a quick look to make sure the issue isn't client side.

tmolitor-stud-tu:
Side note: APNS push tokens have a limited liefetime, so the following happens for removed accounts:

1. the push token is registered with the accounts server
2. the account is removed in monal
3. pushes from this server still wakeup monal
4. the APNS token times out and gets replaced
5. the pushes trigger an error and eventually se push settings get removed from this server

tmolitor-stud-tu:
Well, I never got MAM working with Monal :(

You have to use prosody 0.10 and the MAM community module to get mam:0 support (the module inside prosody 0.10 only supports mam:2). Or prosody 0.9 which provides mam:0 (but I'm not fully sure about this).

anurodhp:
While in theory push tokens have a limited life time (and also the reason why you should never cache  it locally in userdefaults).  In the last 7 years of using APNS I have not seen it change very often on iOS --OSX is another story.  The life is random but is a pretty long time.

tmolitor-stud-tu:
Oh okay...on Android the token changes pretty often (about once every 6 weeks or so).

tmolitor-stud-tu:
I tried the following to test MAM (with offline messages disabled):
1. Use Monal normally to write some messages
2. Force close the app
3. Send some messages to this account
4. Reopen Monal

I would suspect Monal to query the MAM archive, retrieve the missing messages sent in step 3 and display them as new incoming messages as it does for offline messages.

But this never happened and the messages got lost :(

anurodhp:
I wrote the mam implementation a long time ago. I suspect the spec has changed enough to break it. Something to revisit. 

Regarding the message issue, it  does not seem to be a client related.  I'll test rawdebug

tmolitor-stud-tu:
Regarding the mam thing. Well, all spec changes incremented the namespace version, so your implementation of mam:0 should still work if the server supports it.
Mine does and until recently Conversations used mam:0 with my server just fine, they lately upgraded Conversations to mam:2).

Regarding the lost message: Did you successfully resume the session via smacks? If not this explains why the message is lost (assuming mam is not working, too).

anurodhp:
Yeah, it is not resuming. At least in the circumstance where the app is closed and removed from memory, resuming isn't the right call. I'll have to make sure MAM is working i guess. 

"Apr 18 15:30:38 c2s5580c1109940	debug	Destroying session for providence@192.168.2.3/Monal-iOS (providence@192.168.2.3): Replaced by new connection
Apr 18 15:30:38 c2s5580c1109940	warn	Destroying session with 2 unacked stanzas"

Ill test this out with out the voip socket to see how it handles having the app backgrounded.

tmolitor-stud-tu:
Well, it would be nice if resuming worked even after a force close, because MUCs won't work as expected if the session isn't resumed: muc messages, invites etc. would get lost without a proper resume.

andersruneson:
To be developing instant messaging clients you seem to have missunderstood when to use it :)

tmolitor-stud-tu:
@andersruneson what do you mean?

anurodhp:
@tmolitor-stud-tu  I feel like from a UX perspective closing an application should always terminate any state that existed prior to that.  I assumed resume was for a case where there was a disconnect.

tmolitor-stud-tu:
Well, most (if not every) normal, non tech savy users I know will periodically force close their apps to "clean up". The thing is: if resume doesn't work, MUC won't work either, which degrades user experience a lot imho.

tmolitor-stud-tu:
Btw: I think I have an account which is able to connect to gmail users here: thilo.molitor@juforum.de (if you want you can try to add me there :) )

andersruneson:
@tmolitor-stud-tu I mean you could use a chat system, say xmpp, instead of a GitHub issue ;)

tmolitor-stud-tu:
@andersruneson well, the problem was he couldn't write to my usual xmpp id because gmail doesn't use a secure s2s connection, so we just used this issue instead ;)

chrisballinger:
FWIW I enjoyed this conversation. :)

tmolitor-stud-tu:
nice to hear :)

I'll help you implement voip pushes, too, as soon as our webrtc implementation 
is ready :)

maufl:
Should this work without running my own push_appserver? Would the server at https://push.eightysoft.de be used by default? I have mod_cloud_notify running on my prosody server and with Conversations on Android it just works (using push.siacs.eu) but it does not seem to work with Monal on my parents iPhones. :(

tmolitor-stud-tu:
No, the app developer uploading it to the Apple appstore has to host the push sevice, push.eightysoft.de was only for demonstration while I developed the push server and added push support to monal.

Note: I'm NOT the app developer, I don't possess the push certificate issued for Monal by Apple and therefore cannot upload a new version of Monal to iTunes nor can I operate a push server for it.

maufl:
Ok, and is there a push server running for Monal? I assume not because it seems that while the support is merged, push does not work with Monal. Does that mean that we will have to wait for @anurodhp to host a push server?

tmolitor-stud-tu:
@maufl Yes exactly. @anurodhp has to host a push server first.
The current iTunes version of Monal doesn't include this pull request, so a new version including this pull request has to be published as well.

anurodhp:
Yeah, I'm trying to figure out a way to do this so that it reliable and actually will scale up. It slightly changes the scope of what Monal is because for the past 9 years, it was just a client app. Once I add a push server, I have sever infrastructure to maintain.  If this server isn't reliable and secure, its useless.

maufl:
I understand, and I hate apple and Google for doing this to their smartphone OSs. Unfortunately my parents only use Jabber with me and therefore rarely open Monal, so they only receive messages from me when they message me first. Maybe you could use the push server of the chat secure developer? They should have the same problem.

tmolitor-stud-tu:
@maufl It's not so much about the choice of software. I wrote an appserver prefectly suitable for Monal (in fact I wrote it for monal) here: https://github.com/tmolitor-stud-tu/mod_push_appserver

It's about server management, maintenance and infrastructural choices (hardware, operating system etc.).
And about money, of course.

anurodhp:
Yeah, Im planning on a server at AWS.  Im just very tied up with personal and work related stuff at the moment

maufl:
I meant that maybe you could use the same deployed push server as chat secure? So that you would not need to host your own. Or is that impossible because of the way the certificates are tied to the app and the server?

tmolitor-stud-tu:
@maufl That would mean @anurodhp would have to give his private key for app developement to @chrisballinger 
I don't think he will do that.

maufl:
Aw :/ OK, let me know if I can help financially.

PR #77, opened 2017-04-13
SENT Carboncopy is fixed.

pwnartist:
Issue : 
  - SENT Carboncopy messages were reflected into other resources, immediately.

Solution :
  - This is why userDic used recipient for "to", which means recipient's resource of the forwarded message.
  - However, It differ from the actuallyTo when SENT Carboncopy.
  - Therefore, we need to use actuallyTo for showing sent message from other resources of the bare JID, immediately.


anurodhp:
thanks,  the code as it is will crash with nil insertion. i've merged and made the change

PR #80, opened 2017-04-24
Fix smacks bugs and make session persistent across force-close

tmolitor-stud-tu:
**The only two bugs left to fix are:**

1. The "Logging in" toast not disappearing
2. The presence state of all contacts is lost after a smacks resume without force-close in between.
    _If the app was force-closed in between, the presence is recalculated by issuing presence probe stanzas, so the bug doesn't occur in the force-close scenario._

Just ask me if you need help in fixing bug no. 2, point me in the right direction and I'll solve it for you :)


tmolitor-stud-tu:
**Why is that happening???**

screenshot24

tmolitor-stud-tu:
Bug 2 should be fixable by moving the logic, which puts all contacts into "offline" state when the connection is lost, to the case when the smacks resume failes.

This way the contacts will retain their online state when the session is resumed via smacks.
I don't know where in the code the contacts are put into "offline" state and therefore couldn't fix this bug all by myself.

PR #139, opened 2018-12-02
Corrected number of cells in tables in Display Settings

mimi89999:


Echolon:
@mimi89999 Thank you for your commit that time - we highly appreciate commits if you have interest coming back to help out! :+1:

PR #366, opened 2020-03-03
Some more fixes

tmolitor-stud-tu:
Because I enabled the thread sync on some of the operation blocks the deadlock, you encountered could surface again.
Thus this is not ready for a release but for a beta release to test and fix the deadlock.

### DIFF
--- a/Monal/Classes/MLXMPPManager.m
+++ b/Monal/Classes/MLXMPPManager.m
@@ -57,7 +57,7 @@ An array of Dics what have timers to make sure everything was sent
     if(!setDefaults)
     {
         //  [[NSUserDefaults standardUserDefaults] setObject:@"" forKey:@"StatusMessage"]; // we dont want anything set
-        [[NSUserDefaults standardUserDefaults] setObject:@"50" forKey:@"XMPPPriority"];
+        [[NSUserDefaults standardUserDefaults] setObject:@"0" forKey:@"XMPPPriority"];
         [[NSUserDefaults standardUserDefaults] setBool:NO forKey:@"Away"];
         [[NSUserDefaults standardUserDefaults] setBool:YES forKey:@"Visible"];
         [[NSUserDefaults standardUserDefaults] setBool:YES forKey:@"MusicStatus"];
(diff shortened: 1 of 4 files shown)


Echolon:
Thank you! Your efforts are highly appreciated! Lets wait for Anu to review!

Monal/Classes/xmpp.m
```diff
@@ -1778,7 +1769,7 @@ -(void) processInput
                             NSMutableDictionary *settings = [[NSMutableDictionary alloc] init];
                             [settings setObject:self.connectionProperties.identity.domain forKey:kCFStreamSSLPeerName];
                             
-                            [settings setObject:kCFStreamSocketSecurityLevelTLSv1 forKey:kCFStreamSSLLevel];
+                            [settings setObject:kCFStreamSocketSecurityLevelNegotiatedSSL forKey:kCFStreamSSLLevel];
```
anurodhp: Thanks, i had hardcoded TLS1 to block SSL3 but thats blocked by apple now so this change should use pls 1 and up

Monal/Classes/MLXMPPManager.m
```diff
@@ -57,7 +57,7 @@ -(void) defaultSettings
     if(!setDefaults)
     {
         //  [[NSUserDefaults standardUserDefaults] setObject:@"" forKey:@"StatusMessage"]; // we dont want anything set
-        [[NSUserDefaults standardUserDefaults] setObject:@"50" forKey:@"XMPPPriority"];
+        [[NSUserDefaults standardUserDefaults] setObject:@"0" forKey:@"XMPPPriority"];
```
anurodhp: is there a reason to set this to 0. A common problem i have had is adium users on the desktop with a higher priority getting the message instead.  This causes people to think that monal is defective.

Monal/Classes/ParseMessage.h
```diff
@@ -24,7 +24,9 @@
 @property (nonatomic, copy, readonly) NSString* messageText;
 @property (nonatomic, copy, readonly) NSString* messagHTML;
 @property (nonatomic, copy, readonly) NSString* subject;
-@property (nonatomic, assign, readonly) BOOL hasBody; 
+@property (nonatomic, assign, readonly) BOOL hasBody;
+@property (nonatomic, assign, readonly) BOOL isMUC;
```
anurodhp: rather than bools, set a var for the message type with the constant, and later at the bottom for the receipt just echo back the same type without any if statement

Monal/Classes/MLXMPPManager.m
tmolitor-stud-tu: yes, a priority of zero means that messages sent to the bare jid get delivered to every client.
And: monal supports carbons --> even when adium has a higher priority that should not matter because monal will still get the carbons.

See also this: https://docs.modernxmpp.org/client/design/#deprecated-options

Monal/Classes/ParseMessage.h
tmolitor-stud-tu: yes, good catch, fixed in my latest commit :)

Monal/Classes/MLXMPPManager.m
tmolitor-stud-tu: and see this explanation:
https://docs.modernxmpp.org/client/protocol/#multi-device

PR #371, opened 2020-03-10
Refactoring of low-level networking code

tmolitor-stud-tu:
A full refactoring of low-level networking code.

This refactor introduces two operation queues for networking: an incoming queue which processes all incoming stanzas and an outgoing queue which is responsible for serializing a stanza to bytes and pushing them into the tcp socket.
This even buffers low level writes to the tcp socket if it can not consume a whole stanza in bytes.

### DIFF
--- a/Monal/Classes/xmpp.m
+++ b/Monal/Classes/xmpp.m
@@ -1113,7 +1113,29 @@ NSString *const kXMPPPresence = @"presence";
 		[self send:(MLXMLNode*)[dic objectForKey:kStanza]];
 	}];
 	[self persistState];
-    
+}
+
+-(void) resendUnackedMessageStanzasOnly:(NSMutableArray*) stanzas
+{
+    /**
+     Send appends to the unacked stanzas. Not removing it now will create an infinite loop.
+     It may also result in mutation on iteration
+	*/
+	if(stanzas)
+	{
+		NSMutableArray *sendCopy = [[NSMutableArray alloc] initWithArray:stanzas];
+		//clear queue because we don't want to repeat resending these stanzas later if the var stanzas points to self.unAckedStanzas here
+		[stanzas removeAllObjects];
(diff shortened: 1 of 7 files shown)


Echolon:
Big thanks for your effort and commits!

tmolitor-stud-tu:
I also implemented the switch from userdefaults storage to sqlite storage for stream state which should fix the smacks counting bugs we never really could say why they occured.

anurodhp:
a couple of things, this branch doesnt build. Ill try to get the travis status to show up here. 
The other is please squash it all into a single commit as that makes it easier to review.

Monal/Classes/xmpp.m
```diff
@@ -1080,78 +1100,73 @@ -(NSMutableDictionary*) nextStanza
...
--(void) requestSMAck {
+-(void) requestSMAck:(BOOL) queuedSend
+{
     if(!self.smacksRequestInFlight && self.unAckedStanzas.count>0 ) {
         DDLogVerbose(@"requesting smacks ack...");
         MLXMLNode* rNode =[[MLXMLNode alloc] initWithElement:@"r"];
-        NSDictionary *dic=@{kXMLNS:@"urn:xmpp:sm:3"};
-        rNode.attributes =[dic mutableCopy];
-        [self send:rNode];
+        NSDictionary *dic=@{kXMLNS:@"urn:xmpp:sm:3", @"debugh":[NSString stringWithFormat:@"%@", self.unAckedStanzas.count]};
```
anurodhp: what is this debug attribute? is it temporary

anurodhp:
would it be possible to point this to request a merge into the refactor branch. I can run some tests there and deploy a beta from that.

Monal/Classes/xmpp.m
tmolitor-stud-tu: yes it is temporary...I just want to see monals internals in the server logs so that I am able to compare the values.

tmolitor-stud-tu:
I can try to make a pull request against the refactor branch, yes :)

anurodhp:
closing as it's already merged

PR #382, opened 2020-03-26
Basic message draft support

FriedrichAltheide:
Adding basic support for message drafts

- Adding new column messageDraft to buddylist
- Saving message drafts when exiting chatViewController
- Load last message draft when returning to chat


anurodhp:
Thanks for doing this. Could you switch  your new db methods to use a completion handler?

something like this 
 [self executeScalar:query andArguments:params withCompletion:^(NSObject *name) {}]

I am trying to remove the older methods or reduce their usage as much as possible since calling it on the main thread will  block

Echolon:
@FriedrichAltheide Thank you very much for your pull request!

I just saw that you are new to Github - if you are happy to help out amd look for a project, your commits and work is very appreciated here!

Thank you!

PR #385, opened 2020-03-29
Fix #206 (Implement SRV lookup)

### LINKED ISSUE #206: Implement SRV lookup and use correct ALPN value in TLS
Please implement SRV DNS-record lookup (see here: https://wiki.xmpp.org/web/SRV_Records)

This way it is possible to provide an IP/port combination different to the A-record of the xmpp domain.
Even multiple SRV records with different priority are possible and should be tried in order by clients.

This allows server operators to even use port 443 (https) as fallback for environments where only http/https is allowed and all other ports are filtered out.

To multiplex https and xmpp (and possibly others) on port 443 a correct ALPN value of `xmpp-client` and support of `_xmpps-client` SRV records is needed.

One service provider (apart from myself) using port 443 multiplexing based on ALPN values and SRV records is conversations.im

tmolitor-stud-tu:
This fixes #206 and additionally fixes some bugs in the old SRV implementation (no retry with records having lower priority and not using the port value of the SRV record).

### DIFF
--- a/Monal/Classes/xmpp.m
+++ b/Monal/Classes/xmpp.m
@@ -8,6 +8,8 @@
 
 #import <CommonCrypto/CommonCrypto.h>
 #import <CFNetwork/CFSocketStream.h>
+#import <Security/SecureTransport.h>
+
 #import "xmpp.h"
 #import "DataLayer.h"
 #import "EncodingTools.h"
@@ -156,7 +158,9 @@ NSString *const kXMPPPresence = @"presence";
 {
     _accountState = kStateLoggedOut;
     
-    _discoveredServerList=[[NSMutableArray alloc] init];
+    _discoveredServersList=[[NSMutableArray alloc] init];
+	if(!_usableServersList)
+		_usableServersList=[[NSMutableArray alloc] init];
     _inputDataBuffer=[[NSMutableData alloc] init];
     _inputBuffer=[[NSMutableString alloc] init];
     _outputQueue=[[NSMutableArray alloc] init];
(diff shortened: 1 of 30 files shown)


Monal/Classes/xmpp.m
```diff
@@ -242,14 +252,42 @@ -(void) setRunLoop
...
+		DDLogInfo(@"Set TLS properties on streams. Security level %@", [self->_iStream propertyForKey:NSStreamSocketSecurityLevelKey]);
+	else
+	{
+		DDLogError(@"not sure.. Could not confirm Set TLS properties on streams.");
+		DDLogInfo(@"Set TLS properties on streams.security level %@", [self->_iStream propertyForKey:NSStreamSocketSecurityLevelKey]);
+	}
+	
+	//see this for extracting the sslcontext of the cfstream: https://stackoverflow.com/a/26726525/3528174
+	//the ssl context methods (like SSLSetALPNData) are declared in Security/SecureTransportPriv.h and can be found here:
+	//https://opensource.apple.com/source/Security/Security-57740.31.2/OSX/libsecurity_ssl/lib/sslContext.c.auto.html
```
anurodhp, line 277: I think this is a private API and not actually allowed in the app store additionally, It may work but the reason it is private is that there is no guarantee that the interface will not change without warning .

anurodhp:
It may be beneficial to put your changes in different branches. There is a lot that i can merge here but the possible private api use means the app cant but put in the app store.

Monal/Classes/xmpp.m
tmolitor-stud-tu: okay, I'll change the code to use the newer network framework...

tmolitor-stud-tu:
no need to break out the changes, I'll just fix the api usage :)

Monal/Classes/xmpp.m
tmolitor-stud-tu: solved now :)

PR #386, opened 2020-03-29
Show messageDrafts in ActiveChatsViewController

FriedrichAltheide:
Display message drafts in ActiveChatsViewController

### DIFF
--- a/Monal/Classes/MLContactCell.m
+++ b/Monal/Classes/MLContactCell.m
@@ -52,17 +52,30 @@
 {
     if(![self.statusText.text isEqualToString:text]) {
         self.statusText.text=text;
-        if(text)
-        {
-            self.centeredDisplayName.hidden=YES;
-            self.displayName.hidden=NO;
-            self.statusText.hidden=NO;
-        }
-        else {
-            self.centeredDisplayName.hidden=NO;
-            self.displayName.hidden=YES;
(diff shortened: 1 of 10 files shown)


tmolitor-stud-tu:
Thanks for your contribution!
It's nice to know that I'm no longer the only one making contributions :)

If you want, you can join our channel at monal@chat.yax.im or add me to your roster: thilo.molitor@juforum.de

BTW: you are from germany, too?

Monal/Classes/MLContactCell.m
```diff
@@ -52,17 +52,30 @@ -(void) showStatusText:(NSString *) text
...
-            self.displayName.hidden=YES;
-            self.statusText.hidden=YES;
-        }
+        [self setStatusTextLayout:text];
+    }
+}
+
+-(void) showStatusTextItalic:(NSString *) text withItalicRange:(NSRange)italicRange
+{
+    UIFont* italicFont = [UIFont italicSystemFontOfSize:self.statusText.font.pointSize];
```
anurodhp, line 61: this is going to get called continuously even on scrolling. It may be better to not recreate it all the time.

FriedrichAltheide: Now attributedText will be updated only when text and the italic range changed

Issue #370, opened 2020-03-08
Password Change doesn't work with restricted setting in ejabberd 20.01.


sreutter:
With restricted access to the mod_register module in ejabber (no registration, but password change possible)

( mod_register:
    access: none)

the password change isn't possible from within monal.im on ios. I tested it with gajim, it works there.

PR #388, opened 2020-04-04
Fix change password #370

FriedrichAltheide:
Send username instead of jid when changing password

Fixing #370


Echolon:
Thank you for you work! Highly appreciated. We will wait for the feedback from Anu

Issue #316, opened 2019-12-26
[Feature] Show location and interpret geo-location links


Echolon:
Hi,

not much to say - would be great to show (and send!) locations with OSM ressources.

Here is an expample: geo:51.17934297928927,10.4150390625 as Conversations sends it

Cheers

Echolon:
7eaa75c 

ca0207e

Thank you @FriedrichAltheide

PR #390, opened 2020-04-05
display geo location

FriedrichAltheide:
- Added new option to enable inline geolocations
- Displays inbound and outbound strings in form of geo:latitudeRange,longitude
- Messages are only parsed as geolocations when they have a „geo:" prefix and the inline option is enabled
- After disabling the inline option the geo locations are displayed as a underlined text. The message is linked to openstreetmaps

Ref #316

### DIFF
--- a/Monal/Classes/ActiveChatsViewController.m
+++ b/Monal/Classes/ActiveChatsViewController.m
@@ -382,6 +382,9 @@
                 } else if([messageRow.messageType isEqualToString:kMessageTypeImage])
                 {
                     [cell showStatusText:@"📷 An Image"];
+                } else if([messageRow.messageType isEqualToString:kMessageTypeGeo])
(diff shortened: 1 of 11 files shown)


Echolon:
@FriedrichAltheide 
Wow, Friedrich - I personally love you for that PR! THANK YOU SO MUCH! :heart:

FriedrichAltheide:
@Echolon Lets see how the review goes. Xcode changed quite a lot of the storyboard while adding the new UI elements. Sending geoLocations via a button is currently in testing.

Echolon:
@FriedrichAltheide Sure, take your time. @anurodhp is busy with the COVID-19 situation at the moment. So we need to be a bit patient :/

Still its great work you do so, feel free to join our MUC as well.

PR #393, opened 2020-04-13
Refactored table account

FriedrichAltheide:
- Fixing roster version saving into table account
- Setting all rosterVersion entries to 0 if null
- rosterVersion is now 0 by default
- Removed old column password
- Removed column account_name
- Increased length of username, server, resource, domain


anurodhp:
thanks this was a decade worth of tech debt that ive put off doing :)

Echolon:
@FriedrichAltheide have you seen there are conflicts in one file? :(

PR #396, opened 2020-04-17
Allow overriding host and port in advanced settings, fixes #242

### LINKED ISSUE #242: Advanced Server Setting ignored (SSH Forward Tunnel, SRV)
Hi!

When trying to tunnel through a SSH forward tunnel, Monal is unable to establish a connection to the server. It tries to connect to the server specified in DNS SRV RR, but not to the server configured in Advanced setting for the server address, e.g. 127.0.0.1 port 5222, which forwards this to the real XMPP server. 

However, other XMPP clients such as Psi+ or Jitsi do work that way, so I assume this should also work with Monal... :-)

tmolitor-stud-tu:
Again the title says it all :)


tmolitor-stud-tu:
@anurodhp Please merge my pull request using my idividual commits instead of squashing it all into one, that makes it really hard to manage my fork.

Maybe adding a placeholder text saying "Overide host/ip" to the server field would be nice, but I could not find where the textfield is defined / how to add the placeholder text.

PR #398, opened 2020-04-22
Fix connecting to servers without SRV records and display a warning for "." SRV records, fix some dns resolver problems

tmolitor-stud-tu:
This does exactly what the title says and solves some errors in the resolver that sometimes prevented proper dns resolution.


Echolon:
Thank you guys!

PR #404, opened 2020-05-03
Fix parser to be streaming and some other connection errors

tmolitor-stud-tu:
This fixes some errors that made the current beta nearly unusable.

The new `MLPipe` Class makes it possible to detach the `NSInputStream` from the `NSXMLParser` and to attach it to a new `NSXMLParser`.
The old `NSXMLParser` and its parse method will even be correctly stopped and not cause CPU cycles or memory consumption.

Many thanks to @FriedrichAltheide for assisting and testing this!!

### DIFF
--- a/Monal/Classes/xmpp.m
+++ b/Monal/Classes/xmpp.m
@@ -156,6 +156,7 @@ NSString *const kXMPPPresence = @"presence";
 {
     _accountState = kStateLoggedOut;
 
+    _SRVDiscoveryDone = NO;
     _discoveredServersList=[[NSMutableArray alloc] init];
     if(!_usableServersList)
 		    _usableServersList=[[NSMutableArray alloc] init];
@@ -336,7 +337,7 @@ NSString *const kXMPPPresence = @"presence";
 
 -(void) connectionTask
 {
-    //allow override for server and port if one is specified for the account
+    // allow override for server and port if one is specified for the account
     if(![self.connectionProperties.server.host isEqual:@""])
     {
         DDLogInfo(@"Ignoring SRV records for this connection, server manually configured: %@:%@", self.connectionProperties.server.connectServer, self.connectionProperties.server.connectPort);
@@ -345,8 +346,17 @@ NSString *const kXMPPPresence = @"presence";
     }
 
     // do DNS discovery if it hasn't already been set
-    if([_discoveredServersList count]==0) {
+    if(!_SRVDiscoveryDone) {
+        _SRVDiscoveryDone = YES;
         _discoveredServersList=[[[MLDNSLookup alloc] init] dnsDiscoverOnDomain:self.connectionProperties.identity.domain];
+        
+        // no SRV records found, update server to directly connect to specified domain
+        if([_discoveredServersList count]==0)
+        {
+            [self.connectionProperties.server updateConnectServer: self.connectionProperties.identity.domain];
+            [self.connectionProperties.server updateConnectPort: @5222];
+            [self.connectionProperties.server updateConnectTLS: NO];
+        }
     }
 
     // Show warning when xmpp-client srv entry prohibits connections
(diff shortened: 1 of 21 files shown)


tmolitor-stud-tu:
A sync call does exactly the same as your previous code that added the omemo calls to the thread/queue omemoResult() was initially called from.
The only difference is that it does not reorder code flow on this thread/queue.

If omemoResult is called from the main queue, then it will process omemo on the main queue, but in this case your previous code does exactly the same and processes omemo in the main queue, too.

anurodhp:
Very cool
With the parser. I’ll try to look and test it. When I tried something similar, I had an issue with callbacks to the delegate after it was destroyed that caused it to crash. Before I dig in,  Have you tested this with startls?

tmolitor-stud-tu:
Your solution was:
thread X -->       call omemoResult --\    |process some other code |     /-- run omemo calls  
main thread -->                                    \-- do applicationState check --/


My solution is:
thread X -->       call omemoResult --\        |wait for main queue           /-- run omemo calls  
main thread -->                                    \-- do applicationState check --/

tmolitor-stud-tu:
Yes, @FriedrichAltheide and I have tested it with and without starttls in my simulator and in his simulator...no issues :)

I updated the ascii art of my last answer to be more exact btw.

tmolitor-stud-tu:
> So only the if() need to be on the main queue. Try this
> Book check =no
> 
> Synch main queue {
> If (uibackground check) {
> Check = yes;
> }
> }
> 
> If(!check). {
> Expensive onmem code
> }

Oh yes, my bad!
I guess the many preprocessor directives confused me somehow :D 
I changed my commit and force pushed it.

Something else that preventsthis from getting merged?

tmolitor-stud-tu:
I did some further cleanup of the code and force pushed a new version.

This should be the final one :)

Monal/Classes/xmpp.m
```diff
@@ -1549,9 +1549,6 @@ -(void) processInput:(XMPPParser *) parsedStanza
             self.loginCompletion(YES, @"");
             self.loginCompletion=nil;
         }
-        
-        [self queryMAMSinceLastMessageDate];
```
tmolitor-stud-tu, line 1589: This is not needed because we resumed the stream and everything that happened since our last connection is still queued on the server and will be streamed to us right now.

Monal/Classes/xmpp.m
```diff
@@ -1577,16 +1574,13 @@ -(void) processInput:(XMPPParser *) parsedStanza
                 self.loginCompletion(YES, @"");
                 self.loginCompletion=nil;
             }
-            
-            [self queryMAMSinceLastMessageDate];
```
tmolitor-stud-tu, line 1617: This is not needed because MAM will already be queried in initializeSession after resource bind (directly when the bind iq started in line 1574 returns).

In fact this is dangerous, because we pipeline xml stanzas directly after sending out our bind iq and if the bind fails this pipelined stanzas will produce a stream error.

Monal/Classes/xmpp.m
```diff
@@ -1577,16 +1574,13 @@ -(void) processInput:(XMPPParser *) parsedStanza
...
         }
         else        //smacks enable failed
         {
             self.connectionProperties.supportsSM3=NO;
 
             //init session and query disco, roster etc.
             [self initSession];
-            [self queryMAMSinceLastMessageDate];
```
tmolitor-stud-tu, line 1625: This is not needed because the mam query is already done in initSession called directly in the line above.

Monal/Classes/xmpp.m
```diff
@@ -2044,9 +2038,6 @@ -(void) readState
         {
             NSNumber *mamNumber = [dic objectForKey:@"supportsMAM"];
             self.connectionProperties.supportsMam2 = mamNumber.boolValue;
-            if(self.accountState>=kStateLoggedIn) {
-                [self queryMAMSinceLastMessageDate];
```
tmolitor-stud-tu, line 2132: This will duplicate mam queries and is generally unneeded because initSession will already do mam queries.

Monal/Classes/xmpp.m
```diff
@@ -2223,7 +2214,11 @@ -(void) initSession
     [self fetchRoster];
     [self queryDisco];
     [self sendInitalPresence];
-
+    
+    //query mam since last received message because we could not resume the smacks session
+    //(we would not have landed here if we were able to resume the smacks session)
+    //this will do a catchup of everything we might have missed since our last connection
+    [self queryMAMSinceLastMessageDate];
```
tmolitor-stud-tu, line 2221: This is the single needed code position where a mam query should be issued, see the other comments.

tmolitor-stud-tu:
I commented every change of the mam query commit to make clear why I changed it.

tmolitor-stud-tu:
@anurodhp this is ready to be merged :)

PR #410, opened 2020-06-12
PR01

tmolitor-stud-tu:
@FriedrichAltheide  and I did some work, see the commit messages for details.

Every commit represents a logical entity that should be easy to revert, if necessary.

### DIFF
--- a/Monal/Classes/MLXMPPManager.m
+++ b/Monal/Classes/MLXMPPManager.m
@@ -12,14 +12,10 @@
 #import "MLMessageProcessor.h"
 #import "ParseMessage.h"
 
-#ifndef TARGET_IS_EXTENSION
 #if TARGET_OS_IPHONE
 #import "MonalAppDelegate.h"
-@import MobileCoreServices;
 #endif
-#else
 @import MobileCoreServices;
-#endif
 
 @import SAMKeychain;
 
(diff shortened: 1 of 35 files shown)


tmolitor-stud-tu:
I rebased it upon your latest changes and added a fix for the smacks loop and counting bug

Monal/Classes/MLXMPPManager.m
```diff
@@ -12,14 +12,10 @@
 #import "MLMessageProcessor.h"
 #import "ParseMessage.h"
 
-#ifndef TARGET_IS_EXTENSION
```
anurodhp, line 15: this is intentional. it is so that the app delegate is not included in the extension that loads from APNS push , share etc. It should be restored.

tmolitor-stud-tu: Oh sorry, I somehow oversimplyfied this.
You're right of course, fixed this :)

tmolitor-stud-tu: Short question: why should the appdelegate not be included in the extension? What is the reasoning behind this?
Note: I don't know how these extensions and the appdelegate work.

anurodhp: each app delegate is for a different app.  The extension is a full fledged separate app with a different ui and a different delegate

PR #412, opened 2020-06-22
Pr02

tmolitor-stud-tu:
See individual commits

### DIFF
--- a/Monal/Classes/MLAccountPickerViewController.m
+++ b/Monal/Classes/MLAccountPickerViewController.m
@@ -32,10 +32,8 @@
 
 - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath {
     UITableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:@"AccountCell" forIndexPath:indexPath];
-    NSDictionary *row=[MLXMPPManager sharedInstance].connectedXMPP[indexPath.row];
-    xmpp* xmppAccount=[row objectForKey:@"xmppAccount"];
+    xmpp* xmppAccount = [MLXMPPManager sharedInstance].connectedXMPP[indexPath.row];
     cell.textLabel.text=xmppAccount.connectionProperties.identity.jid;
-    
     return cell;
 }
 
--- a/Monal/Classes/MLComposeViewController.m
+++ b/Monal/Classes/MLComposeViewController.m
@@ -56,8 +56,7 @@
             xmpp* account;
             
             if(_selectedRow<[[MLXMPPManager sharedInstance].connectedXMPP count] && _selectedRow>=0) {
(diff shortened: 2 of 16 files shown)


tmolitor-stud-tu:
we are seeing serious bugs in the current alpha, please don't merge!

tmolitor-stud-tu:
okay, the bug seems to be in the catalyst version of the current beta, not the current alpha...I'm still investigating...

anurodhp:
What is the issue? For the most part the two behave the same but there are subtle differences esp with exerts in iOS versions

tmolitor-stud-tu:
it seems like monal does not receive/process any stanzas after the <resumed> element and drops the connection a few seconds after receiving the <resumed> element.
the last stanza monal is sending is the push enable iq (thats te send: call from the receiveQueue operation that processes the <resumed> element.

it seems like there is no other incoming stanza processed on the receive queue after this.
on every resume monal reports the same stanza count.
fun fact:
together with a small monal bug fixed in https://github.com/tmolitor-stud-tu/Monal/commit/ab9720e72d6d78531190ffbb9170377a7a14e70b this triggered a bug in prosody's mod_csi_battery_saver which duplicated all stanzas in the smacks queue and after a few resumes friedrich's account on my server had ~2000000 outgoing stanzas in the smacks queue and my server froze.

tmolitor-stud-tu:
see my updated post (don't know if the mail notification of github includes post edits)

tmolitor-stud-tu:
I don't know if this is reproducible/how this would be reproducible. I'm working with friedrich to get more insights on this front.

tmolitor-stud-tu:
Could you enable the `#DEBUG` define in the beta builds? its hard to identify bugs without log output :(

tmolitor-stud-tu:
do you have any idea why building the catalyst version on travis ci throws the following error?

```
/Users/travis/build/tmolitor-stud-tu/monal.alpha/Monal/Classes/AESGcm.m:11:9: fatal error: 'MLCrypto/MLCrypto-Swift.h' file not found

#import <MLCrypto/MLCrypto-Swift.h>

        ^~~~~~~~~~~~~~~~~~~~~~~~~~~

1 error generated.
```

FriedrichAltheide:
> I don't know if this is reproducible/how this would be reproducible. I'm working with friedrich to get more insights on this front.

I think the MR can be merged as the bug was seen in the current beta release. `#DEBUG` would be create.

tmolitor-stud-tu:
can you merge this? current beta has a bug in the logview which is solved by commit https://github.com/tmolitor-stud-tu/Monal/commit/2e4d9040f7826a5954eb7872f191086e023a2fef

(and the bug that caused me to stop you from merging is already in the beta)

tmolitor-stud-tu:
You could also cherry-pick the https://github.com/tmolitor-stud-tu/Monal/commit/2e4d9040f7826a5954eb7872f191086e023a2fef commit.

I managed to build the catalyst version, but I don't know how to pack the resulting .xcarchive into an .app file :(

PR #415, opened 2020-07-13
Update XEPsupport.md

Echolon:
Add XEP-191 information. Version missing


Echolon:
fixed via #416 
3c08dd9

PR #417, opened 2020-07-18
PR03

tmolitor-stud-tu:
See the commit messages for a (sometimes **really**) detailed description of our changes.


tmolitor-stud-tu:
Additional remark: I already told you that the new chatview header  doesn't look like it should.
There seem to be some contraint problems we could not sort out.

tmolitor-stud-tu:
@Echolon This imlements XEP-0085 and XEP-0319 (both receiving and sending)

tmolitor-stud-tu:
@anurodhp that's it, no more pushes to this branch --> ready to be merged :)

Echolon:
@tmolitor-stud-tu @FriedrichAltheide I am maybe not the person to qualified evaluate the details but I think you do great work and also essential one!

Thank you for this! :pray:

Echolon:
Are there any issues linked/going to be closed with this PR?

Monal/Classes/EncodingTools.m
```diff
@@ -13,36 +13,89 @@
 @implementation EncodingTools
 
 
-+ (NSString *) encodeRandomResource {
++(monal_void_block_t) startTimer:(double) timeout withHandler:(monal_void_block_t) handler
```
anurodhp: not sure this is the right class for this code but it is something that can be moved around later.

Monal/Classes/MonalAppDelegate.m
```diff
@@ -419,21 +423,26 @@ -(void)applicationWillResignActive:(UIApplication *)application
 -(void) applicationDidEnterBackground:(UIApplication *)application
```
anurodhp, line 422: we will have to redo this, but it is out of scope for your work so not an issue, in a nutshell it appears the background foreground event will happen more in the future on macs and iPads that have many apps running concurrently.

Monal/Classes/HelperTools.m
```diff
@@ -1,17 +1,95 @@
...
+        [rfc3339DateFormatter2 setTimeZone:[NSTimeZone timeZoneForSecondsFromGMT:0]];
+        [rfc3339DateFormatter2 setDateFormat:@"yyyy'-'MM'-'dd'T'HH':'mm':'ssZ"];
+        retval = [rfc3339DateFormatter2 dateFromString:datetime];
+    }
+    return retval;
+}
 
++(NSString*) generateDateTimeString:(NSDate*) datetime
+{
+    NSDateFormatter* rfc3339DateFormatter = [[NSDateFormatter alloc] init];
```
anurodhp, line 175: making a date formatter continuously like this is very expensive esp if it is something that doesnt change and is called from the UI

Monal/Classes/xmpp.m
```diff
@@ -1291,21 +1292,11 @@ -(void) processInput:(XMPPParser *) parsedStanza
...
-
-                            [self send:saslXML];
-                        }
-                        else
-                        {
-                            //no supported auth mechanism
-                            DDLogInfo(@"no supported auth mechanism, disconnecting!");
-                            [self disconnect];
-                        }
+                    {
```
anurodhp, line 1469: i am assuming this is because md5 is deprecated.

Monal/Classes/xmpp.m
```diff
@@ -605,7 +609,10 @@ -(void) closeSocket
 
 -(void) reconnect
 {
-    [self reconnect:1.0];
+    if(!_exponentialBackoff)
+        _exponentialBackoff = 1.0;
+    [self reconnect:_exponentialBackoff];
+    _exponentialBackoff = MIN(_exponentialBackoff, 10.0);
```
anurodhp, line 615: if it is 1, this will always get set to 1 won't it? how does the back off increase?

Monal/Classes/ContactDetails.m
```diff
@@ -167,9 +167,10 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
                 detailCell.isContact.hidden = self.isSubscribed;
                 detailCell.isContact.text = self.subMessage;
             }
+            
             // Set human readable lastInteraction field
             NSDate* lastInteractionDate = [[DataLayer sharedInstance] lastInteractionOfJid:self.contact.contactJid forAccountNo:self.contact.accountId];
-            detailCell.lastInteraction.text = [HelperTools formatLastInteraction:lastInteractionDate];
+            detailCell.lastInteraction.text = [NSDateFormatter localizedStringFromDate:lastInteractionDate dateStyle:NSDateFormatterShortStyle timeStyle:NSDateFormatterShortStyle];
```
anurodhp, line 173: this may be expensive

Monal/Classes/HelperTools.m
```diff
@@ -101,29 +101,43 @@ +(NSString* _Nullable) formatLastInteraction:(NSDate*) lastInteraction
...
         unsigned long diff = currentTimestamp - lastInteractionTime;
-        if(diff / 60 < 1)
+        if(diff < 60)
         {
             // less than one minute
             timeString = NSLocalizedString(@"Just seen", @"");
-            diff = 0;
         }
-        else if(diff / 60 < 60)
+        else if(diff < 120)
```
anurodhp, line 109: similar behavior is provided by the relative date formatter and also takes care localization and culture

anurodhp:
generally really good. my only major concern is the date formatters it may massively degrade performance in noticeable ways (e.g. scrolling ). Everything else we can work together and clean up/test.

Monal/Classes/EncodingTools.m
tmolitor-stud-tu: I already moved it around in a later commit :)
Had to wait for a commit coming from Friedrich before I was able to merge 2 helper-classes into one file, that's why I used the EncodingTools in this commit.

Monal/Classes/MonalAppDelegate.m
tmolitor-stud-tu: You mean it would trigger a FG event while not visible in the foreground?
The plain number of those events does not matter imho, they could be twice that often without any negative effect (at least FG means "the app is visible and no throttling is required (in regard to CSI)).

Monal/Classes/HelperTools.m
tmolitor-stud-tu: yeah, fixing this now...

Monal/Classes/xmpp.m
tmolitor-stud-tu: yep exactly, I want to implement SCRAM with sha-1, sha-256 and sha-512 support later on (sadly I did not find any libs/pods already implementing this, but a simple java implementation I will try to port to objc: https://github.com/ogrebgr/scram-sasl/ )

Monal/Classes/xmpp.m
tmolitor-stud-tu: yes, I missed the increment here, fixed now

Monal/Classes/ContactDetails.m
tmolitor-stud-tu: It will only be done exactly once (when opening the detail view), seems to have no speed impact here on my device.

Monal/Classes/HelperTools.m
tmolitor-stud-tu: I and Friedrich did not know this even existed, I will take a look

tmolitor-stud-tu: Well, it would only help for minutes, hours and days, most of the code surrounding that would still be needed. For now I decided to stick to what I already implemented.

tmolitor-stud-tu:
Should all be solved now :)

PR #426, opened 2020-07-21
Fix typo

licaon-kter:


tmolitor-stud-tu:
this has to be changed on the left side of **all** strings files for all languages!

PR #446, opened 2020-08-12
Rename folder for strings

Echolon:
Regarding https://github.com/anurodhp/Monal/issues/144#issuecomment-672871926

@tmolitor-stud-tu


Echolon:
@tmolitor-stud-tu can you also remove the old folder? I dont know how to do that quickly.

PR #450, opened 2020-08-15
Fix typo in english message

cweiske:


tmolitor-stud-tu:
This has to be changed on the left side of **all** strings files for all languages!

PR #463, opened 2020-08-19
Strings reworked

comradekingu:


Monal/Base.lproj/Localizable.strings
```diff
@@ -17,7 +17,7 @@
 "Accept" = "Accept";
 
 /* No comment provided by engineer. */
-"Access message archives on the server." = "Access message archives on the server.";
+"Access message archives on the server." = "Access messages are archived on the server.";
```
licaon-kter, line 20: we "access" the "message archive"... it's a XMPP thing, the original was correct.

comradekingu: Who is the "we" there. I am thinking this could be a whole lot easier to translate if that was clear.

Echolon: Eventually turn the structure to: Archive for access messages is on the server.

?

licaon-kter: We the user of the app...

licaon-kter:
@anurodhp @tmolitor-stud-tu What is this about?

`"string here" = "string here";`
...this PR should modify both, right? And the right side would be translated, I guess

tmolitor-stud-tu:
@licaon-kter the left side is the string as used in code and **must not be changed**.

tmolitor-stud-tu:
"`Access message archive on the server`" would be correct

Echolon:
Shall I close?

Monal/Base.lproj/Localizable.strings
```diff
@@ -17,7 +17,7 @@
 "Accept" = "Accept";
 
 /* No comment provided by engineer. */
-"Access message archives on the server." = "Access message archives on the server.";
+"Access message archives on the server." = "Access your message archives on the server.";
```
comradekingu, line 20:
```suggestion
"Access message archives on the server." = "Access message archive on the server.";
```

comradekingu:
@Echolon Why not merge it? Looks good to me. Went with @tmolitor-stud-tu's suggestion.

tmolitor-stud-tu:
This changes should be done in the code, not in the translation.
Doing them in the translation will make it very complicated to change the code later on.

tmolitor-stud-tu:
--> check the code for each of these strings and change them there. then make a new pull request with these changes.

PR #465, opened 2020-08-20
Pr05 fix02

tmolitor-stud-tu:
Fixed the smacks counting bug.
It was a state race between appex and mainapp.


tmolitor-stud-tu:
don't merge, this seems to still have a bug :(

tmolitor-stud-tu:
This should solve all smacks counting bugs I told you about.

PR #466, opened 2020-08-20
List known server software by popularity

licaon-kter:
Ref: https://twitter.com/iNPUTmice/status/1257543944486785030

### DIFF
--- a/Monal/Base.lproj/Localizable.strings
+++ b/Monal/Base.lproj/Localizable.strings
@@ -284,7 +284,7 @@
 "Jabber ID" = "Jabber ID";
 
 /* No comment provided by engineer. */
-"Jabber, Openfire, Prosody etc.   " = "Jabber, Openfire, Prosody etc.   ";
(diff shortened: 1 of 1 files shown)


tmolitor-stud-tu:
this one will change nothing as long as the string is not changed in the code, too!!!

licaon-kter:
@tmolitor-stud-tu ....do explain though

anurodhp:
Oh ha I just noticed the key changed too ans  the string 

--
Anurodh Pokharel

Echolon:
@licaon-kter the string file is just an export from xcode of all the strings placed in the actual code. You actually should find the string in the Monal code and change it there. Then Thilo can export and merge a new Localizations.string file

tmolitor-stud-tu:
@licaon-kter the left side of the strings file is actually no arbitrary string, but an ID.
This ID has to equal the string used in the code/storyboard.
If you change the ID in the strings file the app won't find it and use the string from code instead (or an empty string, I don't know exactly).

PR #475, opened 2020-08-28
Spell error

poormusic2001:
Change from "Invalid Credentails" to "Invalid Credentials". It has a spell error.


licaon-kter:
Maybe rebase?

PR #487, opened 2020-09-30
Update XEPsupport.md

Echolon:


Echolon:
> I meant to only list those xeps that are either supported OR listed in the compliance suite and remove all other xeps from this list


271f0e7

tmolitor-stud-tu (approved):

PR #496, opened 2020-10-12
Fix blocking of sendqueue after pipe stream open

tmolitor-stud-tu:
The "open" event lacked a break so the switch-case did fall-through to the
"stream has space" event even if the underlying tcp connection was not
opened successful.
The following stream send blocked the send queue which could not be
cleaned up in a disconnect until the tcp connect failed ~75 seconds
later which freed up the send queue blocking which in turn unblocked the
receive queue that blocked in the send queue cleanup call of disconnect.


tmolitor-stud-tu:
will be included in 4.9

PR #514, opened 2020-11-28
Update Localizable.strings

Echolon:


tmolitor-stud-tu:
merged into alpha already

PR #560, opened 2021-02-06
Update and rename AppStoreDescrition to AppStoreDescrition.md

Echolon:


jpds:
`Descrition` isn't an English word, should it be description?

anurodhp (approved):

PR #568, opened 2021-02-12
support both import paths that pycryptodome might use

Boris-de:
Two variants of pycryptodome exist, one installs to "Crypto" and the other installs to "Cryptodome". I guess the latter is probably used to avoid problems when coexisting with the old pycrypto library?

This change allows to use the script with the pycryptodome package in Arch Linux (_community/python-pycryptodome_) for example.

### DIFF
--- a/UDPLogServer/server.py
+++ b/UDPLogServer/server.py
@@ -6,7 +6,11 @@ import ipaddress
 import json
 import zlib
 import hashlib
-from Cryptodome.Cipher import AES
+
+try:
+    from Cryptodome.Cipher import AES  # pycryptodomex
+except ImportError:
+    from Crypto.Cipher import AES  # pycryptodome
 
 def eprint(*args, **kwargs):
     print(*args, file=sys.stderr, **kwargs)


Boris-de:
As an additional note on that topic: the wiki entry is a bit confusing because it tells us to install several packages with pip that are just part of a base python installation. All except "raw-zlib" don't need to be installed via pip - and in fact many of them generate errors of some kind when you try to do it. Additionally raw-zlib does not seem to be used - I can run it without.
So I _think_ all the pip commands should be removed from that page, only pycryptodome or pycryptodomex needs to be installed. Either via the system package manager or pip.

https://github.com/monal-im/Monal/wiki/Introduction-to-use-of-Monal-UDP-Logger

Echolon:
@Boris-de thank you for your PR and comment. I will take a look at it!

We are always looking for new contributors. If you are motivated that would be really appreciated! If you like you can join the developer chat, too.

tmolitor-stud-tu (approved):

Echolon:
I removed all the pip3 command instructions from wiki page

PR #575, opened 2021-02-21
Move logic to controller from cell; Handle cell resue

anurodhp:
Stop duplicate calls. Address race conditoins and memory.
limit previews to text links of max 500kb size


anurodhp:
fixing for 4.9.2  then will make a copy of these fixes for develop.

PR #581, opened 2021-02-25
Fix bugs for attaching files and replace file icon of actions.

### LINKED ISSUE #215: Share button: allow other file types
The :paperclip: button left to the input box only allows sharing of pictures. It would be great to also have sharing of videos (maybe with recompression), and also maybe of other file types like documents etc.

### LINKED ISSUE #256: Allow downloading aesgcm (non image) URLs
When sharing non-image files from Conversations, the recipient gets an aesgcm URL that they can't open. Ideally the file would be downloaded by Monal, decrypted and passed to whatever mechanism iOS has for opening arbitrary files.

poormusic2001:
1. Change the name of the Images action to Files in MacOS.
2. Remove the duplicate image action in iOS.
3. Remove the file_attachment for replacing new images.
4. Add new icon files of the action.


FriedrichAltheide (approved):

Monal/Classes/chatViewController.m
```diff
@@ -1126,11 +1127,8 @@ -(IBAction)attach:(id)sender
         }
         [fileAction setValue:[[UIImage imageNamed:@"file-attatchment"] imageWithRenderingMode:UIImageRenderingModeAlwaysOriginal] forKey:@"image"];
         [actionControll addAction:cameraAction];
-        [actionControll addAction:photosAction];
```
tmolitor-stud-tu, line 1129: Why do you remove the photos action in iOS completely?
There are distctinct UIs for photos versus camera versus file.
Or do I miss something here?

tmolitor-stud-tu (changes requested):

Monal/Classes/chatViewController.m
tmolitor-stud-tu: Sorry, never mind...my fault :)

tmolitor-stud-tu (approved):

PR #590, opened 2021-03-05
Add the UI of setting max size for auto download file transfer.

poormusic2001:
As title.
Add the UI of setting max size for auto download file transfer in privacy setting.


FriedrichAltheide:
Already in develop. closing

PR #593, opened 2021-03-07
5.0b2 1.0

### LINKED ISSUE #256: Allow downloading aesgcm (non image) URLs

FriedrichAltheide:
**Done**
- We no longer crash on xmpp connection start while registering a new account
- Only allow LMC on encrypted messages if the new msg is also encrypted 
- Fingerprint verification via QR-Code now working as expected. Shows a warning if you scan a QR-Code for a other user than expected.

**Testing**
- Run push appex longer (should reduce needed pushes and make apple more happy)
- New maccatalyst FG->BG handling
- More db upgrades. We expect some more crashes. Please provide steps how to reproduce if a crash occurs.
- Privacy Settings reworked
- Longer background time (should reduce battery consumption)

**TODO**:
- Fix trust UI. Show trust stage. ToFU, trusted, untrusted
- Connect Auto Download sliders to auto download code

### DIFF
--- a/Monal/Classes/MLNotificationManager.m
+++ b/Monal/Classes/MLNotificationManager.m
@@ -268,7 +268,7 @@
                             typeHint = (NSString*)kUTTypeJPEG;
                         if([mimeType isEqualToString:@"image/png"])
                             typeHint = (NSString*)kUTTypePNG;
-                        if([mimeType isEqualToString:@"image/png"])
+                        if([mimeType isEqualToString:@"image/gif"])
                             typeHint = (NSString*)kUTTypeGIF;
                         NSError *error;
                         attachment = [UNNotificationAttachment attachmentWithIdentifier:info[@"cacheId"] URL:[NSURL fileURLWithPath:info[@"cacheFile"]] options:@{UNNotificationAttachmentOptionsTypeHintKey:typeHint} error:&error];


Monal/Classes/MLNotificationManager.m
```diff
@@ -253,37 +253,50 @@ -(void) showModernNotificaionForMessage:(MLMessage*) message withSound:(BOOL) so
...
+                    if(![info[@"needsDownloading"] boolValue])
                     {
-                        content.attachments = @[attachment];
-                        content.body = @"";
+                        NSString* typeHint = (NSString*)kUTTypePNG;
+                        if([mimeType isEqualToString:@"image/jpeg"])
+                            typeHint = (NSString*)kUTTypeJPEG;
+                        if([mimeType isEqualToString:@"image/png"])
+                            typeHint = (NSString*)kUTTypePNG;
+                        if([mimeType isEqualToString:@"image/png"])
```
tmolitor-stud-tu: needs image/gif, not image/png

tmolitor-stud-tu (changes requested):

Monal/Classes/MLNotificationManager.m
tmolitor-stud-tu: done

tmolitor-stud-tu (dismissed):

tmolitor-stud-tu (dismissed):

tmolitor-stud-tu (approved):

PR #599, opened 2021-03-25
MLKeysTableViewController: Replace NSLocalizedStrings with new localization codes

Echolon:
Hey,

please treat this as a discussion basis if we are heading in the right direction.

TBD:

[x] What to do with @ % variables in strings?
-- Just replace as usual

- TITLE <-> MSG connection

___
Ref.
https://github.com/monal-im/Monal-localization-main/pull/1


Echolon:
Is this basically fine and you want me to continue with new PRs?

PR #601, opened 2021-03-29
set the correct unread counter so the refresh does not erase it

anurodhp:


FriedrichAltheide:
we call countUserUnreadMessages for every cell in cellForRowAtIndexPath right now. What exactly does this PR change?

anurodhp:
they execute in tdiffernt threads and produce different results if you this call will correctly name n messages the other call in cell for row will have n-1 for the same query.  handle new message will still be in the db transaction that created the message  and the UI code will race that thread.  Actually tested this mode now and the  fix doesnt completely work its only  partial fix. ill make a better one.

tmolitor-stud-tu:
fixed in https://github.com/monal-im/Monal/commit/aea4c83b08a7d07544e498ec606cbd8a12fb62d6

PR #625, opened 2021-04-13
Fix the issues of the UI component in the CallViewController

poormusic2001:
Fix the problem which the UI components are not fit in the CallViewController


tmolitor-stud-tu:
This will be removed completely from monal and replaced later on by a voip implementation based on webrtc.
All call/voip stuff will be rewritten from scratch.

That sadly means your pull request will be of no use.

PR #634, opened 2021-05-05
Only create MLContacts through its own function

janst97:


Monal/Classes/DataLayer.m
```diff
@@ -453,11 +453,11 @@ -(BOOL) resetContactsForAccount:(NSString*) accountNo
     }];
 }
 
--(MLContact*) contactForUsername:(NSString*) username forAccount:(NSString*) accountNo
+-(NSDictionary*) contactDictionaryForUsername:(NSString*) username forAccount:(NSString*) accountNo
```
FriedrichAltheide: (NSDictonary* _Nullable)

Monal/Classes/DataLayer.m
```diff
@@ -474,37 +474,18 @@ -(MLContact*) contactForUsername:(NSString*) username forAccount:(NSString*) acc
...
-                @"pinned": @NO,
-                @"blocked": @NO,
-                @"encrypt": @NO,
-                @"muted": @NO,
-                @"status": @"",
-                @"state": @"offline",
-                @"count": @0,
-                @"isActiveChat": @NO,
-            }];
+            return (NSMutableDictionary*)nil;
```
FriedrichAltheide, line 480: return nil

janst97: Returning nil without the cast won't work.

`Return type 'NSMutableDictionary *' must match previous return type 'void *' when block literal has unspecified explicit return type`

FriedrichAltheide (approved):

tmolitor-stud-tu (approved):

PR #637, opened 2021-05-06
AddContact.m: Replace strings

Echolon:


tmolitor-stud-tu:
Why does this contain all these old commits??
Please rebase upon current develop

Echolon:
Sorry, I have not continued on this. So this task will not be needed anymore?

PR #646, opened 2021-05-18
5.0b6 1.1

tmolitor-stud-tu:
you can ignore the bookmarks stuff, they are deactivated in beta builds anyway.


FriedrichAltheide:
@tmolitor-stud-tu do you want to add 5f54d6b as well?

tmolitor-stud-tu:
This was too late for beta 6, will be included in beta 7

PR #668, opened 2021-06-17
Bugfixes for 5.0 beta

janst97:


Monal/Classes/MLUploadQueueItem.m
```diff
@@ -39,32 +39,37 @@ -(NSURL*) getURL
 
 -(id) initWithImage:(UIImage*) image {
     assert(image != nil);
-    assert(self = [super init]);
-    _type = UPLOAD_QUEUE_TYPE_RAW_IMAGE;
-    self.image = image;
-    self.url = nil;
+    if(self = [super init])
```
tmolitor-stud-tu: [super init]mus ALWAYS be called and it has to be the first line in the init method (according to apple docs)

Monal/Classes/MLUploadQueueItem.m
```diff
@@ -39,32 +39,37 @@ -(NSURL*) getURL
...
 
 -(id) initWithImage:(UIImage*) image imageUrl:(NSURL *)url
 {
     assert(image != nil);
     assert(url != nil);
-    assert(self = [super init]);
-    _type = UPLOAD_QUEUE_TYPE_IMAGE_WITH_URL;
-    self.image = image;
-    self.url = url;
+    if(self = [super init])
```
tmolitor-stud-tu: same here

Monal/Classes/MLUploadQueueItem.m
tmolitor-stud-tu: why that if btw?

Monal/Classes/MLUploadQueueItem.m
```diff
@@ -39,32 +39,37 @@ -(NSURL*) getURL
...
 
 -(id) initWithURL:(NSURL*) url
 {
     assert(url != nil);
-    assert(self = [super init]);
-    // MLUploadQueueItem* item = [MLUploadQueueItem alloc];
-    _type = UPLOAD_QUEUE_TYPE_URL;
-    self.url = url;
-    self.image = nil;
+    if(self = [super init])
```
tmolitor-stud-tu: and again

Monal/Classes/MLUploadQueueItem.m
tmolitor-stud-tu: [super init] will always return a non-nil value, no need to check it...

tmolitor-stud-tu (changes requested):
no

Monal/Classes/MLUploadQueueItem.m
janst97: I used this as reference: https://developer.apple.com/documentation/objectivec/nsobject/1418641-init?language=objc

"If the new object can’t be initialized, the method should return nil. For example, a hypothetical BuiltInCamera class might return nil from its init method if run on a device that has no camera.". However, this does not apply to NSObject, so it's not necessary here. I'll fix it.

tmolitor-stud-tu (approved):

PR #727, opened 2021-08-11
Fix jingle

tmolitor-stud-tu:
- Fix jingle


FriedrichAltheide (approved):

Peter0961:
What's about GroupChats with OMEMO?
I'm waiting for this Feature.
Then the App is the Best in the App-Store!

Echolon:
@Peter0961 yes, we would like to see this and many other features as well. But as you know this project is developed in spare-time only and has no greater funding.

If you want support the general development and help to keep it alive please consider to donate!

PR #761, opened 2022-01-09
Push unregister: use xmpp instead of http

FriedrichAltheide:
- Push unregister: use xmpp instead of http
- Switching to new alpha push server fqdn

### DIFF
--- a/Monal/Classes/MonalAppDelegate.m
+++ b/Monal/Classes/MonalAppDelegate.m
@@ -213,8 +213,8 @@ static NSString* kBackgroundFetchingTask = @"im.monal.fetch";
                 if((oldToken != nil && oldToken.length != 0) || oldGranted)
                 {
                     DDLogWarn(@"Unregistering node from appserver!");
-                    [self unregisterPush];
-                    
+
+                    [[MLXMPPManager sharedInstance] unregisterPush];
                     //this is only needed for better UI (settings --> noifications should reflect the proper state)
                     //both invalidations are needed because we don't know the timing of this notification granting handler
                     DDLogInfo(@"Invalidating all account states...");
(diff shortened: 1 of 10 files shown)


Monal/Classes/MonalAppDelegate.m
```diff
@@ -212,9 +212,6 @@ -(BOOL) application:(UIApplication*) application didFinishLaunchingWithOptions:(
                 //unregister from push appserver
                 if((oldToken != nil && oldToken.length != 0) || oldGranted)
                 {
-                    DDLogWarn(@"Unregistering node from appserver!");
```
tmolitor-stud-tu, line 215: This was intended to deactivate the device on the appserver as soon as the user disables the push entitlement in the ios app settings.

Did you add something similar somewhere else?

Monal/Classes/XMPPEdit.m
```diff
@@ -399,6 +399,14 @@ - (IBAction) deleteAccountClicked: (id) sender
     }];
     UIAlertAction* yesAction = [UIAlertAction actionWithTitle:NSLocalizedString(@"Yes", @"") style:UIAlertActionStyleDefault handler:^(UIAlertAction* action) {
         DDLogVerbose(@"Deleting accountNo %@", self.accountno);
+
```
tmolitor-stud-tu: You don't want to receive push for disabled accounts --> disable push node on appserver once all accounts are deleted/disabled (e.g. when he last account gets disabled/deleted).

This looks like you are taking disabled accounts into account while counting the accounts.
Use connected accounts instead (which include all non-disabled accounts).

tmolitor-stud-tu (changes requested):
See inline comments

tmolitor-stud-tu:
Some more comments:
We want the device to be unregistered when force disconnecting, too: use mlxmppmanager disconnectAccount to check if this was the last account and if yes, remove the registration from the appserver...that solves one of the other points in my review automatically.

Monal/Classes/MonalAppDelegate.m
tmolitor-stud-tu: why remove the DDLlogWarn()?

Monal/Classes/MonalAppDelegate.m
```diff
@@ -212,9 +212,7 @@ -(BOOL) application:(UIApplication*) application didFinishLaunchingWithOptions:(
...
-                    [self unregisterPush];
-                    
+                    [[MLXMPPManager sharedInstance] unregisterPush];
                     //this is only needed for better UI (settings --> noifications should reflect the proper state)
```
tmolitor-stud-tu, line 218: why this missing newline?

Monal/Classes/xmpp.m
```diff
@@ -3765,6 +3765,15 @@ -(void) enablePush
     }
 }
 
+-(void) unregisterPush
+{
+    DDLogInfo(@"Unregistering push");
+    XMPPIQ* unregisterNode = [[XMPPIQ alloc] initWithType:kiqSetType];
```
tmolitor-stud-tu: please use -(id) initWithType:(NSString*) iqType to:(NSString*) to;

Monal/Classes/xmpp.m
```diff
@@ -3765,6 +3765,15 @@ -(void) enablePush
...
+    [unregisterNode setUnregisterOnAppserver];
+    [unregisterNode setiqTo:[HelperTools pushServer]];
+    [self sendIq:unregisterNode withHandler:nil];
```
tmolitor-stud-tu: just use [self send:unregisterNode] if you don't want to handle the result, nil could even be no supported for the handler of sendIq!

tmolitor-stud-tu (changes requested):
some more real small things

Monal/Classes/MonalAppDelegate.m
```diff
@@ -213,8 +213,8 @@ -(BOOL) application:(UIApplication*) application didFinishLaunchingWithOptions:(
                 if((oldToken != nil && oldToken.length != 0) || oldGranted)
                 {
                     DDLogWarn(@"Unregistering node from appserver!");
-                    [self unregisterPush];
-                    
+
```
tmolitor-stud-tu, line 216: what? :D

tmolitor-stud-tu (approved):

PR #805, opened 2022-05-03
Use SwiftUI to show notification settings

janst97:


Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -139,3 +140,16 @@ class ContactDetailsInterface: NSObject {
         return host
     }
 }
+
+@objc
+class SettingsSwiftInterface : NSObject {
```
tmolitor-stud-tu: that's nice, can you port the ContactDetailsInterface.makeContactDetails to this new class, too?

Monal/Classes/NotificationSettings.swift
```diff
@@ -0,0 +1,113 @@
...
+                    Text("Not implemented yet...").font(.footnote)
+                    Picker("Push Server", selection: $pushServer) {
+                        Text("Europe 1").tag(0)
+                        Text("Europe 2").tag(1)
+                    }.pickerStyle(.menu)//.menuStyle(.borderlessButton)
+                }
+            }
+            .navigationBarHidden(true)
+        }
+        .navigationTitle(Text(NSLocalizedString("Notifications", comment: "")))
```
tmolitor-stud-tu: You don't need to wrap literal text inside Text() into NSLocalizedString (you only need to wrap text that has dynamic components)

Monal/Classes/NotificationSettings.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    var body: some View {
+        NavigationView {
+            Form {
+                Group {
+                    Section(header: Text(NSLocalizedString("Status", comment: "")).font(.title3)) {
+                        VStack(alignment: .leading) {
+                            buildLabel("Apple Push Service", isLocalized: true, isWorking: self.applePushEnabled);
+                            Divider()
+                            Text(NSLocalizedString("Apple push service should always be on. If it is off, your device can not talk to Apple's server.", comment: "")).font(.footnote)
```
tmolitor-stud-tu: see comment about Text() below

Monal/Classes/NotificationSettings.swift
```diff
@@ -0,0 +1,113 @@
...
+    let accountName: String
+    let pushEnabled: Bool
+    var id: String { accountName }
+}
+
+struct NotificationSettings: View {
+    func buildLabel(_ description: String, isLocalized: Bool, isWorking: Bool) -> some View {
+        let labelText = (isLocalized == true) ? NSLocalizedString(description, comment: "") : description;
+        if(isWorking == true) {
+            return Label(labelText, systemImage: "checkmark.seal").accentColor(.green);
```
tmolitor-stud-tu: can you use an instance of `Text()` as `description` instead of making `description` a `String`?
That way you could remove the `isLocalized` argument completely.

Monal/Classes/NotificationSettings.swift
```diff
@@ -0,0 +1,113 @@
...
+            Form {
+                Group {
+                    Section(header: Text(NSLocalizedString("Status", comment: "")).font(.title3)) {
+                        VStack(alignment: .leading) {
+                            buildLabel("Apple Push Service", isLocalized: true, isWorking: self.applePushEnabled);
+                            Divider()
+                            Text(NSLocalizedString("Apple push service should always be on. If it is off, your device can not talk to Apple's server.", comment: "")).font(.footnote)
+                        }.onTapGesture(count: 5, perform: {
+                            showPushToken = true
+                        }).alert(isPresented: $showPushToken) {
```
tmolitor-stud-tu, line 49: cool idea!

Monal/Classes/NotificationSettings.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    init() {
+        self.applePushEnabled = MLXMPPManager.sharedInstance().hasAPNSToken;
+        self.applePushToken = MLXMPPManager.sharedInstance().pushToken;
+        var xmppAccounts: [AccountPushInfo] = [];
+        for entry in MLXMPPManager.sharedInstance().connectedXMPP {
+            let xmppAccount = entry as! xmpp
+            xmppAccounts.append(
+                AccountPushInfo(accountName: xmppAccount.connectionProperties.identity.jid, pushEnabled: xmppAccount.connectionProperties.pushEnabled)
```
tmolitor-stud-tu: why packaging those values into `AccountPushInfo` instead of just using the original `xmpp` object from `connectedXMPP`?

tmolitor-stud-tu (changes requested):
overall a very good submission, for details, see my comments.

Monal/Classes/NotificationSettings.swift
janst97: Does not work here because I want to change the font to footnote size.

Monal/Classes/NotificationSettings.swift
janst97: Label takes an argument that needs to conform to StringProtocol, which Text does not. But I moved the initialization of LocalizedString to the function call now. This way I don't need isLocalized too.

Monal/Classes/NotificationSettings.swift
tmolitor-stud-tu: Okay, but all other Text(nslocalizedstring()) occurrences should be fixed

janst97: Except for `navigationTitle` there are no other occurrences, either because I need `.font` or because alert does not take Strings.

janst97: I added a function for localized text to improve readability.

tmolitor-stud-tu (approved):

PR #811, opened 2022-05-26
5.1.1b 1.0

FriedrichAltheide:
- Ported OMEMO Keys to SwiftUI
- removed old ContactDetails UI
- Improved OMEMO in MUCs
- Better check for KeyTransportMessages
- New NotificationSettings UI (still using the old push system)
- Increase memory limit on iOS
- Improve push messages even further


tmolitor-stud-tu (approved):
go!

PR #817, opened 2022-05-31
Try to fix appearance of "Close" button in "Terms of Use" Webview for…

CCordeeate:
… MacOS

Can test behaviour for MacOS in Alpha only ...

### DIFF
--- a/Monal/Classes/RegisterAccount.swift
+++ b/Monal/Classes/RegisterAccount.swift
@@ -150,19 +150,17 @@ struct RegisterAccount: View {
                             .font(.largeTitle.weight(.bold))
                             .multilineTextAlignment(.center)
                         WebView(url: URL(string: actualTermsUrl)!)
-                        if UIDevice.current.userInterfaceIdiom == .mac {
-                            Button (action: {
-                                showWebView.toggle()
-                            }){
-                                Text("Close")
-                                    .padding(9.0)
-                                    .background(Color(red: 0.897, green: 0.878, blue: 0.878))
-                                    .clipShape(RoundedRectangle(cornerRadius: 10, style: .continuous))
-                            }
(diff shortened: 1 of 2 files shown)


Monal/Classes/RegisterAccount.swift
```diff
@@ -150,7 +150,7 @@ struct RegisterAccount: View {
                             .font(.largeTitle.weight(.bold))
                             .multilineTextAlignment(.center)
                         WebView(url: URL(string: actualTermsUrl)!)
-                        if UIDevice.current.userInterfaceIdiom == .mac {
+                        if UIDevice.current.userInterfaceIdiom != .phone && UIDevice.current.userInterfaceIdiom != .pad {
```
FriedrichAltheide: should never be true. catalyst builds should be userInterfaceIdiom == .pad

tmolitor-stud-tu: @FriedrichAltheide are you sure?

CCordeeate: What I need here is a way to differentiate between MacOS and everything else (at least) because most MacOS devices will not have a touchscreen and do need a Close Button ... but iPhone and iPad al least defininitrely have a touchscreen.

tmolitor-stud-tu: I think we need a close button for iOS, too. That would be a much cleaner UX.

CCordeeate: @tmolitor-stud-tu, @FriedrichAltheide ... your decision ... ( if ( UIDevice.current.systemName.range(of: "macOS") != nil ) {...} may do the trick though )

FriedrichAltheide: `.pad && #ifdef catalyst`

PR #901, opened 2023-01-30
Clearify error message

sirdrakeistda:
Change Error Message to prevent confusion with other error messages (e.g. server error).


licaon-kter:
So Monal can't receive PMs in channels?

tmolitor-stud-tu:
> So Monal can't receive PMs in channels?

exactly

PR #946, opened 2023-09-30
Launch Screen: switch the copyright notice to 2023

lissine0:
By the way, Monal is licensed under a permissive license.
Is `All rights reserved.` accurate?
(It may give the impression that this is a proprietary app)


tmolitor-stud-tu:
Yes, it is, see this answer for example: https://opensource.stackexchange.com/a/2123

PR #960, opened 2023-10-15
Small improvements and bugfixes

lissine0:


FriedrichAltheide (approved):
After merging, we will revert b788bd6c2299f26c27f687c747a05430c9347a39. We think that a user should be able to block a group. Could you please adjust the profile image size in the contact details in another PR on catalyst and iPads so that it is a little bit larger than it was

lissine0:
Thanks for the review
> Could you please adjust the profile image size in the contact details in another PR on catalyst and iPads so that it is a little bit larger than it was

I don't think it should be bigger than 200x200 on iPad.
In fact, here are a bunch of screenshots of ContactDetails before vs now on various devices:
[contact_details_screenshots.zip](https://github.com/monal-im/Monal/files/12926238/contact_details_screenshots.zip)
What do you think? Or does 200x200 feel small on mac?

By the way, the unit used here is points, not pixels. So, how many pixels correspond to 150 points in this case? that depends on the device:
260px on iphone SE
450px on iphone 14 plus
300px on iPad
I guess the difference is due to the different sizes for the parent views.
(both iPads in the screenshots have the same ContactDetails window size, and avatar size. What differs is the size of the view in the background)

PR #1004, opened 2024-02-19
Improvements in the menu that pops up after generating an invitation

lissine0:
Display a QR code of newly-generated invitations, instead of a link
And use the built-in share functionality to share the invitation.

Note that I deleted the paragraph explaining what to do with the link, because:
1- A QR code is self-explanatory, especially if the invitation is web-based
2- with the current size of the QR code, the paragraph would be too long that small devices (iphone SE) won't display the action buttons unless the user scrolls down

Here's how the menu would look like after applying my changes: (some parts blurred)

0AD450C3-357A-4D72-9FEA-6B7A05B2E5A3_1_201_a


lissine0:
For comparison, here's how the menu looks like before this pr
6AACFBA4-1320-48A6-A616-5D10F65336D9_1_201_a

Issue #993, opened 2024-01-15
[Feature]: Rewrite the privacy settings UI using SwiftUI


tmolitor-stud-tu:
### Precondition

- [X] I understood that this template is only for feature requests and not for bug reports

### Cross-checked related issues?

- [X] I have cross-checked this overview https://github.com/monal-im/Monal/issues/322 as well as filtered for related labels https://github.com/monal-im/Monal/labels

### Describe your feature

Rewrite the current privacy settings UI using SwiftUI.

Possibly the different settings in there should be split into different settings pages rather than one like it's currently implemented.

Every setting should be accompanied by an explanatory text for better UX.

guodong323:
I hope this message finds you well. I am reaching out as I am eager to participate in Monal's Google Summer of Code 2024 project this year. To prepare adequately, Could you please assign me to tasks related to the Monal project? I am committed to delivering my best and will adhere to all guidelines and requirements.
By the way, Would it be possible for you to provide specific UI designs or relevant design documentation? Having clear guidance will greatly assist me in refactoring the project. If such resources are unavailable, I am prepared to work with the existing UI structure and utilize SwiftUI for the refactoring process.
Thank you for your patience and support. I am eager to contribute to the Monal project and showcase my skills and potential in Google Summer of Code.

mosliem:
Hello @tmolitor-stud-tu  
i 'd like to work on this issue 
can you assign it to me?

thevaidik:
Hey Vaidik here  , i am working on this issue, it would be good if i can be assigned. 
@tmolitor-stud-tu

tmolitor-stud-tu:
Unfortunately @thevaidik was the first one to ask for this issue (via 1:1 chat) so they are assigned now.

tmolitor-stud-tu:
For all interested in participating in GSOC, I wrote a quick "how to" checklist over at our wiki: https://github.com/monal-im/Monal/wiki/GSOC-2024-Guidline

vednp:
Hey @tmolitor-stud-tu, if it is still open can I work on it?

PR #1018, opened 2024-03-15
[Feature]: Rewrite the privacy settings UI using SwiftUI #993

thevaidik:
https://github.com/monal-im/Monal/assets/106925209/4cd5db16-3472-45b6-b719-f2c32dbe848d

Here is attatched recording

### DIFF
--- a/Monal/Classes/xmpp.m
+++ b/Monal/Classes/xmpp.m
@@ -2609,7 +2609,7 @@ NSString* const kStanza = @"stanza";
             NSString* authid = [parsedStanza findFirst:@"authorization-identifier#"];
             NSDictionary* authidParts = [HelperTools splitJid:authid];
             self.connectionProperties.identity.jid = authidParts[@"user"];
-            if(authidParts[@"user"] != nil)
+            if(authidParts[@"resource"] != nil)
                 self.connectionProperties.identity.resource = authidParts[@"resource"];
             
             //record SDDP support


Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,243 @@
...
+
+
+class AppStorageModel: ObservableObject
+{
+    
+    
+    
+    @AppStorage("NotificationPrivacySetting")  var notificationPrivacySetting: NotificationPrivacySettingOption = .displayNameAndMessage
+    
+    @AppStorage("OMEMODefaultOn")  var oMEMODefaultOn = false
```
tmolitor-stud-tu: please use `omemoDefaultOn` as var name

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,243 @@
...
+        
+    }
+}
+
+
+
+
+
+enum NotificationPrivacySettingOption: Int, CaseIterable
+{
```
tmolitor-stud-tu: coding style: move the `{` behind the `CaseIterable` like with classes

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,243 @@
...
+}
+
+
+
+class AppStorageModel: ObservableObject
+{
+    
+    
+    
+    @AppStorage("NotificationPrivacySetting")  var notificationPrivacySetting: NotificationPrivacySettingOption = .displayNameAndMessage
```
tmolitor-stud-tu: This does not read/write to the app storage of our container (the one returned by the helper tools `+(NSUserDefaults*) defaultsDB`), but to the main app one and thus won't be accessible by the push app extension which needs several of these settings to do its job.

See my comment on the wrapper below.

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,243 @@
...
+    
+    @AppStorage("OMEMODefaultOn")  var oMEMODefaultOn = false
+    @AppStorage("AutodeleteAllMessagesAfter3Days")  var autodeleteAllMessagesAfter3Days = false
+    
+    @AppStorage("SendLastUserInteraction")  var sendLastUserInteraction = false
```
tmolitor-stud-tu: why do you set default values (`false`)? all of these settings should always exist (created by the corresponding `MLXMPPManager` method) and if they don't exist this would clearly be a bug and a crash helpful to detect this bug.

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,243 @@
...
+    @AppStorage("NotificationPrivacySetting") private var notificationPrivacySetting: NotificationPrivacySettingOption = .displayNameAndMessage
+    
+    
+    
+    var body: some View
+    {
+        NavigationView {
+            
+            Form
+            {
```
tmolitor-stud-tu: coding style: please open the curly brace in the row above.

yes, I know that our coding style differs between objc and swift, but so do these languages itself.

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,243 @@
...
+                    {
+                        Text("Communications")
+                        
+                    }
+                    
+                    NavigationLink(destination: ViewControllerWrapper()) {
+                        Text("Media Upload & Download")
+                        
+                    }
+
```
tmolitor-stud-tu: why that many empty lines all over the place?

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,243 @@
...
+            
+        }
+    }
+}
+
+struct PrivacyView: View
+{
+    var body: some View
+    {
+        NavigationView
```
tmolitor-stud-tu: don't do this. we have dedicated wrappers to not have a swiftui navigation view inside an uikit one. check how welcomeloginview is wrapped in swiftuihelpers.swift.

Monal/Classes/WrapperPrivacy.swift
```diff
@@ -0,0 +1,86 @@
+//
+//  WrapperPrivacy.swift
+//  Monal
+//
+//  Created by Vaidik Dubey on 13/03/24.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+struct WrapperPrivacySettingsView: View {
```
tmolitor-stud-tu: no, this isn't what I meant when I talked about a wrapper. this wrapper won't work correctly because it only saves settings to the app group settings (the one accessible by the push app extension), but your `@AppStorage` in PrivacySettings.swift will load the settings from the main app settings (which will always be empty)

when I talked about a wrapper, I talked about creating your own `@propertyWrapper` correctly wrapping to the app counter based AppStorage returned by our helper tools

Monal/localization/Base.lproj/Settings.storyboard
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
tmolitor-stud-tu: don't change or use the storyboard based uikit implementation for download settings, but use swiftui to implement this

tmolitor-stud-tu (changes requested):

Monal/Classes/PrivacySettings.swift
tmolitor-stud-tu: please check the integration of the "add account" settings entry and its submenu when tapping onto "register a new account" for an inspiration on how to do it properly

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -459,7 +459,7 @@ class SwiftuiInterface : NSObject {
             case "ChatPlaceholder":
                 host.rootView = AnyView(ChatPlaceholder())
             case "Privacy" :
-                host.rootView = AnyView(Privacy())
+                host.rootView = AnyView(Privacy(appStorageModel: AppStorageModel()))
```
tmolitor-stud-tu: don't provide that when creating the view but create that model inside the init of that view

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -493,3 +493,70 @@ func getContactList(viewContact: (ObservableKVOWrapper<MLContact>?)) -> OrderedS
         return []
     }
 }
+
+@propertyWrapper
```
tmolitor-stud-tu: I don't like your @AppStorage wrapper....the original wrapper already provided something like this: @AppStorage("username", store: UserDefaults(suiteName: "group.com.hackingwithswift.unwrap")) var username: String = "Anonymous"
(e.g. you could provide the app group name to use instead of the default storage)

subclassing or otherwise wrapping this @AppStorage propertywrapper to automatically set the app group like done in the HelperTools defaultsDB method would suffice...

see: https://www.hackingwithswift.com/quick-start/swiftui/what-is-the-appstorage-property-wrapper

tmolitor-stud-tu (changes requested):
I wanted something like this:

```
//not this
@AppStorage("webrtcUseFallbackTurn", store:UserDefaults(suiteName:"group.monalalpha")) var hasFallback: Bool?
//but this
@defaultsDB("webrtcUseFallbackTurn") var hasFallback: Bool?
```

I've written the wrapper myself, see my commit c9496c6da7b0b3aa2098ec1201e5aa6ab0251656

tmolitor-stud-tu:
We are using a rebase based workflow rather than a merge based and sometimes force-push something into one of the top commits.

Please rebase this PR upon current develop and remove all commits not being yours by calling: `git rebase -i upstream/develop` (assuming you added the remote named "upstream" pointing to our github repo).

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,202 @@
...
+                        Text("Media Upload & Download")
+                        
+                    }
+                    
+                    
+                }
+            }
+            .navigationTitle("Privacy Settings")
+            .onAppear {
+                UserDefaults.standard.set(true, forKey: "HasSeenPrivacySettings")
```
tmolitor-stud-tu: Wrong user defaults: **always** use the user defaults provided by the defaultsDB method of HelperTools!

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,202 @@
...
+        super.init(coder: aDecoder, rootView: Privacy(appStorageModel: AppStorageModel()))
+    }
+}
+
+struct Privacy: View{
+    
+    @ObservedObject var appStorageModel: AppStorageModel
+    
+    var body: some View {
+        NavigationView {
```
tmolitor-stud-tu: Never use any NavigationView in your code, **always** use the provided wrappers in swiftuihelpers.swift

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,202 @@
...
+    
+    
+}
+
+
+func getNotificationPrivacyOption(_ option: NotificationPrivacySettingOption) -> String{
+    
+    switch option{
+        
+    case .displayNameAndMessage:
```
tmolitor-stud-tu: case has to be indented one level deeper than the switch statement (everything in curly braces should be indented)

Monal/Classes/PrivacySettings.swift
```diff
@@ -0,0 +1,202 @@
...
+                Toggle("Send message received state", isOn: $appStorageModel.sendReceivedMarkers)
+                Toggle("Sync Read-Markers", isOn: $appStorageModel.sendDisplayedMarkers)
+                
+            }
+            
+        }
+    }
+}
+
+struct thirdscreen: View {
```
tmolitor-stud-tu: Again: use proper names for your views. These names should reflect what the view is for / what's visible "inside" the view.

Monal/Classes/PrivacySettings.swift
tmolitor-stud-tu: again: never use NavigationView! we talked about this in chat already.

Monal/localization/Base.lproj/Settings.storyboard
tmolitor-stud-tu: As I said: this whole file should never make it into the PR

tmolitor-stud-tu (changes requested):
Please solve **all** of these issues, then ping me to request a new review.

Issue #704, opened 2021-07-18
Autofocus on textbox in chat window


roman901:
By default, I must directly click to it to activate text input, or must do much "tabbing"

tmolitor-stud-tu:
For all interested in participating in GSOC, I wrote a quick "how to" checklist over at our wiki: https://github.com/monal-im/Monal/wiki/GSOC-2024-Guidline

guodong323:
Could you please assign me to tasks related to the Monal project? I am committed to delivering my best and will adhere to all guidelines and requirements.

tmolitor-stud-tu:
@guodong323  nothing happened here, removing you from this issue

PR #1023, opened 2024-03-27
Finish Autofocus on textbox in chat window

guodong323:


tmolitor-stud-tu (changes requested):
Please only commit changes strictly needed for that fix/feature.
Changes of project.pbxproj are not needed. In general, use `git add -p` to only add parts of changed files strictly neccessary.

Please also change the commit message to state something like "Autofocus on textbox in chat window, fixes #704"

tmolitor-stud-tu:
You don't need to close your PR to change the commit message, just use the interactive rebase of git to do so: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message

Issue #990, opened 2024-01-15
[Feature]: Rewrite debug menu using SwiftUI


tmolitor-stud-tu:
### Precondition

- [X] I understood that this template is only for feature requests and not for bug reports

### Cross-checked related issues?

- [X] I have cross-checked this overview https://github.com/monal-im/Monal/issues/322 as well as filtered for related labels https://github.com/monal-im/Monal/labels

### Describe your feature

Rewrite the debug menu to contain at least the following:

1. The current text fields and the switch for configuring the UDP-logger
2. A list of current logfiles ordered by date (newest first) and a way to export (e.g. share) them
3. A button to export the database
4. A few buttons to trigger crashes like in the current debug menu

mosliem:
Hello @tmolitor-stud-tu  i 'd like to work on this issue

tmolitor-stud-tu:
For all interested in participating in GSOC, I wrote a quick "how to" checklist over at our wiki: https://github.com/monal-im/Monal/wiki/GSOC-2024-Guidline

tmolitor-stud-tu:
Nothing happened in 2 weeks, I'll remove the assignment to make it possible for other gsoc applicant to complete this task.

vednp:
Hi @tmolitor-stud-tu can I work on it?

tmolitor-stud-tu:
@vednp I just saw that you did not yet complete (or even got assigned) any of the short gsoc teaser tasks. Please complete two of these first, before starting to work on one of the bigger tasks.
I removed your assignment to this issue for now.

I'm sorry I did not realize this before.

Short teaser tasks: https://github.com/monal-im/Monal/issues?q=is%3Aissue+is%3Aopen+label%3A%22%F0%9F%8E%83+Short+GSOC+Teaser%22
Wiki entry explaining the application process: https://github.com/monal-im/Monal/wiki/GSOC-2024-Guidline

zainashraf1562:
Gsoc Applicant: I would like to work on this issue.

PR #1027, opened 2024-03-29
Rewrite debug menu using SwiftUI, Issue-990

zainashraf1562:
issue: #990 
Video Clip:
https://github.com/monal-im/Monal/assets/155921778/3e892bc4-8090-40b5-b2aa-245441a0c690


Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,154 @@
...
+//
+//  Created by Zain Ashraf on 3/23/24.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+import SwiftUI
+import monalxmpp
+
+struct LogView: View {
+    @State private var logUDPHostname: String = ""
```
tmolitor-stud-tu: Please use the appropriate property wrapper to automatically read and write from/to the defaults db, like done over here: https://github.com/monal-im/Monal/blob/develop/Monal/Classes/PrivacySettings.swift#L43

(you probably don't need an extra object to pack all these vars into, though)

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,154 @@
...
+        .navigationBarItems(trailing:
+            Button("Reconnect All", action: {
+                isReconnecting = true
+                MLXMPPManager.sharedInstance().reconnectAll()
+                DispatchQueue.main.asyncAfter(deadline: .now() + 3.0) {
+                    isReconnecting = false
+                }
+            })
+        )
+        .overlay(
```
tmolitor-stud-tu: Please use our [Loading Overlay](https://github.com/monal-im/Monal/blob/develop/Monal/Classes/LoadingOverlay.swift) instead of `.overlay` and `ProgressHUD` (we want to ditch `ProgressHUD` once everything is ported to swiftui).

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,154 @@
...
+                    logText = logFileContents
+                } catch {
+                    logText = "Only shareable for now"
+                }
+            }
+            if let dbFile = DataLayer.sharedInstance().exportDB() {
+                dbFileURL = URL(fileURLWithPath: dbFile)
+            }
+        })
+        .onDisappear(perform: {
```
tmolitor-stud-tu: You can ditch the complete `onDisappear` invocation once you migrated all your defaults db vars using the appropriate `@defaultsDB` property wrapper.

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,154 @@
...
+            
+        }
+        .padding()
+        .onAppear(perform: {
+            let db = HelperTools.defaultsDB()
+            logUDPSwitch = db.bool(forKey: "udpLoggerEnabled")
+            logUDPPort = db.string(forKey: "udpLoggerPort") ?? ""
+            logUDPHostname = db.string(forKey: "udpLoggerHostname") ?? ""
+            logUDPAESKey = db.string(forKey: "udpLoggerKey") ?? ""
+            if logUDPSwitch, let sortedLogFileInfos = HelperTools.fileLogger?.logFileManager.sortedLogFileInfos,let firstLogFileInfo = sortedLogFileInfos.first {
```
tmolitor-stud-tu: Don't check for `logUDPSwitch` here. That switch has nothing to do with whether the logfiles are accessible or not.

Besides that: the idea is to let the user not just access the newest logfile (like the old ui did), but also every other logfile listed in `sortedLogFileInfos`.

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,154 @@
...
+                logFileURL =  URL(fileURLWithPath: firstLogFileInfo.filePath)
+                do {
```
tmolitor-stud-tu: Don't try to load logfiles into memory. That's not possible and this "Only shareable for now" is an old artifact of the old UI from 2020 where the log could be displayed directly in the UI.

We don't want that now and it isn't even possible because logfiles are binary now.

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,154 @@
...
+    @State private var isReconnecting: Bool = false
+    @State private var showLogShareSheet = false
+    @State private var showDBShareSheet = false
+    @State private var logText = ""
+    @State private var logFileURL:URL?
+    @State private var dbFileURL:URL?
+    
+    var body: some View {
+        VStack {
+            HStack {
```
tmolitor-stud-tu: Didn't we talk about a tabbed layout having these tabs?

1. A tab listing all logfiles being able to export them by tapping onto them. This tab should also list the database file to export it. (A simple list of filenames would be enough to distinguish logfiles and database files)
2. A tab for configuring the udp logger
3. A tab for inducing several different crashes, to test crash reporting and logging etc.

tmolitor-stud-tu (changes requested):

zainashraf1562:
Update Video Clip:
https://github.com/monal-im/Monal/assets/155921778/e70ae13f-7bfb-4b6f-bb53-5aac95990705

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+    @State private var sortedLogFileInfos: [DDLogFileInfo] = []
+    @State private var dbFileURL: URL?
+    @State private var showShareSheet:Bool = false
+    @State private var fileURL: URL?
+    
+    var body: some View {
+        List {
+            Section(header: Text("Log Files")) {
+                ForEach(sortedLogFileInfos, id: \.self) { logFileInfo in
+                    Text(logFileInfo.fileName)
```
tmolitor-stud-tu: Why do you use `Text()` instead of `Button()`?

Why not:

```
Button(logFileInfo.fileName) {
    fileURL = URL(fileURLWithPath: logFileInfo.filePath)
} 
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+        }
+        .onAppear {
+            if let sortedLogFileInfos = HelperTools.fileLogger?.logFileManager.sortedLogFileInfos {
+                self.sortedLogFileInfos=sortedLogFileInfos
+            }
+            if let dbFile = DataLayer.sharedInstance().exportDB() {
+                dbFileURL = URL(fileURLWithPath: dbFile)
+            }
+        }
+        .sheet(isPresented: $showShareSheet, onDismiss: {
```
tmolitor-stud-tu: This can be shortened to:

```
.sheet(isPresented:$showShareSheet) {
    if let fileURL = fileURL {
        ActivityViewController(activityItems: [fileURL])
    }
}
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+            showShareSheet=false
```
tmolitor-stud-tu: You don't need to set that to false, the `.sheet()` implementation will already do this for you.

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+            Section(header: Text("Database File")) {
+                if let dbFileURL = dbFileURL {
+                    Text(dbFileURL.lastPathComponent)
+                        .onTapGesture {
+                            self.fileURL=dbFileURL
+                        }
+                }
+            }
+        }
+        .onChange(of: fileURL) { newValue in
```
tmolitor-stud-tu: Why not write a generic binding that projects an optional value to a boolean?
That way you could ditch the `.onChange` and instead use this binding and the optional `fileURL` to set the `isPresented:` of the sheet.

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+            fileURL=nil
+        }, content: {
+            if let fileURL = fileURL {
+                ActivityViewController(activityItems: [fileURL])
+            }
+        })
+    }
+}
+
+class LogDefaultDB: ObservableObject {
```
tmolitor-stud-tu: Why did you use this additional class? Couldn't you set the `@defaultsDB("udpLoggerEnabled") var udpLoggerEnabled:Bool` on the `UDPConfigView` itself?

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+struct UDPConfigView: View {
+    @ObservedObject var defaultDB: LogDefaultDB
+    var body: some View {
+        VStack{
+            HStack {
+                TextField("Hostname/IP of Logserver", text: $defaultDB.udpLoggerHostname)
+                    .textFieldStyle(.roundedBorder)
+                TextField("Port of Logserver", text: $defaultDB.udpLoggerPort)
+                    .textFieldStyle(.roundedBorder)
+                Toggle("", isOn: $defaultDB.udpLoggerEnabled)
```
tmolitor-stud-tu: why not label this toggle?

In general: wouldn't a layout having all these fields listed one per line and having a proper label on them look nicer and easier to understand (hint: `Form`)?

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+        VStack{
+            Button(action: {
+                DispatchQueue.global(qos: .default).async(execute: {
+                    HelperTools.flushLogs(withTimeout: 0.100)
+                    let handler = MLHandler(delegate: self, handlerName: "IDontKnowThis", andBoundArguments: [:])
+                    handler.call(withArguments: nil)
+                })
+            }, label: {
+                Text("Runtime Crash")
+                    .foregroundColor(.red)
```
tmolitor-stud-tu: Why not set the foreground color onto the whole `VStack` instead of every individual `Text`?

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+    var body: some View {
+        VStack{
+            Button(action: {
+                DispatchQueue.global(qos: .default).async(execute: {
+                    HelperTools.flushLogs(withTimeout: 0.100)
+                    let handler = MLHandler(delegate: self, handlerName: "IDontKnowThis", andBoundArguments: [:])
+                    handler.call(withArguments: nil)
+                })
+            }, label: {
+                Text("Runtime Crash")
```
tmolitor-stud-tu: That's not "runtime crash" but "Try to call unknown handler method".

All these crashes are runtime crashes, strictly speaking ;)

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+                    Text("Crash Testing")
+                }
+        }
+        .addLoadingOverlay(overlay)
+        .onAppear(perform: {
+            hideLoadingOverlay(overlay)
+        })
+        .onChange(of:isReconnecting) { _ in
+            if isReconnecting{
+                showLoadingOverlay(overlay, headline: NSLocalizedString("Reconnecting", comment: "Will logout and reconnect any connected accounts."))
```
tmolitor-stud-tu: you don't need `NSLocalizedString` here, `showLoadingOverlay` takes a `StringProtocol` string which will be automatically translatable.

tmolitor-stud-tu: Do you want the comment to be visible on screen? Or is that comment meant for translators?

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+            }
+            else{
+                hideLoadingOverlay(overlay)
+            }
+        }
+        .navigationBarItems(trailing:
+            Button("Reconnect All", action: {
```
tmolitor-stud-tu: A more swiftui-ish way of writing this would be:

```
Button("Reconnect All") {
    isReconnecting = true
    [...]
}
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
...
+            })
+            Spacer()
+            Button(action: {
+                HelperTools.flushLogs(withTimeout: 0.100)
+                assert(false)
+            }, label: {
+                Text("Assertion Crash")
+                    .foregroundColor(.red)
+            })
+            Spacer()
```
tmolitor-stud-tu: don't space them out evenly. on bigger screens that looks rather ugly. use a spacer with a hardcoded size instead and add a dynamic `Spacer()` as last item of your `VStack` to cover the remaining height of the view.

tmolitor-stud-tu (changes requested):

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,200 @@
+//
+//  LogView.swift
+//  Monal
+//
+//  Created by Zain Ashraf on 3/23/24.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+import SwiftUI
```
tmolitor-stud-tu: you don't need to import `SwiftUI` nor do you need to import `monalxmpp`, these are already imported globally in `SwiftuiHelpers.swift`

Monal/Classes/LogView.swift
zainashraf1562: The idea in my mind was to have it similar to a tableview as a list of buttons just seems a bit weird UI wise.

Monal/Classes/LogView.swift
zainashraf1562: i had that initially but the view was loading empty the first time we loaded a sharesheet. when i waited for logurl to change before sharing it loaded properly. I made the assumption it was because of some background task while retrieving the url for first time that cause sharesheet to show before url was set but ill look into it again.

Monal/Classes/LogView.swift
zainashraf1562: well @defaultsDB in not a Binding Value and directly making it binding by using @state has various complexities like now you have a binding<defaultsDB<String>> instead of expexted  binding<String>. So unless we change @defaultsDB to be binding in itself encapsulating the values under a class and using ObservableObject seems to be the best alternative at least from my trails.

Monal/Classes/LogView.swift
tmolitor-stud-tu: Normal clickable text in swiftui should be a button to allow blind people to use the accessibility tools of iOS. For not visually impaired people a `Text()` or `Button()` makes no visual difference.

Monal/Classes/LogView.swift
tmolitor-stud-tu: Okay, good point :)

Monal/Classes/LogView.swift
tmolitor-stud-tu: This seems still pending?

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+import monalxmpp
```
tmolitor-stud-tu: You don't need to import this either.

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -494,3 +496,17 @@ func getContactList(viewContact: (ObservableKVOWrapper<MLContact>?)) -> OrderedS
         return []
     }
 }
+
+struct ActivityViewController: UIViewControllerRepresentable {
```
tmolitor-stud-tu, line 509: Nice that you put this into `SwiftuiHelpers.swift` to be reusable :)

tmolitor-stud-tu (changes requested):
Sorry for the long delay, I was on vacation and just recently got up to speed again.

There are still some comments open, please fix them.

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
+//
+//  LogView.swift
+//  Monal
+//
+//  Created by Zain Ashraf on 3/23/24.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+import SwiftUI
```
tmolitor-stud-tu:
```suggestion
```

Monal/Classes/LogView.swift
tmolitor-stud-tu:
```suggestion
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+    @State private var fileURL: URL?
+    
+    var body: some View {
+        List {
+            Section(header: Text("Log Files")) {
+                ForEach(sortedLogFileInfos, id: \.self) { logFileInfo in
+                    Text(logFileInfo.fileName)
+                        .onTapGesture {
+                            fileURL=URL(fileURLWithPath: logFileInfo.filePath)
+                        }
```
tmolitor-stud-tu:
```suggestion
                    Button(logFileInfo.fileName) {
                        fileURL = URL(fileURLWithPath: logFileInfo.filePath)
                    }
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+                }
+            }
+            Section(header: Text("Database File")) {
+                if let dbFileURL = dbFileURL {
+                    Text(dbFileURL.lastPathComponent)
+                        .onTapGesture {
+                            self.fileURL=dbFileURL
+                        }
```
tmolitor-stud-tu:
```suggestion
                    Button(dbFileURL.lastPathComponent) {
                        self.fileURL=dbFileURL
                    }
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+                }
+            }
+        }
+        .onAppear {
+            if let sortedLogFileInfos = HelperTools.fileLogger?.logFileManager.sortedLogFileInfos {
+                self.sortedLogFileInfos=sortedLogFileInfos
```
tmolitor-stud-tu:
```suggestion
                self.sortedLogFileInfos = sortedLogFileInfos
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+        VStack(spacing: 25){
+            Button(action: {
+                DispatchQueue.global(qos: .default).async(execute: {
+                    HelperTools.flushLogs(withTimeout: 0.100)
+                    let handler = MLHandler(delegate: self, handlerName: "IDontKnowThis", andBoundArguments: [:])
+                    handler.call(withArguments: nil)
+                })
+            }, label: {
+                Text("Try to call unknown handler method")
+            })
```
tmolitor-stud-tu:
```suggestion
            Button("Try to call unknown handler method") {
                DispatchQueue.global(qos: .default).async(execute: {
                    HelperTools.flushLogs(withTimeout: 0.100)
                    let handler = MLHandler(delegate: self, handlerName: "IDontKnowThis", andBoundArguments: [:])
                    handler.call(withArguments: nil)
                })
            }
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+            Button(action: {
+                HelperTools.flushLogs(withTimeout: 0.100)
+                let delegate: AnyClass? = NSClassFromString("MonalAppDelegate")
+                print(delegate.unsafelyUnwrapped.audiovisualTypes())
+                
+            }, label: {
+                Text("Bad Access Crash")
+            })
```
tmolitor-stud-tu:
```suggestion
            Button("Bad Access Crash") {
                HelperTools.flushLogs(withTimeout: 0.100)
                let delegate: AnyClass? = NSClassFromString("MonalAppDelegate")
                print(delegate.unsafelyUnwrapped.audiovisualTypes())
                
            }
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+            }, label: {
+                Text("Fatal Error Crash")
+            })
+            Button(action: {
+                HelperTools.flushLogs(withTimeout: 0.100)
+                let crasher:Int? = nil
+                print(crasher!)
+            }, label: {
+                Text("Nil Crash")
+            })
```
tmolitor-stud-tu:
```suggestion
            Button("Assertion Crash") {
                HelperTools.flushLogs(withTimeout: 0.100)
                assert(false)
            }
            Button("Fatal Error Crash") {
                HelperTools.flushLogs(withTimeout: 0.100)
                fatalError("fatalError_example")
            }
            Button("Nil Crash") {
                HelperTools.flushLogs(withTimeout: 0.100)
                let crasher:Int? = nil
                print(crasher!)
            }
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+        TabView {
+            LogFilesView()
+                .tabItem {
+                    Image(systemName: "list.bullet")
+                    Text("Logs")
+                }
+            UDPConfigView(defaultDB: LogDefaultDB())
+                .tabItem {
+                    Image(systemName: "gear")
+                    Text("UDP Config")
```
tmolitor-stud-tu:
```suggestion
                    Text("UDP Logger")
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+            CrashTestingView()
+                .tabItem {
+                    Image(systemName: "bolt.fill")
+                    Text("Crash Testing")
+                }
+        }
+        .addLoadingOverlay(overlay)
+        .onAppear(perform: {
+            hideLoadingOverlay(overlay)
+        })
```
tmolitor-stud-tu:
```suggestion
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+        .onChange(of:isReconnecting) { _ in
```
tmolitor-stud-tu:
```suggestion
        .onChange(of: isReconnecting) { _ in
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+            if isReconnecting{
+                showLoadingOverlay(overlay, headline: "Reconnecting", description: "Will logout and reconnect any connected accounts.")
```
tmolitor-stud-tu:
```suggestion
                showLoadingOverlay(overlay, headline: "Reconnecting", description: "Will log out and reconnect all (connected) accounts.")
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+            }
+            else{
```
tmolitor-stud-tu:
```suggestion
            } else {
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+                isReconnecting = false
+            }
+        }
+        )
+    }
+}
+
+#Preview {
+    NavigationView {
+        LogView()
```
tmolitor-stud-tu:
```suggestion
        DebugView()
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+import monalxmpp
+
+extension Binding where Value == Bool {
+  static func mappedTo<Wrapped>(bindingToOptional: Binding<Wrapped?>) -> Binding<Bool> {
+    Binding<Bool>(
+      get: { bindingToOptional.wrappedValue != nil },
+      set: { newValue in if !newValue { bindingToOptional.wrappedValue = nil } }
+    )
+  }
+}
```
tmolitor-stud-tu:
```suggestion
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+        }
+        .sheet(isPresented:Binding.mappedTo(bindingToOptional: $fileURL)) {
+            if let fileURL = fileURL {
+                ActivityViewController(activityItems: [fileURL])
+            }
+        }
+    }
+}
+
+class LogDefaultDB: ObservableObject {
```
tmolitor-stud-tu:
```suggestion
class DebugDefaultDB: ObservableObject {
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+    
+    @defaultsDB("udpLoggerHostname")
+    var udpLoggerHostname: String
+    
+    @defaultsDB("udpLoggerKey")
+    var udpLoggerKey: String
+}
+
+struct UDPConfigView: View {
+    @ObservedObject var defaultDB: LogDefaultDB
```
tmolitor-stud-tu:
```suggestion
    @ObservedObject var defaultDB: DebugDefaultDB
```

Monal/Classes/LogView.swift
```diff
@@ -0,0 +1,187 @@
...
+                Text("Nil Crash")
+            })
+            Spacer()
+        }
+        .foregroundColor(.red)
+        .font(.title)
+    }
+}
+
+struct LogView: View {
```
tmolitor-stud-tu:
```suggestion
struct DebugView: View {
```

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -494,3 +496,17 @@ func getContactList(viewContact: (ObservableKVOWrapper<MLContact>?)) -> OrderedS
         return []
     }
 }
+
```
tmolitor-stud-tu, line 499:
```suggestion

extension Binding where Value == Bool {
  static func mappedTo<Wrapped>(bindingToOptional: Binding<Wrapped?>) -> Binding<Bool> {
    Binding<Bool>(
      get: { bindingToOptional.wrappedValue != nil },
      set: { newValue in if !newValue { bindingToOptional.wrappedValue = nil } }
    )
  }
}

```

tmolitor-stud-tu (approved):
I've changed various things myself and now it can be merged.

tmolitor-stud-tu:
I've fixed some more stuff, see here: https://github.com/monal-im/Monal/commit/1c0ad61aa9e6fc6a155c1073b5a2f178dac94f99

Monal/Classes/LogView.swift
zainashraf1562: Ahh, that's a perspective I overlooked. Thanks for shedding light on it.

Monal/Classes/LogView.swift
zainashraf1562: Seems like i do make the error of replacing button with clickable text. Noted :)

Monal/Classes/LogView.swift
zainashraf1562: Ah for some reason it showed as outdated so it slipped my radar.

PR #1033, opened 2024-04-06
autoDownloadSettings rewrite 

thevaidik:
swiftUI rewrite of autoDownloadSettings


tmolitor-stud-tu (changes requested):
While `MLAutoDownloadFiletransferSettingView` is good, this struct should be part of the `PrivacySettings.swift` file (remember to merge the vars in `AutoDefaultDB` into `PrivacyDefaultDB`, too).

Rationale: either every subview of the privacy settings get its own file, or none of them. Creating an extra file for every subview would generate quite a lot of files, so I opted to merge all these subviews into one file.

tmolitor-stud-tu (approved):

PR #1037, opened 2024-04-23
adding icons in privacy settings

thevaidik:
Screenshot 2024-04-23 at 19 17 02


tmolitor-stud-tu (changes requested):
Don't add image files but use the icons delivered with ios and macos: use the macos app "SF Symbols" to decide which icon to use.

See here: https://developer.apple.com/sf-symbols/

PR #1069, opened 2024-05-26
docs: update Homebrew beta install instructions

krehel:
Hi - Homebrew maintainer here 👋 

Just a small doc update.  Noticed the install instructions for the beta version of Monal was outdated - we recently deprecated `homebrew/cask-versions` and moved the contents into the main Cask repository. 

Along with that, we renamed from -suffix to `@suffix` for alpha, beta, etc.


tmolitor-stud-tu:
Cool, thanks!

Issue #1055, opened 2024-05-02
[Feature]: Get rid of sheetDismisser and use environment


tmolitor-stud-tu:
### Precondition

- [X] I understood that this template is only for feature requests and not for bug reports

### Cross-checked related issues?

- [X] I have cross-checked this overview https://github.com/monal-im/Monal/issues/322 as well as filtered for related labels https://github.com/monal-im/Monal/labels

### Describe your feature

Don't use our own `sheetDismisser` but `@Environment(\.presentationMode) var presentationMode` and `presentationMode.wrappedValue.dismiss()` instead. Should even be valid on iOS 14.

tmolitor-stud-tu:
Only cases still present in code are dismisses without animation, there seems no appropriate replacement (yet) in swiftui.

PR #1092, opened 2024-06-24
Replace sheet dismisser with dismiss action (iOS 16+ changes)

matthewrfennell:
I've only made this change for the `ImageViewer` for now, to make sure I'm on the right track first :)

If this looks good, I can remove the other usages as well.

Also - just wanted to check - we still need to support iOS 14, right?

I ask as I saw `PresentationMode` is [deprecated](https://developer.apple.com/documentation/swiftui/presentationmode), but that its replacement `DismissAction` is [iOS 15+](https://developer.apple.com/documentation/swiftui/environmentvalues/dismiss) only.

https://github.com/monal-im/Monal/assets/148400980/4043baab-0a6b-4600-a91d-403e85b9d19f

#1055

### DIFF
--- a/Monal/Classes/AddContactMenu.swift
+++ b/Monal/Classes/AddContactMenu.swift
@@ -10,7 +10,6 @@ import MobileCoreServices
 import UniformTypeIdentifiers
 
 struct AddContactMenu: View {
-    var delegate: SheetDismisserProtocol
     static private let jidFaultyPattern = "^([^@]+@)?.+(\\..{2,})?$"
 
     @State private var connectedAccounts: [xmpp]
@@ -33,11 +32,12 @@ struct AddContactMenu: View {
 
     @State private var isEditingJid = false
 
+    @Environment(\.dismiss) private var dismiss
+
     private let dismissWithNewContact: (MLContact) -> ()
     private let preauthToken: String?
 
-    init(delegate: SheetDismisserProtocol, dismissWithNewContact: @escaping (MLContact) -> (), prefillJid: String = "", preauthToken:String? = nil, prefillAccount:xmpp? = nil, omemoFingerprints: [NSNumber:Data]? = nil) {
-        self.delegate = delegate
+    init(dismissWithNewContact: @escaping (MLContact) -> (), prefillJid: String = "", preauthToken:String? = nil, prefillAccount:xmpp? = nil, omemoFingerprints: [NSNumber:Data]? = nil) {
         self.dismissWithNewContact = dismissWithNewContact
         //self.toAdd = State(wrappedValue: prefillJid)
         self.toAdd = prefillJid
@@ -258,7 +258,7 @@ struct AddContactMenu: View {
                     if self.newContact != nil {
                         self.dismissWithNewContact(newContact!)
                     } else {
-                        self.delegate.dismiss()
+                        dismiss()
                     }
                 }
             }))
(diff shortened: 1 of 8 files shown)


tmolitor-stud-tu:
You are on the right track, but I'd prefer to use the iOS 15+ DismissAction.

The current plan is to start the development of next major version (only supporting iOS 16+) in ~1 month.

I'll leave that PR open until then, but you can already add more commits getting rid of sheetDismisser in other parts of our codebase to it, if you want :)

matthewrfennell:
Ok makes sense, I'll do that. Thank you!

matthewrfennell:
This branch lives off #1100. If there are updates to that branch, I'll rebase this one on top.

I've made all the "vanilla" changes so far. A quick demo of some of the cases:

https://github.com/monal-im/Monal/assets/148400980/d52e8e82-ed23-4762-8062-4e882af3bce2

The `AVCallUI` and `AddContactMenu` are still left. These use `dismissWithoutAnimation` currently. I haven't managed to replicate the same behaviour so far with the `DismissAction`.

What do you think about dismissing these with a regular animation as well? Here's what they look like:

AVCallUI proposed:

https://github.com/monal-im/Monal/assets/148400980/ac358e5d-8f7b-4403-881e-b42cc7e0e051

CreateGroupMenu proposed:

https://github.com/monal-im/Monal/assets/148400980/1715ba9b-3375-4443-9784-4863f95a1c4b

If we want to stick without animations for these, I'll keep playing around and see if I can get it to work. But just though I'd check if there's an easy way first :)

tmolitor-stud-tu:
Yes, at least the AVCall should be without animation to match its appearance. And the appearance hast to be without animation to directly show the call screen when the app is brought into view after being in the background.

tmolitor-stud-tu:
Do you have an xmpp account? would be cool to be able to write instant mesages to discuss things rather than using github .
You can add me at xmpp:thilo.molitor@juforum.de

tmolitor-stud-tu:
This should do it:

```
withAnimation(nil) {
    dismiss()
}
```

--> [https://developer.apple.com/documentation/swiftui/managing-user-interface-state#Animate-state-transitions](https://developer.apple.com/documentation/swiftui/managing-user-interface-state#Animate-state-transitions)

tmolitor-stud-tu:
I'll do the switch to iOS 16+ either today or tomorrow and merge this PR directly afterwards.

tmolitor-stud-tu:
can you rebase this to solve the conflicts?

matthewrfennell:
> can you rebase this to solve the conflicts?

Yes! I'll get it in a good state tonight

matthewrfennell:
The only real change was in `WelcomeLogIn.swift`. Here's the demo of that part:

https://github.com/user-attachments/assets/32249d78-0b37-447d-b582-f5ee829a80f6

tmolitor-stud-tu:
That change is okay :)

tmolitor-stud-tu:
Are that all dismissers that can be removed or are there any pending? A grep for `SheetDismisserProtocol` still lists quite a bunch.

matthewrfennell:
All the remaining usages are classes where `dismissWithoutAnimation` are used.

I couldn't replace these as I could not disable the animation where Obj-C had presented the view being dismissed.

I tested disabling the `SheetDismisser` animation in a small `SwiftUI` only project, and it does work. So I think once we are full migrated to SwiftUI it will become easy to remove these usages.

PR #1100, opened 2024-06-30
Remove sheet dismisser from classes where it is unused (iOS 14 changes)

matthewrfennell:
While working through replacing `SheetDismisserProtocol` usages with `DismissAction`s, I came across these cases where the sheet dismisser is created but not used.

Therefore, we can remove these usages without needing to bump the iOS version.

I couldn't find a way to trigger the `AccountPicker`, but here is a demo of the other two views:

https://github.com/monal-im/Monal/assets/148400980/ba8ec0ec-646e-420d-ae61-a3ab07713aa1

#1055

### DIFF
--- a/Monal/Classes/AccountPicker.swift
+++ b/Monal/Classes/AccountPicker.swift
@@ -7,7 +7,6 @@
 //
 
 struct AccountPicker: View {
-    let delegate: SheetDismisserProtocol
     let contacts: [MLContact]
     let callType: MLCallType
 #if IS_ALPHA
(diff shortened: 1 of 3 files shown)


tmolitor-stud-tu:
To trigger the account view, you'll have to set up two accounts having the same contact in their contact list, then call that contact and after that, sue the iphone native call history interface to call that contact again...it then will ask you which account to use for that call (the account isn't saved into the iphone call history, only the contact jid).

Issue #1101, opened 2024-07-01
[Feature]: Implement dynamic UI updates for our OMEMO keys list


tmolitor-stud-tu:
### Precondition

- [X] I understood that this template is only for feature requests and not for bug reports

### Cross-checked related issues?

- [X] I have cross-checked this overview https://github.com/monal-im/Monal/issues/322 as well as filtered for related labels https://github.com/monal-im/Monal/labels

### Describe your feature

Make the whole `OmemoKeys.swift` file dynamic by adding several `[[MLNotificationQueue currentQueue] postNotificationName:[...] object:self.account userInfo:@{[...]}];` notifications to `MLOMEMO.m` and using them in the omemo view by `.onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("[...]")).receive(on: RunLoop.main)) { notification in [...] }`

The view should dynamically show all changes in trust, and newly added or removed keys as well as newly added or removed users in a group (and their keys).

@matthewrfennell That one could be something for you, I guess :)

tmolitor-stud-tu:
Notification names can be defined in `MLConstants.h`

tmolitor-stud-tu:
can this be closed now or is still something missing?

matthewrfennell:
Still missing one more thing unfortunately, which is making sure the device trust is synced (it could even be working at the moment,but I haven't tested it). I haven't started that yet, but should be able to take a look this weekend.

matthewrfennell:
Actually, I think this can be closed after all :)
I had mistakenly thought that device trust was synced between devices. But, if I mark a device as trusted, nothing gets sent to the server.
So, since we have handled the other two cases, I think we are now syncing everything we can.

tmolitor-stud-tu:
Perfect, thanks!! :)

PR #1105, opened 2024-07-12
Implement dynamic UI updates for OMEMO device updates

matthewrfennell:
This PR only addresses new / removed devices, and is the first part of #1101

I show the group chat view, as this is the most complex case. The "own keys" view works similarly.

https://github.com/user-attachments/assets/3c631e27-a3c7-4e84-9cd5-36961d5ab071

I removed the explicit delete action, as this is now handled by refreshing the keys. To check it still works:

https://github.com/user-attachments/assets/16cd8aac-b755-467c-8d5b-7b46533c5ff8

Even after this change, I noticed some cases where the device list does not update after creating a device (on other accounts - not the user's own account). However, these all turned out to be differences between Monal's view and the server's view, instead of mismatches between the view and Monal's own DB.

Still to do (in future PR's, if it's ok to merge this as-is):
* show added / removed users
* show changes in trust
* investigate above cases where Monal and server are out of sync (probably because server did not send the update message to Monal?)
* add animation to delete button

### DIFF
--- a/Monal/Classes/MLConstants.h
+++ b/Monal/Classes/MLConstants.h
@@ -164,6 +164,7 @@ static inline NSString* _Nonnull LocalizationNotNeeded(NSString* _Nonnull s)
 #define kMLResourceBoundNotice @"kMLResourceBoundNotice"
 #define kMonalFinishedCatchup @"kMonalFinishedCatchup"
 #define kMonalFinishedOmemoBundleFetch @"kMonalFinishedOmemoBundleFetch"
+#define kMonalOmemoStateUpdated @"kMonalOmemoStateUpdated"
 #define kMonalUpdateBundleFetchStatus @"kMonalUpdateBundleFetchStatus"
 #define kMonalIdle @"kMonalIdle"
 #define kMonalFiletransfersIdle @"kMonalFiletransfersIdle"
--- a/Monal/Classes/MLOMEMO.m
+++ b/Monal/Classes/MLOMEMO.m
@@ -101,6 +101,13 @@ static const int KEY_SIZE = 16;
     return [NSSet setWithArray:[self.monalSignalStore knownDevicesForAddressName:addressName]];
 }
 
+-(void) notifyKnownDevicesUpdated:(NSString*) jid
+{
+    [[MLNotificationQueue currentQueue] postNotificationName:kMonalOmemoStateUpdated object:self.account userInfo:@{
+        @"jid": jid
+    }];
+}
+
 -(BOOL) createLocalIdentiyKeyPairIfNeeded
 {
     if(self.monalSignalStore.deviceid == 0)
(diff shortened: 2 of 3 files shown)


Monal/Classes/MLOMEMO.m
```diff
@@ -470,6 +478,8 @@ -(void) processOMEMODevices:(NSSet<NSNumber*>*) receivedDevices from:(NSString*)
     //handle our own devicelist
     if([self.account.connectionProperties.identity.jid isEqualToString:source])
         [self handleOwnDevicelistUpdate:receivedDevices];
+
+    [self notifyKnownDevicesUpdated:source];
```
tmolitor-stud-tu: this should be in an else to not trigger the notification twice for own devicelist updates

Monal/Classes/MLConstants.h
```diff
@@ -164,6 +164,7 @@ static inline NSString* _Nonnull LocalizationNotNeeded(NSString* _Nonnull s)
 #define kMLResourceBoundNotice @"kMLResourceBoundNotice"
 #define kMonalFinishedCatchup @"kMonalFinishedCatchup"
 #define kMonalFinishedOmemoBundleFetch @"kMonalFinishedOmemoBundleFetch"
+#define kMonalKnownDevicesUpdated @"kMonalKnownDevicesUpdated"
```
tmolitor-stud-tu: since this is used for every omemo update, maybe use this name

```suggestion
#define kMonalOmemoStateUpdated @"kMonalOmemoStateUpdated"
```

tmolitor-stud-tu (changes requested):
some really small changes, everything else looks quite good! :)

after these changes it'll be okay to merge this as is, but I'm still keen to see a rawlog (recorded using the udp log receiver) of the sync issue :)

Monal/Classes/MLOMEMO.m
matthewrfennell: Thanks for catching this!
I've done it a slightly different way - removing the call from `handleOwnDevicelistUpdate` since it is only called from here. Hopefully ok, but let me know what you think :)

tmolitor-stud-tu: while removing the call to `notifyKnownDevicesUpdated` from `handleOwnDevicelistUpdate` seems fine at first glance, having it in the `handleOwnDevicelistUpdate` could be beneficial if this method is called from somewhere else, too (in the future)...so I'd rather use an if at the call site (`processOMEMODevices:from:`) rather than removing the call to `notifyKnownDevicesUpdated` from `handleOwnDevicelistUpdate`....

tmolitor-stud-tu (changes requested):

Monal/Classes/MLOMEMO.m
matthewrfennell: That's fair. Thanks for the re-review!

tmolitor-stud-tu (approved):
Great! :)

PR #1106, opened 2024-07-13
Updated BoardingCards

ryanlintott:
BoardingCards view will now scroll if necessary and will take the full available height no matter what.
- TabView was removed as it creates additional visible UI elements when scrolling.
- ForEach was changed to use a zip so that both cards and cards.indices are passed through. This avoids using the index on the full cards array with each access and prevents crashes if an invalid index were used.
- if statements were removed around optional views as they will only render if non-nil
- The Back button label was changed to a Label so that the text "Back" would be used for accessibility.

### DIFF
--- a/Monal/Classes/BoardingCards.swift
+++ b/Monal/Classes/BoardingCards.swift
@@ -30,104 +30,94 @@ struct OnboardingView: View {
     
     var body: some View {
         ZStack {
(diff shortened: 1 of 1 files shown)


tmolitor-stud-tu:
Great, thanks!!!!

PR #1112, opened 2024-07-20
Updating calls that are deprecated in iOS 16

matthewrfennell:
I've added the `if #available` guards in for now to allow for merging these changes in separate commits to bumping the version. Once the version is bumped, I'll go through and clean all of these up (along with the other guards that already exist).

There is one more change we need to do before bumping the minimum version, which is the OMEMO QR code button in group chats. I'm still working on it, but it's proving quite tricky (at least to me!) :)


tmolitor-stud-tu:
No need to do this step by step while maintaining compilability. Its okay to temporarily break things by bumping the minimum version and then fix them step by step until it compiles again :)

Monal/Classes/MemberList.swift
```diff
@@ -126,36 +125,36 @@ struct MemberList: View {
             }
             if self.muc.mucType == "group" {
                 if ownAffiliation == "owner" {
-                    return [/*"profile"*/] + reinviteEntry + ["owner", "admin", "member", "outcast"]
```
tmolitor-stud-tu: this comment should be left there because it was intentional. but of course it can be changed to `/*["profile"] +*/`

tmolitor-stud-tu: same for the other /*profile*/ parts of course

Monal/Classes/MemberList.swift
```diff
@@ -167,10 +166,7 @@ struct MemberList: View {
                 if newAffiliation == affiliations[contact] {
                     return
                 }
-                if newAffiliation == "profile" {
```
tmolitor-stud-tu: please keep that, too

Monal/Classes/MemberList.swift
```diff
@@ -278,7 +274,7 @@ struct MemberList: View {
                             .accessibilityLabel(Text("Open Profile of \(contact.contactDisplayName as String)"))
                             //invisible navigation link that can be triggered programmatically
                             .background(
-                                NavigationLink(destination: LazyClosureView(ContactDetails(delegate:SheetDismisserProtocol(), contact:contact)), tag:contact, selection:$navigationActive) { EmptyView() }
```
tmolitor-stud-tu: and this, to keep the profile navigation intact

Monal/Classes/MemberList.swift
```diff
@@ -22,7 +22,6 @@ struct MemberList: View {
     @State private var affiliations: Dictionary<ObservableKVOWrapper<MLContact>, String>
     @State private var online: Dictionary<ObservableKVOWrapper<MLContact>, Bool>
     @State private var nicknames: Dictionary<ObservableKVOWrapper<MLContact>, String>
-    @State private var navigationActive: ObservableKVOWrapper<MLContact>?
```
tmolitor-stud-tu: after reviewing the complete file: please keep everything as is, the "profile" stuff was intentionally kept in to be able to uncomment it later on if needed

Monal/Classes/RegisterAccount.swift
```diff
@@ -273,6 +273,16 @@ struct RegisterAccount: View {
         }
     }
 
+    private func termsSiteForCurrentLanguage() -> URL {
+        var languageCode: String? = nil
+        if #available(iOS 16.0, *) {
```
tmolitor-stud-tu: just do it the ios 16 way, we'll do a hard cut...that's easier :)

Monal/Classes/RegisterAccount.swift
```diff
@@ -273,6 +273,16 @@ struct RegisterAccount: View {
         }
     }
 
+    private func termsSiteForCurrentLanguage() -> URL {
```
tmolitor-stud-tu, line 276: good idea to put that into an extra func! :)

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -125,8 +125,6 @@ func mucAffiliationToString(_ affiliation: String?) -> String {
             return NSLocalizedString("Participant", comment:"muc affiliation")
         } else if affiliation == "outcast" {
             return NSLocalizedString("Blocked", comment:"muc affiliation")
-        } else if affiliation == "profile" {
```
tmolitor-stud-tu: please keep that in, too (same for the "profile" stuff below=

tmolitor-stud-tu (changes requested):

Monal/Classes/MemberList.swift
matthewrfennell: This was the change that triggered all the deletions; this format of `NavigationLink(destination:tag:selection)` is [deprecated now](https://developer.apple.com/documentation/swiftui/navigationlink/init(_:destination:tag:selection:)-8ohtn) in iOS 16.

Since it's currently not possible to trigger it (from what I can tell), should I comment it out for now or leave as-is? It won't compile for the time being until we change it.

Monal/Classes/MemberList.swift
matthewrfennell: Ah ok, thanks for the heads up! I know to check first next time :)

Monal/Classes/MemberList.swift
tmolitor-stud-tu: The proper solution would be to use a `NavigationStack` instead of a `NavigationView` in `AddTopLevelNavigation` (located at `SwiftuiHelpers.swift`) alongside the `init(_:value:)` of `NavigationLink` and `navigationDestination(for:destination:)`.

Monal/Classes/Quicksy_RegisterAccount.swift
```diff
@@ -6,12 +6,13 @@
...
 //
 
+let DEFAULT_REGION_CODE = "en";
 let QUICKSY_BASE_URL = "https://api.quicksy.im";
 
 func sendSMSRequest(to number:String) -> Promise<(data: Data, response: URLResponse)> {
     var rq = URLRequest(url: URL(string: "\(QUICKSY_BASE_URL)/authentication/\(number)")!)
     rq.httpMethod = "GET"
-    rq.addValue(Locale.current.languageCode ?? "en", forHTTPHeaderField: "Accept-Language")
+    rq.addValue(Locale.current.language.languageCode?.identifier ?? "en", forHTTPHeaderField: "Accept-Language")
```
tmolitor-stud-tu, line 15: why not use the `DEFAULT_REGION_CODE` here, too?

Monal/Classes/Quicksy_RegisterAccount.swift
```diff
@@ -265,11 +266,12 @@ struct Quicksy_RegisterAccount: View {
                     }
                 })
                 .onAppear {
+                    let regionCode = Locale.current.region?.identifier ?? DEFAULT_REGION_CODE
                     selectedCountry = countries[0]
-                    print("######## \(String(describing:Locale.current.regionCode))")
-                    print("######## \(String(describing:Locale(identifier: "en_US").localizedString(forRegionCode:Locale.current.regionCode ?? "en")))")
+                    print("######## \(String(describing:regionCode))")
```
tmolitor-stud-tu, line 271: oh, that's my fault, the `print` statements in here were just for debugging, you can safely remove them :)

tmolitor-stud-tu (approved):
Since my requested changes are only small, I'll merge this and do them myself afterwards to not delay this further :)

Thanks for your contribution!

PR #1183, opened 2024-08-02
rewrite the Change Password view in SwiftUI

lissine0:
**Notable changes:**
- Disable the "Change Password" button if at least one of the input fields is empty (instead of having a check and an error message)
- Don't prefill the old password for non-Quicksy users (see https://github.com/monal-im/Monal/commit/a426c9a77eaa4f9cd791cf52b2aecbffb37c5b04)
- Changing some of the strings:
    - "Invalid Password" -> "Wrong Password"
So it's clearer that we're talking about the "Current Password" field, not the "New Password" one (e.g. a server may have a restriction on the minimum length.)
Other suggestions: "Incorrect Password"
     - "No accounts connected" -> "Account offline"
This string is displayed when the account we're editing is offline, even if other accounts are online. (the old string is wrong)

**Notes:**
- Changing the password can be so fast that the LoadingOverlay disappears pretty quickly. Should we add a small waiting to make it stay on the screen a bit longer?
- Currently, if we're not connected to the internet, but our account on Monal is enabled, `MLXMPPManager.sharedInstance().getConnectedAccount(forID: accountId)` returns our account instead of returning `nil`, which is presumably a bug
Moreover, the checking of the old password is currently done using the local database: `MLXMPPManager.sharedInstance().isValidPassword(oldPass, forAccount: accountId)`
Combined together, these two things might cause confusion to users in edge-cases. (I can provide an example if you're curious)
 
**Potential improvements:**

- Navigate from the first input field to the second using the keyboard: use .onSubmit and https://developer.apple.com/documentation/swiftui/view/focused(_:equals:)
- Dismiss the "Change Password" view on success (Not sure if this is recommended / a good idea)

Feel free to suggest / make improvements.


tmolitor-stud-tu:
Just use the relatively new `showPromisingLoadingOverlay()` instead of the old pait of `showLoadingOverlay()` and `hideLoadingOverlay()`.
The promising loading overlay will not only return a promise, but will make that promise wait at least for one second before it gets resolved (it can take longer, if the underlying action (a promise itself) takes longer, though, so that's a minimum bound).

You can see it in action in `MemberList.swift` for example.

> Currently, if we're not connected to the internet, but our account on Monal is enabled, MLXMPPManager.sharedInstance().getConnectedAccount(forID: accountId) returns our account instead of returning nil, which is presumably a bug

No, that's not a bug. The name "connectedAccounts" is merely historic. The right name would be "enabledAccounts" as that's what it is.

You can use the `accountState` property of an xmpp object and check if it is at state `kStateBound` to know if the account is currently connected and logged in etc.

tmolitor-stud-tu:
That said: even if the account isn't currently connected due to some network outage, the password change IQ will still be sent out on the next connect after doing a smacks resume. If the resume isn't possible, the invalidation handler of the IQ will be called, if there is any.

That can be used to trigger a notification that the password change failed due to network problems and should be retried. That way it won't matter if the account is connected the moment the user tries to change the password.

Monal/Classes/ChangePassword.swift
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        guard MLXMPPManager.sharedInstance().isValidPassword(oldPass, forAccount: accountId) else {
+            errorAlert(title: Text("Wrong Password!"), message: Text("The current password is not correct."))
+            return
+        }
+
+        showLoadingOverlay(overlay, headlineView: Text("Changing Password"), descriptionView: Text(""))
+
+        account.changePassword(newPass) { success, message in
```
tmolitor-stud-tu: Rewrite `changePassword:` to return a promise so that you can easily use it together with `showPromisingLoadingOverlay`.

Monal/Classes/ChangePassword.swift
```diff
@@ -0,0 +1,98 @@
...
+            DispatchQueue.main.async {
+                hideLoadingOverlay(overlay)
+                if success {
+                    successAlert(title: Text("Success"), message: Text("The password has been changed"))
+                    MLXMPPManager.sharedInstance().updatePassword(newPass, forAccount: accountId)
```
tmolitor-stud-tu: Rewrite `changePassword:` to use an MLHandler and make that handler fulfill the pending promise (if there is one). That way you will have the promising behaviour usable for the UI while still handling the IQ result even if it comes in after the UI was closed or even swiped away (the handler and invalidation handler I already mentioned).

You could show a success or error notification if there is no promise to fulfill (meaning the user closed the UI instead of waiting) and show something in the UI otherwise. I think that would be the most robust solution and provide the best UX :)

tmolitor-stud-tu (changes requested):

tmolitor-stud-tu:
BTW:

The handler framework is documented over here: https://github.com/monal-im/Monal/wiki/Handler-Framework/

Or in this blog article: https://monal-im.org/post/00007-monal-internals-handlers/

lissine0:
(I rebased without introducing any new changes, so you can easily git diff when I do make changes)

Monal/Classes/ChangePassword.swift
lissine0: This has been a bit hard for me, especially as there are no examples in the codebase of a handler fulfilling a promise. Can I set this aside?

tmolitor-stud-tu: The proper solution would be to make promises somehow "serializable". Since promises can't truly survive a process switch/restart, adding all such promises to a global dictionary mapping some random uuid to the promise and use the uuid as serializable argument for MLHandlers, those handlers could simply ask some helper function to provide the corresponding promise, if any, and only fulfill that promise, if one is returned.

--> we need some (small) new framework for this.

Regarding promises in general: `-(AnyPromise*) checkJidType:(NSString*) jid` in `xmpp.m` might give you some inisght on how to return a promise and fulfill it, the usage of that function (in conjunction with the promising loading overlay) can be seen in `AddContactMenu.swift`

lissine0: > --> we need some (small) new framework for this.

But this is a bit out of scope for this PR, don't you think? 🙂

> Regarding promises in general

The latest version I pushed already uses promises

tmolitor-stud-tu:
Your code has a flaw: initiating a password change while either being offline or in a slow network condition and then swiping the app away will change the password on the server, but not in monal's password storage.

Use `sendIq:withHandler:` instead of `sendIq:withResponseHandler:andErrorHandler:`

tmolitor-stud-tu:
...and update the password store in the supplied handler (I'd add the handler to MLIQProcessor.m)

To inform the UI of success/error, we'll need a generalized mapping of uuids to PMKResolver functions. That uuid can be bound to the handler when its created and the handler can later on retrieve the PMKResolver corresponding to that uuid and resolve it, so that the ui is informed properly.

tmolitor-stud-tu:
That way, `-(AnyPromise*) changePassword:(NSString*) newPass` would still return a promise, but swiping away the app will just kill that promise instead of killing the handling of the iq response...

lissine0:
> To inform the UI of success/error, we'll need a generalized mapping of uuids to PMKResolver functions. That uuid can be bound to the handler when its created and the handler can later on retrieve the PMKResolver corresponding to that uuid and resolve it, so that the ui is informed properly.

Another possibility would be using a normal showLoadingOverlay, and posting a notification from the Handler's code, containing any eventual error message.
The ChangePassword view would listen for that notification, and when it arrives, it would hide the LoadingOverlay, and either show success alert or error alert, depending on the metadata in the notification.

tmolitor-stud-tu:
No, that would mean we'd have to invent new notifications for every signal we want to send to the UI. We need a generalized way to do this and what I suggested is exactly this.

@matthewrfennell could you build that "promise registry" I described in https://github.com/monal-im/Monal/pull/1183#issuecomment-2372653687 (and the following comment)? That would be really nice! Unfortunately my own time is still very limited :(

lissine0:
Recent changes:
- Rebase on the develop branch.
- Dismiss the view on success of changing password.
- Don't show the password in cleartext for Quicksy users. But prefill it still.
Instead, it will be shown in cleartext in the Account Settings -> Login Credentials

tmolitor-stud-tu:
Our new serializable promise framework was now mergend (thanks @matthewrfennell !)

Please adapt this PR to this new framework, you can see an example of such an adaption in 71465822c08b9e7227649e7da5270edda3a83202

Monal/Classes/ChangePassword.swift
```diff
@@ -0,0 +1,96 @@
+//
+//  ChangePassword.swift
+//  Monal
+//
+//  Created by lissine on 2/8/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+struct ChangePassword: View {
+    @Environment(\.presentationMode) var presentationMode
```
matthewrfennell: [PresentationMode is deprecated now](https://developer.apple.com/documentation/swiftui/presentationmode) - does it work with [DismissAction](https://developer.apple.com/documentation/swiftui/environmentvalues/dismiss)?

Monal/Classes/xmpp.m
```diff
@@ -4634,24 +4634,15 @@ -(void) createInvitationWithCompletion:(monal_id_block_t) completion
...
-                completion(NO, error ? [HelperTools extractXMPPError:error withDescription:NSLocalizedString(@"Could not change password", @"")] : NSLocalizedString(@"Could not change password: your account is currently not connected", @""));
-            });
-    }];
+    MLPromise* promise = [MLPromise new];
+    XMPPIQ* iqNode = [[XMPPIQ alloc] initWithType:kiqSetType];
+    [iqNode setiqTo:self.connectionProperties.identity.domain];
+    [iqNode changePasswordForUser:self.connectionProperties.identity.user newPassword:newPass];
+
+    [self sendIq:iqNode withHandler:$newHandlerWithInvalidation(MLIQProcessor, handlePasswordChange,handlePasswordChangeInvalidation, $ID(iqNode), $ID(newPass), $ID(promise))];
+    return [promise toAnyPromise];
```
matthewrfennell, line 4653: Promise usage looks good to me!
I tested a password change where the handler gets run by the appex - it got consumed correctly

Monal/Classes/xmpp.m
```diff
@@ -4634,24 +4634,15 @@ -(void) createInvitationWithCompletion:(monal_id_block_t) completion
...
-            dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-                completion(NO, error ? [HelperTools extractXMPPError:error withDescription:NSLocalizedString(@"Could not change password", @"")] : NSLocalizedString(@"Could not change password: your account is currently not connected", @""));
-            });
-    }];
+    MLPromise* promise = [MLPromise new];
+    XMPPIQ* iqNode = [[XMPPIQ alloc] initWithType:kiqSetType];
+    [iqNode setiqTo:self.connectionProperties.identity.domain];
+    [iqNode changePasswordForUser:self.connectionProperties.identity.user newPassword:newPass];
+
+    [self sendIq:iqNode withHandler:$newHandlerWithInvalidation(MLIQProcessor, handlePasswordChange,handlePasswordChangeInvalidation, $ID(iqNode), $ID(newPass), $ID(promise))];
```
tmolitor-stud-tu: binding `iqNode` here on handler creation is a no-op because this var will be shadowed by a call-bound `iqNode` by the iq handling subsystem...the same subsystem binding `account` when calling the handler...

Monal/Classes/ChangePassword.swift
```diff
@@ -0,0 +1,96 @@
...
+        .catch { error in
+            errorAlert(title: Text("Error"), message: Text(error.localizedDescription))
+        }
+
+    }
+
+    var body: some View {
+        Form {
+            Section(header: Text("Enter your new password. Passwords may not be empty. They may also be governed by server or company policies.")) {
+                SecureField("Current Password", text: $oldPass)
```
tmolitor-stud-tu: old pass should be visible to the user so that it can be copied and used in other non-quicksy clients...

Monal/Classes/ChangePassword.swift
```diff
@@ -0,0 +1,96 @@
...
+                        .frame(maxWidth: .infinity, alignment: .center)
+                }
+                .disabled(oldPass.isEmpty || newPass.isEmpty)
+            }
+
+        }
+        .alert(
+            alertPrompt.title,
+            isPresented: $showAlert,
+            actions: { Button("Close", action: alertPrompt.dismissCallback ?? {}) },
```
tmolitor-stud-tu: this isn't translatable, only Text() or NSLocalizedString() is translatable until someone repairs our translatable string extractor (it normally compiles the code, but that fails, so the NSLocalizedString + Text extraction is some sort of fallback)

Monal/Classes/MLIQProcessor.m
```diff
@@ -732,6 +733,28 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
     }
 $$
 
+$$class_handler(handlePasswordChange, $$ID(XMPPIQ*, iqNode), $$ID(xmpp*, account), $$ID(NSString*, newPass), $$ID(MLPromise*, promise))
```
tmolitor-stud-tu: up to now we never stored the password in the db...with this, the password will be stored in the db at least until the result iq is received (and possibly longer in the binary db data)...

I think a better fix would be to store the password in a temporary keychain using SAMKeychain and a random uuid and only bind the uuid to the handler, then use samkeychain to retrieve the actual password in the handler and write it using MLXMPPManager. Don't forget to remove the temporary password from the temp-keychain in the handler, too.

Monal/Classes/ChangePassword.swift
lissine0: Yes, it does
Thanks

Monal/Classes/MLIQProcessor.m
```diff
@@ -732,6 +733,28 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
...
+        DDLogError(@"Changing the password of the account '%@' returned error: %@", iqNode.toUser, [iqNode findFirst:@"error"]);
+        [promise reject:error];
+        return;
+    }
+    [[MLXMPPManager sharedInstance] updatePassword:newPass forAccount: account.accountID];
+    DDLogInfo(@"Successfully changed the password of the account: %@", iqNode.toUser);
+    [promise fulfill:nil];
+$$
+
+$$class_handler(handlePasswordChangeInvalidation, $$ID(XMPPIQ*, iqNode), $$ID(MLPromise*, promise))
```
tmolitor-stud-tu: invalidations should be above the corresponding normal handler in code

Monal/Classes/MLIQProcessor.m
```diff
@@ -732,6 +733,28 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
...
+    NSString* errorMessage = [NSString stringWithFormat:NSLocalizedString(@"Changing the password of the account '%@' returned timeout", @""), iqNode.fromUser];
```
tmolitor-stud-tu: it isn't a timeout but some non-smacks reconnect that can be triggered by a multitude of things. just use a generic `"Could not change password of '%@'!", iqNode.fromUser`

Monal/Classes/MLIQProcessor.m
```diff
@@ -732,6 +733,28 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
...
+    NSError* error = [NSError errorWithDomain:@"Monal" code:0 userInfo:@{NSLocalizedDescriptionKey: errorMessage}];
+    DDLogError(@"Changing the password of the account '%@' returned timeout", iqNode.fromUser);
```
tmolitor-stud-tu: always do the logging first, in case the remaining code has some error crashing the app

Monal/Classes/MLIQProcessor.m
```diff
@@ -732,6 +733,28 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
     }
 $$
 
+$$class_handler(handlePasswordChange, $$ID(XMPPIQ*, iqNode), $$ID(xmpp*, account), $$ID(NSString*, newPass), $$ID(MLPromise*, promise))
+    if([iqNode check:@"/<type=error>"])
+    {
+        NSString* errorMessage = [HelperTools extractXMPPError:iqNode withDescription:NSLocalizedString(@"Could not change password", @"")];
```
tmolitor-stud-tu: this error should name the account in question, too, just like done in the invalidation.

tmolitor-stud-tu (changes requested):

Monal/Classes/ChangePassword.swift
lissine0: I disagree about this
If the password is visible, someone who has physical access to your device can steal your password.
The reason why it's shown in clear text in Quicksy, is because Quicksy passwords are auto-generated by default.
Ideally, if a Quicksy user picks a different password, this field wouldn't be prefilled anymore.

Monal/Classes/MLIQProcessor.m
lissine0: Should these error messages be changed as well?
https://github.com/monal-im/Monal/blob/9865fd0e2/Monal/Classes/MLMucProcessor.m#L1312
https://github.com/monal-im/Monal/blob/9865fd0e2/Monal/Classes/MLMucProcessor.m#L1257
https://github.com/monal-im/Monal/blob/9865fd0e2/Monal/Classes/MLMucProcessor.m#L1057
https://github.com/monal-im/Monal/blob/9865fd0e2/Monal/Classes/MLMucProcessor.m#L602
https://github.com/monal-im/Monal/blob/9865fd0e2/Monal/Classes/MLMucProcessor.m#L527

tmolitor-stud-tu: hmm, good point :D I guess I used "timeout" instead of no description because it sounds slightly more concrete while still not being completely wrong (the no. one reason for an invalidation is a smacks timeout). At the same time it makes clear that the user can just try again and the error is only ephemeral.

I'm unsure what to do, @matthewrfennell what do you think? Keep the "timeout" or replace all these error messages with a more generic one?

matthewrfennell: I think generic is better personally, but making it clear it can be tried again: I can imagine some non-technical users would not even guess that timeout = ephemeral.

Something like this: "Could not do xyz: please try again"?

Monal/Classes/ChangePassword.swift
lissine0: Actually it is translatable, here's the initializer being used: https://developer.apple.com/documentation/swiftui/button/init(_:action:)
In fact, here's a button with a slightly different initializer, and a unique string (`Block`), being translated to French:
https://github.com/monal-im/Monal/blob/develop/Monal/Classes/BlockedUsers.swift#L93

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: Good idea, sounds perfect :)

Monal/Classes/ChangePassword.swift
tmolitor-stud-tu: This is about extracting translatable strings into our `Localizable.strings`file, not about strings detected as translated once being in that file.

We are using this set to `NO` because using `YES` tries to compile our code and that somehow fails: https://developer.apple.com/documentation/xcode/build-settings-reference#Localized-String-SwiftUI-Support

Used here: https://github.com/monal-im/Monal/blob/develop/scripts/updateLocalization.sh#L119

I've had multiple instances of `LocalizedStringKey` as argument to some SwiftUI view that weren't translatable in this mode (which is in accordance to the documentation I linked above).

You are right, the button text *is* translatable, but not *extractable*.
You can try the extraction script yourself, if you want to confirm this.

lissine0: BlockedUsers.swift was merged on September 29th https://github.com/monal-im/Monal/pull/1236#event-14443115204
And the localization extraction script was last run on October 12th, https://github.com/monal-im/Monal-localization-main/commits/main/Base.lproj/Localizable.strings
The string "Block" only appears in BlockedUsers.swift
Since it's still in Localizable.strings, this means it was extracted, right?

lissine0: The [documentation](https://developer.apple.com/documentation/swiftui/button/init(_:action:)#discussion) says that the Button initializer creates a Text view on your behalf.
If the localization extraction script works on Buttons, that's probably why.

Monal/Classes/MLIQProcessor.m
lissine0: I agree that it's a good idea for the user-facing errors.
But for logs, wouldn't it be better to explain why the operation failed? (non-smacks reconnect)

Monal/Classes/MLIQProcessor.m
lissine0: The latest version I force-pushed implemented this

Monal/Classes/ChangePassword.swift
tmolitor-stud-tu: okay, sounds reasonable.
So we should only show the password as long as it is the autogenerated one.

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: the logging already includes the function that logs this (the invalidation handler). that means the information is already there and I don't care that much, if the logged text contains "timeout", "some error" or any other description (as long as the description still contains the fact that there was an error setting the password).

Monal/Classes/ChangePassword.swift
tmolitor-stud-tu: I now spend >1 hour testing your hypothesis (my local dev environment differed from the build server so that the extraction didn't work at all) and the result is: button texts aren't extracted by our extraction script when `SWIFT_EMIT_LOC_STRINGS = NO`. Just like the documentation says. The reason is: swift code isn't compiled in that case. And without compiling, only a simple regex matcher extracting strings from `Text()` is used.

Example code for you to test it yourself:
```
Button("lksfjkfjlkfa", role:.destructive) {
    print("doing nothing...")
}
Text("lksfjkfjlkfa: XXX")
```

`lksfjkfjlkfa: XXX` gets added to our base `Localizable.strings`, but `lksfjkfjlkfa` isn't. Just like I said.

Why is `Block` still in there? I guess the auto-removal of strings not used anymore doesn't work correctly (I always tried to be conservative here, because we don't want to delete needed strings already translated. Undoing their work would make translators really unhappy.).

The "Block" string was already present in the objc code (that got extracted correctly) and was never removed, so the button still gets correctly translated now, even though the string isn't extractable and removing "Block" from the `Localizable.strings` file and running the extraction script wouldn't add it back in.

That should be fixed, of course (e.g. using the button `label` argument with a `Text()`).

The better fix would be to make the extraction work with `SWIFT_EMIT_LOC_STRINGS = YES`. All quirks would go away in this case and every invocation of a function taking a `LocalizedStringKey` would automatically be correctly extracted, even our own functions :)
@matthewrfennell if you want to experiment a bit, you can try to fix that. Last time I tried, I couldn't get it to work and eventually gave up, but maybe you'll have more luck :)

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: You implementation has a few small quirks that should be addressed (order of SAMKeychain calls, usage of uuid and jid for the keychain etc.). I'll do this myself once merged to not delay this PR further.

Monal/Classes/ChangePassword.swift
tmolitor-stud-tu: Correction: deleting unused string works as intended: "Block" doesn't exist after running our extraction script.

We are only extracting strings from beta builds to make the strings more stable. Your new blocking interface isn't yet in beta/stable but only in alpha builds.
--> "Block" is still present in the current `Localizable.strings` files.

Monal/Classes/ChangePassword.swift
tmolitor-stud-tu: I know of many users that are actively using and needing this functionality, I can't merge this without showing at least the autogenerated password.

Monal/Classes/ChangePassword.swift
```diff
@@ -0,0 +1,96 @@
...
+    var body: some View {
+        Form {
+            Section(header: Text("Enter your new password. Passwords may not be empty. They may also be governed by server or company policies.")) {
+                SecureField("Current Password", text: $oldPass)
+#if IS_QUICKSY
+                    .onAppear {
+                        oldPass = MLXMPPManager.sharedInstance().getPasswordForAccount(accountID)
+                    }
+#endif
+                SecureField("New Password", text: $newPass)
```
tmolitor-stud-tu: Same as for Buttons, not translatable without NSLocalizedString (see PasswordMigration.swift for an example).

tmolitor-stud-tu (changes requested):

Monal/Classes/ChangePassword.swift
lissine0: If they are Quicksy users, the old password will be pre-filled, but not shown in cleartext.
They can change it without knowing the old password.

tmolitor-stud-tu: I know, but to set up the same account on other apps, they need a way to get the password in cleartext.

Please make the password visible in cleartext if it is the autogenerated one.

lissine0: I'm now prefilling it and showing it in clear text for all Quicksy users.
But ideally, this should be changed in the future so it happens only with auto-generated passwords.

tmolitor-stud-tu: > But ideally, this should be changed in the future so it happens only with auto-generated passwords.

yes, that would be perfect. I'd use defaultsdb setting for this, since quicksy can only have one account. don't forget to set/reset it on account creation.

tmolitor-stud-tu (approved):

tmolitor-stud-tu:
lissine: could you squash the first 4 commits (the commits implementing the change password stuff)?

lissine0:
> lissine: could you squash the first 4 commits (the commits implementing the change password stuff)?

Done

PR #1187, opened 2024-08-02
Adding MediaGallery

thevaidik:
As described adding media gallery, where users can see sent and received media in a chat.


Monal/Classes/MediaGallery.swift
```diff
@@ -1,11 +1,11 @@
-//////////
-//////////  MediaGallery.swift
-//////////  Monal
-//////////
-//////////  Created by Vaidik(vaidik50000@gmail.com).
-//////////  Copyright © 2021 Monal.im. All rights reserved.
-
-
+//////////////
```
tmolitor-stud-tu: why add even more `//`?

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+                                    item: GifRepresentation(getData: {
+                                        try! NSData(contentsOfFile:info["cacheFile"] as! String) as Data
+                                    }), preview: SharePreview("Share image", image: Image(uiImage: image))
+                                )
+                                    .labelStyle(.iconOnly)
+                                    .foregroundColor(.primary)
+                            } else {
+                                ShareLink(
+                                    item: JpegRepresentation(getData: {
+                                        try! NSData(contentsOfFile:info["cacheFile"] as! String) as Data
```
tmolitor-stud-tu: that won't work properly for videos, because NSData is an in-memory structure and videos are usually really big

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+        // Clear existing player
+        player = nil
+        playerViewController = nil
+        
+        // Create URL
+        let videoFileUrl = URL(fileURLWithPath: filePath)
+        
+        // Create asset with MIME type
+        let videoAsset = AVURLAsset(url: videoFileUrl, options: [
+            "AVURLAssetOutOfBandMIMETypeKey": mimeType
```
tmolitor-stud-tu: That's an apple-private symbol not allowed to be used in apps uploaded to the appstore!

tmolitor-stud-tu (changes requested):

Monal/Classes/ContactDetails.swift
```diff
@@ -245,11 +245,14 @@ struct ContactDetails: View {
                                 .fixedSize(horizontal: false, vertical: true)
                         }
                     }
-                }
+                    NavigationLink(destination: MediaGalleryView(contact: contact.contactJid as String, accountNo: NSNumber(integerLiteral: contact.accountId))) {
```
tmolitor-stud-tu: You want to load this lazy to not load every media file if the user only opens the contact details, example:

```
NavigationLink(destination: LazyClosureView(OmemoKeysView(omemoKeys: OmemoKeysForChat(viewContact: contact)))) {
    Text("Encryption Keys")
}
```

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+struct MediaGalleryView: View {
+    @State private var mediaItems: [MediaItem] = []
+    let contact: String
+    let accountNo: NSNumber
+
+    var body: some View {
+        ScrollView {
+            LazyVGrid(columns: [GridItem(.adaptive(minimum: 100))], spacing: 10) {
+                ForEach(mediaItems) { item in
+                    NavigationLink(destination: MediaItemSwipeView(currentItem: item, allItems: mediaItems)) {
```
tmolitor-stud-tu: Load this lazy, too

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+        do {
+            let (cgImage, _) = try await imageGenerator.image(at: time)
+            return UIImage(cgImage: cgImage)
+        } catch {
+            DDLogError("Error generating thumbnail: \(error)")
+            return nil
+        }
+    }
+
+    func getValue<T>(_ key: String, as type: T.Type) -> T? {
```
tmolitor-stud-tu: why that? you can just use `as! String` or `as? String` instead of this `getValue()`

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+        return fileInfo[key] as? T
+    }
+
+    func getFilePath() -> String? {
```
tmolitor-stud-tu: This doesn't seem to be used anywhere. Please don't add dead code.

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+        return getValue("cacheFile", as: String.self)
+    }
+
+    func getMimeType() -> String? {
```
tmolitor-stud-tu: This doesn't seem to be used anywhere. Please don't add dead code.

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+        return getValue("mimeType", as: String.self)
+    }
+
+    func getFileName() -> String? {
```
tmolitor-stud-tu: This doesn't seem to be used anywhere. Please don't add dead code.

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+    @ObservedObject var item: MediaItem
+
+    var body: some View {
+        Group {
+            if let thumbnail = item.thumbnail {
+                Image(uiImage: thumbnail)
+                    .resizable()
+                    .scaledToFill()
+            } else {
+                Image(systemName: "exclamationmark.triangle")
```
tmolitor-stud-tu: Show a loading indicator while this is `nil`, `generateThumbnail` will already take care that a proper icon is used if loading fails (exclamationmark.triangle or a photo/video/doc icon)

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+    @Environment(\.presentationMode) var presentationMode
+
+    init(currentItem: MediaItem, allItems: [MediaItem]) {
+        self._currentIndex = State(initialValue: allItems.firstIndex(where: { $0.id == currentItem.id }) ?? 0)
+        self.allItems = allItems
+    }
+
+    var body: some View {
+        TabView(selection: $currentIndex) {
+            ForEach(allItems.indices, id: \.self) { index in
```
tmolitor-stud-tu, line 230: This won't be lazy. Please try your implementation using 4000 photos and videos to check if it is performant enough.

Just use a loop to fill the gallery with the same photo and video in an alternating fashion to test this.

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+            }
+            .padding()
+        }
+        .navigationTitle("Media Gallery")
+        .onAppear {
+            loadMediaItems()
+        }
+    }
+
+    private func loadMediaItems() {
```
tmolitor-stud-tu: This won't be lazy, generate the thumbnail (meaning: create the `MediaItem` instance only when the view displaying it is initialized (e.g. move the creation of `MediaItem` into the constructor of `MediaItemView`

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+        }
+        .frame(width: 100, height: 100)
+        .clipShape(RoundedRectangle(cornerRadius: 10))
+        .overlay(RoundedRectangle(cornerRadius: 10).stroke(Color.gray, lineWidth: 1))
+    }
+}
+
+struct MediaItemSwipeView: View {
+    @State private var currentIndex: Int
+    let allItems: [MediaItem]
```
tmolitor-stud-tu: Use the filetransfer info instead of a fully initialized `MediaItem` and only initialize a `MediaItem` once it is needed (e.g. displayed by swiping left/right)

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+
+struct MediaGalleryView: View {
+    @State private var mediaItems: [MediaItem] = []
+    let contact: String
+    let accountNo: NSNumber
+
+    var body: some View {
+        ScrollView {
+            LazyVGrid(columns: [GridItem(.adaptive(minimum: 100))], spacing: 10) {
+                ForEach(mediaItems) { item in
```
tmolitor-stud-tu: See my comment (about looping over the info dictionary) below.

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+        .frame(width: 100, height: 100)
+        .clipShape(RoundedRectangle(cornerRadius: 10))
+        .overlay(RoundedRectangle(cornerRadius: 10).stroke(Color.gray, lineWidth: 1))
+    }
+}
+
+struct MediaItemSwipeView: View {
+    @State private var currentIndex: Int
+    let allItems: [MediaItem]
+    @Environment(\.presentationMode) var presentationMode
```
tmolitor-stud-tu: This isn't used anywhere!

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+            if let _ = info["mimeType"] as? String {
+                try? ImageViewer(delegate: dismisser, info: info)
+            } else {
+                Text("Invalid file data")
+            }
+        }
+    }
+}
+
+extension UIImage {
```
tmolitor-stud-tu: This should be part of `SwiftHelpers.swift` or, even better, `HelperTools.m` or, the best option, be a new objc category `UIImage+Thumbnail.h` + `UIImage+Thumbnail.m`

Monal/Classes/ContactDetails.swift
```diff
@@ -245,11 +245,14 @@ struct ContactDetails: View {
...
+                        Text("View Media Gallery")
+                    }                }
```
matthewrfennell: Minor: the formatting looks a bit strange

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: I wrote you an explanation in chat, but here it is again for completeness:

This will try to share a jpeg image instead of the video, but that apparent jpeg image will have the contents of the video.

Furthermore the video data will be completely loaded into memory (`NSData`) before sharing it. Videos can get quite big (several hundrets or even thousands of megabytes), loading them all into memory is a bad idea.

`ShareLink` has a constructor allowing you to use a file url instead of `NSData`

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: Please refactor the code to use my code creating an `AVURLAsset` using the symlink method on ios 16 and the proper public symbol (`AVURLAssetOverrideMIMETypeKey`) on ios 17.

tmolitor-stud-tu: Please refactor the code of `MLFileTransferVideoCell.m` to use that `AVURLAsset` creating utility function, too. This utility function should be part of the  `SwiftHelpers` class in `SwiftHelpers.swift` and be bridged from `HelperTools.m` like done with the `_renderUIImageFromSVG` method already in `SwiftHelpers.swift`.

tmolitor-stud-tu (changes requested):

Monal/Classes/ContactDetails.swift
tmolitor-stud-tu: This is the wrong place to add a media gallery navigation link, this should be somewhere below encryption keys (beneath the already existing files link)

Here: https://github.com/monal-im/Monal/pull/1187/files#diff-11602ed66122cc75bd36b7ee23bfb51b8f4f67a06fc9b184f1459d0d84f2b048R395

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+        .onAppear {
+            loadMediaItems()
+        }
+    }
+
+    private func loadMediaItems() {
+        if let attachments = DataLayer.sharedInstance().allAttachments(fromContact: contact, forAccount: accountNo) as? [[String: Any]] {
+            mediaItems = attachments.compactMap { fileInfo in
+                if let mimeType = fileInfo["mimeType"] as? String,
+                   !((fileInfo["needsDownloading"] as? NSNumber)?.boolValue ?? true) &&
```
matthewrfennell, line 40: I feel extracting this expression into several conditions would make it easier to understand in one go. E.g something like:

```swift
fileIsDownloaded(fileInfo) && isDisplayableMediaType(mimeType)
```

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+
+struct MediaItemDetailView: View {
+    let item: MediaItem
+    @StateObject private var dismisser = SheetDismisserProtocol()
+    
+    var body: some View {
+        ImageViewerWrapper(info: item.fileInfo as [String: AnyObject],dismisser: dismisser)
+            .onAppear {
+                // Assuming this view is presented in a UIHostingController
+                if let hostingController = UIApplication.shared.windows.first?.rootViewController?.presentedViewController as? UIHostingController<AnyView> {
```
matthewrfennell: My XCode is complaining saying windows was deprecated in iOS 15.0.
I haven't tried it - but could something [like this](https://stackoverflow.com/a/68989580) work?

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+    init(fileInfo: [String: Any]) {
+        self.fileInfo = fileInfo
+        self.thumbnail = nil
+        Task {
+            await generateThumbnail()
+        }
+    }
+
+    @MainActor
+    func generateThumbnail() async {
```
matthewrfennell, line 64: If I have a wide image (or video) the thumbnail gets squished. For instance - this wide image looks like this:

wide-image-full

But then gets horizontally squished in the thumbnail:

wide-image-preview

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+            .onAppear {
+                // Assuming this view is presented in a UIHostingController
+                if let hostingController = UIApplication.shared.windows.first?.rootViewController?.presentedViewController as? UIHostingController<AnyView> {
+                    dismisser.host = hostingController
+                }
+            }
+    }
+}
+
+struct ImageViewerWrapper: View {
```
matthewrfennell, line 243: If I close an image, I feel it should take me back to the media gallery. Instead, it takes me back to the chat itself:

https://github.com/user-attachments/assets/04a3095d-1fc9-4498-a146-23df393f047d

Maybe even with the system back button?

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+class MediaItem: Identifiable, ObservableObject {
+    let id = UUID()
+    let fileInfo: [String: Any]
+    @Published var thumbnail: UIImage?
+
+    init(fileInfo: [String: Any]) {
+        self.fileInfo = fileInfo
+        self.thumbnail = nil
+        Task {
+            await generateThumbnail()
```
matthewrfennell, line 59: My XCode is giving a warning saying (here and on line 78 below):

"Passing argument of non-sendable type 'MediaItem' into main actor-isolated context may introduce data races"

To be honest, I don't fully understand it :) but [this guide](https://code.kiwi.com/articles/swift-5-10-concurrency-survival-guide/) seems like it might help.

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+    static var transferRepresentation: some TransferRepresentation {
+        DataRepresentation(exportedContentType: .jpeg) { item in
+            { () -> Data in
+                return item.getData()
+            }()
+        }
+    }
+}
+
+struct SVGRepresentation: Transferable {
```
matthewrfennell, line 37: Is there any way to combine / deduplicate these three classes? It feels like they're doing _almost_ exactly the same thing.

Monal/Classes/MediaGallery.swift
matthewrfennell, line 243: The close button works on iPhone, but doesn't seem to work on iPad:


https://github.com/user-attachments/assets/67aec727-7a0f-4c8d-8e06-f04f94cde48a

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+
+struct ImageViewer: View {
+    var delegate: SheetDismisserProtocol
+    let info: [String:AnyObject]
+    @State private var previewImage: UIImage?
+    @State private var controlsVisible = false
+    @StateObject private var customPlayer = CustomAVPlayer()
+    @State private var isPlayerReady = false
+
+    init(delegate: SheetDismisserProtocol, info:[String:AnyObject]) throws {
```
matthewrfennell: Minor: spaces after the colons should be consistent

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+                }
+            } else {
+                invalidFileView
+            }
+            
+            if controlsVisible {
+                controlsOverlay
+            }
+        }.onTapGesture(count: 1) {
+            controlsVisible = !controlsVisible
```
matthewrfennell: Will do the same thing, but I think `.toggle()` is more idiomatic

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+            "AVURLAssetOutOfBandMIMETypeKey": mimeType
+        ])
+        
+        // Create player and player view controller
+        let playerItem = AVPlayerItem(asset: videoAsset)
+        player = AVPlayer(playerItem: playerItem)
+        playerViewController = AVPlayerViewController()
+        playerViewController?.player = player
+        
+        print("Created AVPlayer(\(mimeType)): \(String(describing: player))")
```
matthewrfennell: We should use `DDLogDebug` or similar here

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+                )
+        }.frame(height: 80)
+    }
+    
+    private func loadPreviewAndConfigurePlayer() async {
+        if (info["mimeType"] as! String).hasPrefix("image/svg") {
+            previewImage = await HelperTools.renderUIImage(fromSVGURL:URL(fileURLWithPath:info["cacheFile"] as! String)).toGuarantee().asyncOnMainActor()
+        } else if (info["mimeType"] as! String).hasPrefix("image/") {
+            previewImage = UIImage(contentsOfFile:info["cacheFile"] as! String)
+        } else if (info["mimeType"] as! String).hasPrefix("video/") {
```
matthewrfennell: I'm not sure if I have any solutions :) but it feels like we are always having to check between these three mime types in multiple places in the code. Is there any way we can put all of the logic for each one in their own class, or extract these checks out some other way?

Ideally, if we want to add support for a new mime type, we would have to do it in one place only. It may not be possible to reach that ideal, but thought I'd comment in case you have any ideas :)

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
+//
+// MediaGallery.swift
+//  Monal
+//
+//  Created by Vaidik(vaidik50000@gmail.com).
+//  Copyright © 2021 Monal.im. All rights reserved.
```
lissine0, line 6: Please set the correct year here. It's useful to easily know roughly when a source file was created.

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
+//
+// MediaGallery.swift
```
lissine0: Please add a space to have the same indentation

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
+////
+////  ImageViewer.swift
+////  Monal
+////
+////  Created by Friedrich Altheide on 07.10.23.
+////  Copyright © 2023 monal-im.org. All rights reserved.
+////
+//
```
lissine0, line 1: Please remove the first set of vertical //

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+//  Monal
+//
+//  Created by Vaidik(vaidik50000@gmail.com).
```
lissine0: You can remove the email address, and add the creation date (you can set it to today), so it's consistent with the rest of the codebase.
Should it be needed, your email address can be found in the git commit metadata. It's visible in git log.

Monal/Classes/ContactDetails.swift
lissine0: contact.accountId is already an NSNumber https://github.com/monal-im/Monal/blob/develop/Monal/Classes/MLContact.h#L51
No need to cast it (implicitly) to integer and immediately back to NSNumber.
You can pass it directly.

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+                controlsOverlay
+            }
+        }.onTapGesture(count: 1) {
+            controlsVisible = !controlsVisible
+        }.task {
+            await loadPreviewAndConfigurePlayer()
+        }
+    }
+    
+    private var invalidFileView: some View {
```
lissine0: For better performance, make this a struct that conforms to View.
For more details, see [this](https://medium.com/@vladislavshkodich/mastering-swiftui-are-you-really-as-good-as-you-think-40a4953f7e88#4127)

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+            Image(systemName: "xmark.square.fill")
+                .resizable()
+                .frame(width: 128.0, height: 128.0)
+                .symbolRenderingMode(.palette)
+                .foregroundStyle(.white, .red)
+            Spacer()
+        }
+    }
+    
+    private var controlsOverlay: some View {
```
lissine0: Same here. Make this a struct that conforms to View

lissine0: When you change it to struct, you should also make it conform to Equatable, as that'll achieve better performance.
For more details, see [this](https://medium.com/@vladislavshkodich/mastering-swiftui-are-you-really-as-good-as-you-think-40a4953f7e88#1adf)

Monal/Classes/MediaGallery.swift
lissine0: This is due to .scaledToFill in L177
See if it looks better with .scaledToFit instead

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,253 @@
...
+
+    var body: some View {
+        Group {
+            if let thumbnail = item.thumbnail {
+                Image(uiImage: thumbnail)
+                    .resizable()
+                    .scaledToFill()
+            } else {
+                Image(systemName: "exclamationmark.triangle")
+                    .resizable()
```
lissine0, line 182: .resizable() is applied in both cases. So, it should be applied on the whole Group instead.

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+            
+            if (info["mimeType"] as! String).hasPrefix("image/svg") {
+                VStack {
+                    ZoomableContainer(maxScale:8.0, doubleTapScale:4.0) {
+                        SVGView(contentsOf: URL(fileURLWithPath:info["cacheFile"] as! String))
+                    }
+                }
+            } else if (info["mimeType"] as! String).hasPrefix("image/") {
+                if let image = UIImage(contentsOfFile:info["cacheFile"] as! String) {
+                    VStack {
```
lissine0, line 75: ~This VStack isn't needed because it only contains one subview.~
Edit: not sure if it's needed or not. We have to test https://github.com/monal-im/Monal/pull/1187#issuecomment-2301893442

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+        self.delegate = delegate
+        self.info = info
+    }
+    
+    var body: some View {
+        ZStack(alignment: .top) {
+            Color.background
+                .edgesIgnoringSafeArea(.all)
+            
+            if (info["mimeType"] as! String).hasPrefix("image/svg") {
```
lissine0, line 67: ~The case of svg images should be inside the L73,85 code block, like the case of gif images.
(also, .scaledToFit() is missing )~

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+                VStack {
+                    ZoomableContainer(maxScale:8.0, doubleTapScale:4.0) {
+                        SVGView(contentsOf: URL(fileURLWithPath:info["cacheFile"] as! String))
+                    }
+                }
+            } else if (info["mimeType"] as! String).hasPrefix("image/") {
+                if let image = UIImage(contentsOfFile:info["cacheFile"] as! String) {
```
lissine0, line 74: The image constant is only used in the else statement (L81)
Is there a reason why it's needed for gifs (but not for SVGs)?
If it's not needed in both of those cases, then the declaration should be moved down to L80
@tmolitor-stud-tu thoughts?

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+                    HStack {
+                        Spacer().frame(width:20)
+                        Text(info["filename"] as! String).foregroundColor(.primary)
+                        Spacer()
+                        
+                        if let image = previewImage {
+                            if (info["mimeType"] as! String).hasPrefix("image/svg") {
+                                ShareLink(
+                                    item: SVGRepresentation(getData: {
+                                        try! NSData(contentsOfFile:info["cacheFile"] as! String) as Data
```
lissine0: The argument to SVGRepresentation, GifRepresentation,(L148) JpegRepresentation (L156) is the same.
It can be made a constant that's computed once beforehand, and passed to the structs.

tmolitor-stud-tu:
Thats code originally written by me. If I recall correctly there was a reason to not integrate svg images more. I think it was because the svgview is NOT an image and as such doesn't allow us to use the same view modifiers as the Image view...

Am 21. August 2024 13:10:05 MESZ schrieb lissine0 ***@***.***>:

tmolitor-stud-tu:
I *think* there was a layout/spacing issue (for gifs of certain sizes/aspect ratios?) without that VStack, so I added this extra VStack to fix that...but maybe that was something superseded later on in the development cycle of my original image viewer and isn't needed anymore?

Am 21. August 2024 13:03:46 MESZ schrieb lissine0 ***@***.***>:

tmolitor-stud-tu:
No, that won't work, svg images are rendered entirely different and aren't image views at all...

Am 21. August 2024 13:10:05 MESZ schrieb lissine0 ***@***.***>:

tmolitor-stud-tu:
Yeah, good catch. I was a but lazy when writing the original image viewer...

Am 21. August 2024 13:38:23 MESZ schrieb lissine0 ***@***.***>:

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,230 @@
...
+                                    }), preview: SharePreview("Share image", image: Image(uiImage: image))
+                                )
+                                    .labelStyle(.iconOnly)
+                                    .foregroundColor(.primary)
+                            } else if (info["mimeType"] as! String).hasPrefix("image/gif") {
+                                ShareLink(
+                                    item: GifRepresentation(getData: {
+                                        try! NSData(contentsOfFile:info["cacheFile"] as! String) as Data
+                                    }), preview: SharePreview("Share image", image: Image(uiImage: image))
+                                )
```
lissine0, line 236: The only difference between the three Sharelinks, is the argument passed to `item:`
And that argument always [depends on the same value](https://github.com/monal-im/Monal/pull/1187/commits/0f9faea#r1724902386).
Thus, only one ShareLink should be used, and the correct *Representation struct should be passed to it.

lissine0:
@tmolitor-stud-tu the question is if MediaViewer [L74](https://github.com/thevaidik/Monal-dev/blob/0f9faea/Monal/Classes/MediaViewer.swift#L74) is needed or not when the image is a gif.
If it is not needed for gifs, then that `if let image = ...` can be moved to [L80](https://github.com/thevaidik/Monal-dev/blob/0f9faea/Monal/Classes/MediaViewer.swift#L80).
In that case, the the test for image/svg prefix can be moved down next to the test for image/gif, and it won't be a problem as the GIFViewer and SVGView are separate views, and only one would be visible at a time.
Moreover, the modifiers are already directly applied to the `GIFViewer` and `Image`, and thus image modifiers won't be an issue

lissine0:
> I *think* there was a layout/spacing issue (for gifs of certain sizes/aspect ratios?) without that VStack, so I added this extra VStack to fix that...but maybe that was something superseded later on in the development cycle of my original image viewer and isn't needed anymore?

Maybe you're thinking about [this](https://github.com/thevaidik/Monal-dev/blob/0f9faea/Monal/Classes/MediaViewer.swift#L127) VStack instead?
It has only one subview, but it has [`.frame(height: 80)`](https://github.com/thevaidik/Monal-dev/blob/0f9faea/Monal/Classes/MediaViewer.swift#L175) applied to it.

Monal/Classes/ContactDetails.swift
```diff
@@ -250,6 +250,7 @@ struct ContactDetails: View {
                 .padding([.top, .bottom])
                 .frame(maxWidth: .infinity)
             }
+
```
tmolitor-stud-tu, line 253: why that extra empty line?

Monal/Classes/ContactDetails.swift
```diff
@@ -391,6 +392,9 @@ struct ContactDetails: View {
                     }) {
                         Text("Show shared Media and Files")
                     }
+                    NavigationLink(destination: LazyClosureView{MediaGalleryView(contact: contact.contactJid as String, accountNo: NSNumber(integerLiteral: contact.accountId))}) {
```
tmolitor-stud-tu: `contact.accountId` already is a `NSNumber` no need to convert it

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,240 @@
+//
+//  MediaGallery.swift
+//  Monal
+//
+//  Created by Vaidik on 21.09.24.
```
tmolitor-stud-tu: why is that date in the future? shouldn't it e roughly around the date you first created this pull request (~3 weeks ago)?

Monal/Classes/MediaGallery.swift
tmolitor-stud-tu: Lets ignore that warning for now :)

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,240 @@
...
+    var body: some View {
+        Group {
+            if item.thumbnail == nil {
+                ProgressView()
+                    .progressViewStyle(CircularProgressViewStyle())
+                    .frame(width: 50, height: 50)
+            } else if let thumbnail = item.thumbnail {
+                Image(uiImage: thumbnail)
+                    .resizable()
+                    .scaledToFill()
```
tmolitor-stud-tu: Your if statement is not clean and the else case is never reached at all, consider cleaning this up:
```suggestion
            if if let thumbnail = item.thumbnail {
                Image(uiImage: thumbnail)
                    .resizable()
                    .scaledToFill()
            } else {
                ProgressView()
                    .progressViewStyle(CircularProgressViewStyle())
                    .frame(width: 50, height: 50)
            }
```

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,240 @@
...
+            } else {
```
tmolitor-stud-tu: this else will *never* be reached.

Monal/Classes/MediaGallery.swift
tmolitor-stud-tu: That's resolved, the else is never hit now :D

Monal/Classes/MediaGallery.swift
tmolitor-stud-tu: What about this?

Monal/Classes/MediaGallery.swift
tmolitor-stud-tu: I think this is because of line 212: https://github.com/monal-im/Monal/pull/1187/files#diff-3d6e9cc2651763650c1a8ebdb0c7e9c0c6afa12a96d5d995df6a4ce5fd364568R212

tmolitor-stud-tu: Also Line 212, I guess

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: the `if let image` is used to detect if this is a regular image and show the "invalid image" ui if it is not. even though it will not be used in the gif case (we won't show the static image, but an animated version in this case).

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: loading all the date even though the user ma never press the share button would cost a lot of memory.

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: you will only be able to create the correct representation struct once you checked the mime type. creating a let constant containing the representation struct will make you replicate the mime-type if statements. I tested both and found the replicated sharelink code slightly more readable than the replicated if statements.

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: hmm, that would be really cool. I don't have a concrete idea on how to accomplish this, though. if you have a concrete code suggestion, let me know :)

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: this one isn't solved yet

tmolitor-stud-tu (changes requested):
still things to solve, in particular cleaning up the code and solving the performance issues (lazy loading!).

tmolitor-stud-tu:
Why didn't you add a preview image to the share link in case of a video? We even have a thumbnail generating function for videos now, why not use that?

Monal/Classes/MediaGallery.swift
thevaidik: I'll fix it, someone said to make it on the day of final pr in comments, so I did it.

Monal/Classes/MediaGallery.swift
```diff
@@ -2,24 +2,32 @@
 //  MediaGallery.swift
 //  Monal
 //
-//  Created by Vaidik on 21.09.24.
+//  Created by Vaidik on 01.09.24.
```
tmolitor-stud-tu: still in the future, right? :D

Monal/Classes/MediaGallery.swift
```diff
@@ -0,0 +1,313 @@
...
+    //                   !((fileInfo["needsDownloading"] as? NSNumber)?.boolValue ?? true) &&
+    //                    (mimeType.starts(with: "image/") || mimeType.starts(with: "video/")) {
+    //                    return MediaItem(fileInfo: fileInfo)
+    //                }
+    //                return nil
+    //            }
+    //        }
+    //    }
+    //}
+    private func loadMediaItems() {
```
tmolitor-stud-tu: wrong name of this function

Monal/Classes/MediaGallery.swift
tmolitor-stud-tu: Is that fixed now?

tmolitor-stud-tu (changes requested):

Monal/Classes/MediaGallery.swift
thevaidik: what about "retrieveAvailableMediaItems()" ?

thevaidik: or "fetchDownloadedMediaItems()"

tmolitor-stud-tu: `fetchDownloadedMediaItems` sounds good... `getDownloadedMediaItems` woul be another joice

Monal/Classes/MediaGallery.swift
tmolitor-stud-tu: you are right...but now the if statement is multiline which is okay imho :)

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: Lets keep that and, if you want, you can test it after we merged this PR :)

Monal/Classes/MediaViewer.swift
tmolitor-stud-tu: Let's merge this PR and figure that out afterwards :)

tmolitor-stud-tu (changes requested):

Monal/Classes/HelperTools.m
```diff
@@ -2835,4 +2835,56 @@ +(NSURLSession*) createEphemeralURLSession
...
+        [[NSFileManager defaultManager] createSymbolicLinkAtURL:symlinkPath withDestinationURL:videoFileUrl error:&error];
+        if (error) {
+            NSLog(@"Error creating symlink: %@", error);
+            return nil;
+        }
+        
+        AVURLAsset *asset = [AVURLAsset URLAssetWithURL:symlinkPath options:nil];
+        
+        // Schedule symlink removal after a short delay
+        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
tmolitor-stud-tu, line 2939: this is complete bullshit and creates all sorts of race conditions!

let the user provide a callback taking the AVURLAsset as argument and remove the symlink after that callback returns

Monal/Classes/HelperTools.m
```diff
@@ -2835,4 +2835,56 @@ +(NSURLSession*) createEphemeralURLSession
     return [NSURLSession sessionWithConfiguration:sessionConfig];
 }
 
++ (nullable AVURLAsset *)createAVURLAssetWithFilePath:(NSString *)filePath
```
tmolitor-stud-tu, line 2902: wrong coding style!
https://github.com/monal-im/Monal/wiki/Code-style

and: _Nullable instead of nullable

and: don't use NSLog but proper DDLogXXX

Monal/Classes/MediaGallery.swift
tmolitor-stud-tu: @matthewrfennell @lissine0 I guess we'll have to fix this ourselves after merging the PR

Monal/Classes/MediaViewer.swift
```diff
@@ -0,0 +1,268 @@
...
+               let mimeType = info["mimeType"] as? String {
+                let fileExtension: String? = nil
+                customPlayer.configurePlayer(filePath: filePath, mimeType: mimeType, fileExtension: fileExtension)
+                isPlayerReady = true
+            }
+        }
+    }
+}
+
+//class CustomAVPlayer: ObservableObject {
```
tmolitor-stud-tu, line 130: why this huge commented out code block??

Monal/Classes/SwiftHelpers.swift
```diff
@@ -440,3 +440,12 @@ public class JingleSDPBridge : NSObject {
         return nil
     }
 }
+
+extension UIImage {
```
tmolitor-stud-tu, line 447: @matthewrfennell @lissine we will have to fix that (write a HelperTools function instead of this swift extension so that it can be used everywhere in monal, not just by 
swift code

or, even better: use the already existing HelperTools function `+(void) addUploadItemPreviewForItem:(NSURL* _Nullable) url provider:(NSItemProvider* _Nullable) provider andPayload:(NSMutableDictionary*) payload withCompletionHandler:(void(^)(NSMutableDictionary* _Nullable)) completion` for generating thumbnails (maybe extend it to properly generate thumbnails for videos if not already supported)

tmolitor-stud-tu (changes requested):

tmolitor-stud-tu:
Cherry-picked and closed now, see here for a tracker of open issues: https://github.com/monal-im/Monal/pull/1221

PR #1189, opened 2024-08-03
Preparation for OmemoKeys ViewController

matthewrfennell:
These are all purely refactors, there should be no change in behaviour.
Main purpose is to simplify `OmemoKeys` where there is logic that can be pushed out of the view, or removing state where we can use functions instead.

### DIFF
--- a/Monal/Classes/MemberList.swift
+++ b/Monal/Classes/MemberList.swift
@@ -30,7 +30,7 @@ struct MemberList: View {
     @StateObject private var overlay = LoadingOverlayState()
 
     init(mucContact: ObservableKVOWrapper<MLContact>) {
-        account = MLXMPPManager.sharedInstance().getConnectedAccount(forID: mucContact.accountId)! as xmpp
+        account = mucContact.obj.account! as xmpp
         _muc = StateObject(wrappedValue:mucContact)
         _ownAffiliation = State(wrappedValue:"none")
         _memberList = State(wrappedValue:OrderedSet<ObservableKVOWrapper<MLContact>>())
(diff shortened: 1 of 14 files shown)


tmolitor-stud-tu (changes requested):
Please change `MLContact` so that it has an `account` property that does what `getAccountBelongingToContact` does. This is cleaner than an aux function in Swift. And it has the nice side effect that it can also be used from objc.

Please update `MLCall` to use this new property of `MLContact` in its `-(xmpp*) account` method.

BTW: `getAccountBelongingToContact` should have been in `SwiftHelpers.swift` instead of `SwiftuiHelpers.swift`.

Monal/Classes/MemberList.swift
```diff
@@ -30,7 +30,7 @@ struct MemberList: View {
     @StateObject private var overlay = LoadingOverlayState()
 
     init(mucContact: ObservableKVOWrapper<MLContact>) {
-        account = MLXMPPManager.sharedInstance().getConnectedAccount(forID: mucContact.accountId)! as xmpp
+        account = mucContact.obj.account! as xmpp
```
tmolitor-stud-tu, line 33: You only have to use `.obj` if you are accessing a method of an `ObservableKVOWrapper` wrapped objc object, accessing a property of such an object (without n`.obj`) works just fine. And it has the additional benefit of emitting SwiftUI `ObservableObject` signals on KVO changes of that property in objc land.

tmolitor-stud-tu (approved):
I marked a minor nitpick. I saw you using `.obj` only in `.init` methods, so that's probably fine, but for clarity and to counter copy-paste bugs later on, this should be changed.

I'll merge it nonetheless, you can open a new PR for that minor thing :)

Monal/Classes/MemberList.swift
matthewrfennell: Is there something special I have to do to allow it? When I try this I get "Generic parameter 'T' could not be inferred".

screenshot-1

tmolitor-stud-tu:
You have to cast it for Swift's type system to be able to infer the type of the generic parameter...
An 'as! xmpp' should suffice :)

Am 6. August 2024 09:56:39 MESZ schrieb Matthew Fennell ***@***.***>:

PR #1197, opened 2024-08-06
Add ContentUnavailableView to OmemoKeys

matthewrfennell:
iPhone iOS 18:

iphone-ios18-no-contacts
iphone-ios18-account-disabled

iPhone iOS 16:

iphone-ios-16-no-contacts
iphone-ios-16-account-disabled

iPad iOS 18:

ipad-ios18-no-contacts
ipad-ios18-account-disabled

iPad iOS 16:

ipad-ios16-no-contacts
ipad-ios16-account-disabled

One extra thing I noticed is that there is no back button on the `OmemoKeys` view on the iPad. It's already like that even before this commit (even when there are keys to display), but I'll take a look and see if I can find a way to add it in.

Also - I modified the wording a little, but appreciate that will generate new translations, so if we'd rather stick with the existing wording, we can just drop the last commit.

### DIFF
--- a/Monal/Classes/ContentUnavailableShimView.swift
+++ b/Monal/Classes/ContentUnavailableShimView.swift
@@ -1,0 +1,45 @@
+//
+//  ContentNotAvailableView.swift
+//  Monal
+//
+//  Created by Matthew Fennell <matthew@fennell.dev> on 05/08/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+import SwiftUI
+
+struct ContentUnavailableShimView: View {
+    private var reason: String
+    private var systemImage: String
+    private var description: Text
+
+    init(_ reason: String, systemImage: String, description: Text) {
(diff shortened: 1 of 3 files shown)


tmolitor-stud-tu:
Changing the wording is okay, translators are usually relatively fast and given that this will be part of the long beta period for monal 7.0, wording changes are no problem, too :)

The missing back button also confused me. I guess that must be fixed in `XMPPEdit.m` or we'd leave it that way until the XMPPEdit is replaced by a swiftui implementation anyways :)

matthewrfennell:
> The missing back button also confused me. I guess that must be fixed in XMPPEdit.m or we'd leave it that way until the XMPPEdit is replaced by a swiftui implementation anyways :)

I had a quick look and couldn't find a straightforward way, so personally prefer the latter :)

tmolitor-stud-tu:
okay, I'll merge this then :)

PR #1200, opened 2024-08-06
Remove unnecessary calls to .obj on MLContact

matthewrfennell:
We must explicitly cast MLContact.account to type xmpp?, as Swift struggles to infer this by itself.

### DIFF
--- a/Monal/Classes/ChannelMemberList.swift
+++ b/Monal/Classes/ChannelMemberList.swift
@@ -15,7 +15,7 @@ struct ChannelMemberList: View {
     @State private var participants: OrderedDictionary<String, String>
 
     init(mucContact: ObservableKVOWrapper<MLContact>) {
-        account = mucContact.obj.account! as xmpp
+        account = (mucContact.account as xmpp?)!
         _channel = StateObject(wrappedValue:mucContact)
         _ownAffiliation = State(wrappedValue:"none")
         _participants = State(wrappedValue:OrderedDictionary<String, String>())
--- a/Monal/Classes/ContactDetails.swift
+++ b/Monal/Classes/ContactDetails.swift
@@ -35,7 +35,7 @@ struct ContactDetails: View {
     init(delegate: SheetDismisserProtocol, contact: ObservableKVOWrapper<MLContact>) {
         self.delegate = delegate
         _contact = StateObject(wrappedValue: contact)
-        self.account = contact.obj.account!
+        self.account = (contact.account as xmpp?)!
     }
 
     private func updateRoleAndAffiliation() {
--- a/Monal/Classes/EditGroupSubject.swift
+++ b/Monal/Classes/EditGroupSubject.swift
@@ -19,7 +19,7 @@ struct EditGroupSubject: View {
         
         _subject = State(wrappedValue: contact.obj.groupSubject)
         _contact = StateObject(wrappedValue: contact)
-        self.account = contact.obj.account! as xmpp
+        self.account = (contact.account as xmpp?)!
     }
 
     var body: some View {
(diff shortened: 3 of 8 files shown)


tmolitor-stud-tu (changes requested):
why not use `contact.account as! xmpp`? That should work, too, no?

Monal/Classes/OmemoQrCodeView.swift
```diff
@@ -30,7 +30,7 @@ struct OmemoQrCodeView: View {
     init(contact: ObservableKVOWrapper<MLContact>)
     {
         self.jid = contact.obj.contactJid
-        if let account = contact.obj.account {
+        if let account = contact.account as xmpp? {
```
tmolitor-stud-tu, line 33:
```suggestion
        if let account = contact.account as? xmpp {
```

Monal/Classes/OmemoKeys.swift
```diff
@@ -292,7 +292,7 @@ struct OmemoKeys: View {
         self.viewContact = contact
 
         if let contact = contact {
-            if let account = contact.obj.account {
+            if let account = contact.account as xmpp? {
```
tmolitor-stud-tu, line 295: same as below

Monal/Classes/MemberList.swift
```diff
@@ -30,7 +30,7 @@ struct MemberList: View {
     @StateObject private var overlay = LoadingOverlayState()
 
     init(mucContact: ObservableKVOWrapper<MLContact>) {
-        account = mucContact.obj.account! as xmpp
+        account = (mucContact.account as xmpp?)!
```
tmolitor-stud-tu, line 33: same as above

Monal/Classes/EditGroupSubject.swift
```diff
@@ -19,7 +19,7 @@ struct EditGroupSubject: View {
         
         _subject = State(wrappedValue: contact.obj.groupSubject)
         _contact = StateObject(wrappedValue: contact)
-        self.account = contact.obj.account! as xmpp
+        self.account = (contact.account as xmpp?)!
```
tmolitor-stud-tu, line 22: same as above

Monal/Classes/ContactDetails.swift
```diff
@@ -35,7 +35,7 @@ struct ContactDetails: View {
     init(delegate: SheetDismisserProtocol, contact: ObservableKVOWrapper<MLContact>) {
         self.delegate = delegate
         _contact = StateObject(wrappedValue: contact)
-        self.account = contact.obj.account!
+        self.account = (contact.account as xmpp?)!
```
tmolitor-stud-tu, line 38: same as above

Monal/Classes/ChannelMemberList.swift
```diff
@@ -15,7 +15,7 @@ struct ChannelMemberList: View {
     @State private var participants: OrderedDictionary<String, String>
 
     init(mucContact: ObservableKVOWrapper<MLContact>) {
-        account = mucContact.obj.account! as xmpp
+        account = (mucContact.account as xmpp?)!
```
tmolitor-stud-tu, line 18:
```suggestion
        account = mucContact.account as! xmpp
```

tmolitor-stud-tu (changes requested):

matthewrfennell:
> why not use `contact.account as! xmpp`? That should work, too, no?

I should've mentioned - I tried these but in each case I get "Generic parameter 'T' could not be inferred"

It gets confused because of the `_Nullable` specifier I put in `MLContact`. If I remove that, I can apply these changes, but I think we need the `_Nullable` since it can return `nil`.

tmolitor-stud-tu:
Yes, we should keep the `_Nullable`...what about `as! xmpp?`?

tmolitor-stud-tu:
Ah: and we should use an `if let` unless we can be 100% sure that the account can never be nil...

Issue #1094, opened 2024-06-25
[Feature]: Rewrite contact list using SwiftUI


tmolitor-stud-tu:
### Precondition

- [X] I understood that this template is only for feature requests and not for bug reports

### Cross-checked related issues?

- [X] I have cross-checked this overview https://github.com/monal-im/Monal/issues/322 as well as filtered for related labels https://github.com/monal-im/Monal/labels

### Describe your feature

- [x] Rewrite `ContactsViewController.m` using SwiftUI and our new `ContactEntry` SwiftUI view
- [ ] Add support for roster groups as tags to `ContactEntry`, will fix #121

tmolitor-stud-tu:
@matthewrfennell despite stating otherwise in the roster groups issue, we should add support for setting/changing roster groups, too...could be a simple button beneath the list of roster groups in the contact details leading to a text entry box and a "save/cancel" button...possibly using RichAlert to display that edit window.

Issue #121, opened 2018-09-27
[feature request] support for user groups


zippydan:
I'm running a private OpenFire server with many users divided into defined groups based on departments.

I've used the following XMPP clients that all support displaying the user list with users organized by groups:

Windows:
Spark

Android:
Xabber

macOS:
Messages (native macOS app)

iOS:
JabberB

Is there any way that Monal could also support this standard? (I assume it is a standard.)
Right now, Monal is just showing me an overwhelming list of online users without any way to easily find the person I need.

P.S. I'm looking for a new macOS client now that Apple in their infinite wisdom decided to drop XMPP from the latest version of Messages in Mojave.  :(  Off to try Adium now, but I much prefer the seemingly overall less bloated approach of your app.

benjaminbischoff:
One point as an iOS user, I think if XMPP hat the plan to be used from many people, on iOS it should not be actived by default! :)

zippydan:
Don't really care if it is default on or off, as long as there is an option to show groups.

In my opinion, it should default to on, as the XMPP server admin decides whether there are groups or not, and if he decides the users should be organized by groups, there is usually a good business or organizational reason for doing so.  In our case, we have so many users that the list becomes overwhelming and unwieldy without them.

zippydan:
Any chance this will be added?

Echolon:
Is that related to a specific XEP actually?

@zippydan I think that can be added one day, however I think Monal is fighting some more important fronts atm. But we will keep that open. Try Siskin or Beagle in the mean time. Converse.js in the webbrowser does support it I think

zippydan:
I just want to confirm that the feature request here matches my intentions.  I'm a little hung up on the word "create" since the user groups I use are "created" and defined SERVER side.  The client then simply supports this central organization and DISPLAYS the server-defined groups.  So, maybe this is just a misunderstanding on my part or a poor choice of words, but what I'm asking for would NOT be the ability to "create" user groups on the CLIENT side.

I could see someone wishing they could organize users on their local client, so that might be a different feature request.  It's not relevant to my own use case and is not what I was originally requesting.

If we're on the same page then just ignore this.  I just thought it would be better to clarify just in case.

Echolon:
@zippydan okay, sorry. the I mistook it. May you try to edit the title and first description again?

Echolon:
And just to say the prios in Monal are at the moment fighting bugs and some other things. So no expectations when this might be implemented

[...]

PR #1205, opened 2024-08-12
SwiftUI contact list

matthewrfennell:
# Introduction

This PR implements the first part of #1094

To do in (if it's ok) future PRs:

- Use `ObservableKVOWrapper<MLContact>` everywhere for live updates
- Implement notification badge for `ContactEntry` to match the old contacts screen
- Delete unused code from `ContactsViewController.m` etc.
- #121

# Videos

## iPhone

https://github.com/user-attachments/assets/60c4f99f-fb5d-4bab-ba37-6a3662573d6e

## iPad

At 2:22, the contact gets deleted from another device.

https://github.com/user-attachments/assets/990247f4-475e-41a4-99ad-f22e6f260401

## Animated notification badge

Just before raising this PR, I noticed the contact request notification badge wasn't animated.
I didn't want to re-record everything, so here's a recording of just that animation.

https://github.com/user-attachments/assets/5bafecbe-fab8-4e68-87a1-d9216fce8ea8

### DIFF
--- a/Monal/Classes/ContactsView.swift
+++ b/Monal/Classes/ContactsView.swift
@@ -108,6 +108,11 @@ struct ContactsView: View {
     private var contactList: [MLContact] {
         return contacts.contacts
             .filter(ContactsView.shouldDisplayContact)
+            .sorted { ContactsView.sortingCriteria($0) < ContactsView.sortingCriteria($1) }
+    }
+
+    private static func sortingCriteria(_ contact: MLContact) -> (String, String) {
+        return (contact.contactDisplayName.lowercased(), contact.contactJid.lowercased())
     }
 
     init(contacts: Contacts, delegate: SheetDismisserProtocol, dismissWithContact: @escaping (MLContact) -> ()) {
--- a/Monal/Classes/ContactsView.swift
+++ b/Monal/Classes/ContactsView.swift
@@ -1,0 +1,171 @@
+//
+//  ContactsView.swift
+//  Monal
+//
+//  Created by Matthew Fennell <matthew@fennell.dev> on 10/08/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+import SwiftUI
+
+struct ContactViewEntry: View {
+    private let contact: MLContact
+    @Binding private var selectedContactForContactDetails: ObservableKVOWrapper<MLContact>?
+    private let dismissWithContact: (MLContact) -> ()
+
+    @State private var shouldPresentRemoveContactAlert: Bool = false
+
+    private var removeContactButtonText: String {
+        if (!isDeletable) {
+            return "Cannot delete notes to self"
+        }
+        return contact.isGroup ? "Remove Conversation" : "Remove Contact"
+    }
+
+    private var removeContactConfirmationTitle: String {
+        contact.isGroup ? "Leave this converstion?" : "Remove \(contact.contactJid) from contacts?"
+    }
+
+    private var removeContactConfirmationDetail: String {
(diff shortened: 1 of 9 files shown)


Monal/Classes/ContactsView.swift
```diff
@@ -0,0 +1,217 @@
...
+                                .tint(.blue)
+                                .imageScale(.large)
+                        }
+                        .accessibilityLabel("Open contact details")
+                    }
+                }
+            }
+        }
+        .swipeActions(allowsFullSwipe: false) {
+            // We do not use a Button with destructive role here as we would like to display the confirmation dialog first.
```
tmolitor-stud-tu, line 71: That's very nice! could you add that behaviour to the group members list, too?

Monal/Classes/ContactsView.swift
```diff
@@ -0,0 +1,217 @@
...
+            }
+        }
+        .sheet(isPresented: $createGroupIsPresented) {
+            NavigationStack {
+                CreateGroupMenu(delegate: SheetDismisserProtocol())
+                    .modifier(Dismissable(delegate: delegate))
+            }
+        }
+        .sheet(isPresented: $addContactIsPresented) {
+            NavigationStack {
```
tmolitor-stud-tu: Would be nicer to not open a new navigation stack here, but use the current one (that would look nice even on ipad/macos, I guess) and let the user return to the contact list by using the back button.

tmolitor-stud-tu (changes requested):

matthewrfennell:
# Popover

Here's how it looks now:

https://github.com/user-attachments/assets/d2aa2eff-910a-41ba-8767-ef528ab720a5

There is one edge case: if the view size changes from regular to compact (or vice versa) while the view is open, the button doesn't get added or removed:

https://github.com/user-attachments/assets/dee7262e-f553-42d2-9b7c-7aac94b39d6c

I'm looking into this now, but it might take some time to work out. It may be easier to wait until the ActiveChatsViewController is also SwiftUI.

# Group list swipe action

Swipe action now doesn't allow full swipe, and users joining and leaving are now animated (presence changes are not animated yet).

https://github.com/user-attachments/assets/f1b3ab9c-36d7-4b7e-95f3-3fb93405b974

Here's how the sheet looks on the iPhone:

group-swipe-action

(sheet looks the same as before in both cases, just wanted to show it)

tmolitor-stud-tu:
Regarding the back button stuff: have a look at `MLSplitViewDelegate.m`.

This should allow you to trigger an event if the split view type (compact/expanded) changes and react accordingly in swiftui.

It will allow you to remove the code passing in the state (`horizontalSizeClass`, `shouldDisplayBackButton` etc.) all the way from active chats down to the swiftui view, too :)

matthewrfennell:
GitHub is getting stuck uploading the video. So here's a blurry gif instead :)

https://github.com/user-attachments/assets/9af4f125-b06f-4377-a179-231c8479e18b

tmolitor-stud-tu:
I feel like `shouldAlwaysDisplayBackButton`isn't needed anymore, right?

Is the `SizeClassWrapper` needed? Wouldn't it be better to just add that `horizontal` property to `ActiveChats` directly (renamed to `horizontalSizeClass`, of course)

matthewrfennell:
Here's how the `ContactDetails` view dismiss button looks now:

https://github.com/user-attachments/assets/6c53de4a-93bd-4b17-b6bd-4e048652fc53

Monal/Classes/ContactsView.swift
tmolitor-stud-tu: Don't forget this one :)

Issue #1188, opened 2024-08-03
[Feature]: Streamline code


tmolitor-stud-tu:
- [x] The current code has different names for the same thing: `isGroup` in `MLContact`, but `isMuc` in `MLMessage`. Use `isMuc` everywhere.
- [x] `mucType` is named same in both files, but is a `NSString`. Make this an enum having the values `channel` and `group` (and unknown, if the db has the value NULL for this information which is needed to trigger a proper update, see 4ac649597f31187cd2fb663d43cdd9d9dc84dd29).
- [x] Rename `connectedAccounts` (and `connectedAccount`) to `enabledAccounts` (`enabledAccount`) everywhere
- [x] Use `accountID` as `NSNumber` everywhere (not `accountId` or `accountNo` and not as `NSString` either).

tmolitor-stud-tu:
@lissine0 This is complete, right?

lissine0:
Yes it is complete (though someone may want to switch from String constants to enums at some point)

PR #1208, opened 2024-08-17
Streamline code

lissine0:
This PR implements everything mentioned in #1188
But, instead of changing mucType to enum, I made the values it takes String constants.
I also changed MUC roles and affiliations to String constants


Monal/Classes/MLConstants.h
```diff
@@ -47,6 +47,9 @@ static const DDLogLevel ddLogLevel = LOG_LEVEL_STDOUT;
 
 #define kMonalKeychainName @"Monal"
 
+#define mucTypeGroup @"group"
```
tmolitor-stud-tu: Define this in the block starting at line ~143

Monal/Classes/ContactDetails.swift
```diff
@@ -202,15 +202,15 @@ struct ContactDetails: View {
...
+                    if !contact.isMuc, let statusMessage = contact.statusMessage as String?, statusMessage.count > 0 {
                         VStack {
                             Text("Status message:")
                             Text(contact.statusMessage as String)
                                 .fixedSize(horizontal: false, vertical: true)
                         }
                     }
                     
-                    if contact.isGroup && ((contact.groupSubject as String).count > 0 || ownRole == "moderator") {
+                    if contact.isMuc && ((contact.groupSubject as String).count > 0 || ownRole == "moderator") {
```
tmolitor-stud-tu: Roles should be made string constants, too. At least those we are using in the code. Same with affiliations ('owner', 'admin', 'member' etc.).

Monal/Classes/SwiftHelpers.swift
```diff
@@ -24,6 +24,8 @@ let kMonalOpenURL = HelperTools.getObjcDefinedValue(.kMonalOpenURL)
 let kBackgroundProcessingTask = HelperTools.getObjcDefinedValue(.kBackgroundProcessingTask)
 let kBackgroundRefreshingTask = HelperTools.getObjcDefinedValue(.kBackgroundRefreshingTask)
 let kMonalKeychainName = HelperTools.getObjcDefinedValue(.kMonalKeychainName)
+let mucTypeGroup = HelperTools.getObjcDefinedValue(.mucTypeGroup)
```
tmolitor-stud-tu: Should also have a `k` prefix here.

Monal/Classes/HelperTools.m
```diff
@@ -565,6 +565,8 @@ +(id) getObjcDefinedValue:(MLDefinedIdentifier) identifier
         case MLDefinedIdentifier_kBackgroundProcessingTask: return kBackgroundProcessingTask; break;
         case MLDefinedIdentifier_kBackgroundRefreshingTask: return kBackgroundRefreshingTask; break;
         case MLDefinedIdentifier_kMonalKeychainName: return kMonalKeychainName; break;
+        case MLDefinedIdentifier_mucTypeGroup: return mucTypeGroup; break;
```
tmolitor-stud-tu: `k` prefix ;)

tmolitor-stud-tu (changes requested):
Didn't we settle with `accountId` instead of `accountID`? I'm a bit unsure because we switched that back and forth while discussing. It could well be that I'm mistaken, of course.

There are other constants (muc affiliations and muc roles) currently hardcoded as strings throughout the code. These ones should be mase string constants, too (sorry I did not mention this earlier, I just catched that one while reviewing your pr).

lissine0:
> Didn't we settle with `accountId` instead of `accountID`? I'm a bit unsure because we switched that back and forth while discussing. It could well be that I'm mistaken, of course.

We did when we found out that `accountId` is used more in the codebase than `accountID`
But I was curious what Apple uses. And I learned that in their frameworks (Foundation, SwiftUI), >90% of the time they use Identifier. But when they don't, they use ID.
Examples: [1](https://developer.apple.com/documentation/foundation/nsdictionary/1412281-fileowneraccountid) [2](https://developer.apple.com/documentation/swiftui/scrollposition/viewid) [3](https://developer.apple.com/documentation/swiftui/tabviewcustomization/subscript(sectionid:)) [4](https://developer.apple.com/documentation/foundation/nsxmlparser/1411917-systemid?language=objc) [5](https://developer.apple.com/documentation/foundation/nsxmlparser/1411917-systemid?language=objc)

However, I also discovered that in some of their WWDC swift code examples, they used [bookId](https://developer.apple.com/videos/play/wwdc2022/10061/) and [recipeId](https://developer.apple.com/videos/play/wwdc2022/10054/), and in some they used [uniqueID](https://developer.apple.com/videos/play/wwdc2024/10185/).
But bookId was a local closure parameter, and recipeID was the argument to the defintion of a subscript.
I guess they used fooId for local variables, and barID for anything viewable from outside the class.
But I still decided to follow the naming practice of Apple's official Frameworks, and to use the same for everything.

I have no strong opinion here. If you want I'll switch to `accountId`

tmolitor-stud-tu:
Okay, lets use somethingID then. Your reasoning makes sense here :)

Am 20. August 2024 23:55:19 MESZ schrieb lissine0 ***@***.***>:

tmolitor-stud-tu:
I will wait for the merge of Vaidik's media gallery before merging this.

I think you'll have to add some more commits to rename stuff after that Pr got merged. I'll merge this once that additional renaming is done, too.

tmolitor-stud-tu:
The gallery is merged now. Let me know once you fixed the conflicts and did the additional renaming and I'll merge the PR :)

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -110,19 +110,19 @@ func promisifyMucAction(account: xmpp, mucJid: String, action: @escaping () thro
...
+        } else if affiliation == kMucAffiliationMember {
             return NSLocalizedString("Member", comment:"muc affiliation")
-        } else if affiliation == "none" {
+        } else if affiliation == kMucAffiliationNone {
             return NSLocalizedString("Participant", comment:"muc affiliation")
-        } else if affiliation == "outcast" {
+        } else if affiliation == kMucAffiliationOutcast {
             return NSLocalizedString("Blocked", comment:"muc affiliation")
-        } else if affiliation == "profile" {
+        } else if affiliation == kMucAffiliationProfile {
```
tmolitor-stud-tu: Could you rename "profile" and "reinvite" to something making clear that this is not a normal XEP-0045 affiliation? Maybe `kMucActionShowProfile` and `kMucActionReinvite`

tmolitor-stud-tu (changes requested):
one small thing :)

PR #1212, opened 2024-08-18
sort MUC participants in a case-insensitive manner

lissine0:
match other clients in this regard, so users have a consistent experience.


tmolitor-stud-tu:
perfect, thanks! :)

PR #1216, opened 2024-08-23
Improve the discovery of room names

lissine0:
Make the discovery of a room's name more reliable by getting it from the identity element, which MUST be returned by the MUC service, instead of relying on muc#roomconfig_roomname, which MAY be returned.

see https://xmpp.org/extensions/xep-0045.html#disco-roominfo

> The room MUST return its identity

> A chatroom MAY return more detailed information in its disco#info response using Service Discovery Extensions

> any field defined for the muc#roomconfig FORM_TYPE can be included in the extended service discovery fields


tmolitor-stud-tu:
Good catch, thanks! :)

PR #1218, opened 2024-08-27
Set monalGreen as the app's accent color

lissine0:
Pick monalGreen as accent color for the entire app. This means that all the navigation UI elements, and most buttons would be colored in it (instead of our current mix between monalGreen and blue)
Also, it adapts to dark mode on its own. (two variants are provided, and SwiftUI selects the relevant one based on context)
In addition, macOS users will be able to customize the accent color from their OS settings. But by default they get monalGreen. (previously they could also customize it, but some UI elements didn't get overrided by their choice)

Note that I did not apply the accentColor to the Rich Alert's buttons, as it wasn't applied before, and I didn't know if you want it.

I also added a custom button style, that's used mostly in the Quicksy registration (but also the Add account button, and in Onboarding)
Be sure to check out its look in dark & light mode, when it's disabled and enabled.
(I made the text of the button only bold when it's enabled, and I changed the text's color from always white, to matching the background color, in an attempt to improve constrast in dark mode)

How the button style looks in various pages of the Quicksy registration:
(Note: some of these may be blurry)

 

On a semi-related note, in the ContactDetails, there's an option to select when you should receive notifications (always, on mention, and never)
A ring bell gets colored accordingly in each case.
But the entire list entry is by default colored in the accentColor, because the entire entry is a button that's not a navigation link.

I'm not sure how we should do in this case with regard to stying. Should the color of the list entry match the color of the ring bell? Or would a red color imply something rather strong to the user? (i.e. it might give the idea it's dangerous or destructive).
We could also change it to the primary color; or keep it as it is now.
(side question: when it is set to on-mention, should it be black or in the accent color?)

There's also another entry below it that's colored in the accent Color ("show shared Media and Files")
Just like the other one, It is in the accent color by default because it's a button that's not a navigation link.
I think it should be in primary color like the other entries, but I didn't modify it.


lissine0:
For the sake of comparison, here's how the button style looks like with white text in dark mode




(I thought it a little hard on the eyes)

tmolitor-stud-tu:
I like the new `MonalProminentButtonStyle`very much :) And the design looks good, too :)

Two questions:
- could you use the 2 green values of the quicksy icon as the two monalGreen values for the quicksy build flavor
- what do you mean by the richAlert buttons you did not change? Do you have a screenshot of these buttons?

tmolitor-stud-tu:
See here for the 2 quicksy colors (the 2 green colors in the background): https://github.com/monal-im/Monal/blob/develop/Monal/Images.xcassets/QuicksyAppLogo.imageset/Quicksy-ios-1024.png

tmolitor-stud-tu:
Regarding your other questions: I think the bell icon should be colored, but the text beneath it should be primary. The icon should be green/black/red (always/mention/never)...but I'd use normal green like the one used by ios for the switch below.

regarding the media button: that one should be primary, too.

regarding your last dark mode screenshot: why is the text of that button in white? shouldn't it be in black like in your other quicksy screenshots?

lissine0:
> I like the new `MonalProminentButtonStyle`very much :) And the design looks good, too :)

glad you liked it :)


> what do you mean by the richAlert buttons you did not change? Do you have a screenshot of these buttons?

I did not change the RichAlert at all. (and its buttons weren't automatically colored by SwiftUI)
Here's how it looks: (the same as before the PR)






> regarding your last dark mode screenshot: why is the text of that button in white? shouldn't it be in black like in your other quicksy screenshots?

That's from an old version of my code. I posted it to demonstrate why I had to change the style of the button. (before this PR, the text on the button was always white, regardless of dark / light mode. It looked fine back then because its background was blue)

lissine0:
FYI, in Quicksy's logo, the left side has a constant color, wherease the right side has variable brightness.
I took the RGB value from the heads of the two people on the right (R: 102, G: 185, B: 105, A: 1.0)

Screenshot

tmolitor-stud-tu:
> That's from an old version of my code. I posted it to demonstrate why I had to change the style of the button.

Ah! Everything alright then :)

The quicksy colors you picked are okay, too :)

btw: sorry for creating some conflicts

Monal/Classes/RegisterAccount.swift
```diff
@@ -404,12 +404,11 @@ struct RegisterAccount: View {
...
                                     .frame(maxWidth: .infinity)
-                                    .padding(9.0)
-                                    .background(Color(UIColor.tertiarySystemFill))
-                                    .foregroundColor(buttonColor)
+                                    .padding(2.0)
                                     .clipShape(RoundedRectangle(cornerRadius: 10, style: .continuous))
                             }
-                            .buttonStyle(BorderlessButtonStyle())
+                            .disabled(registerButtonDisabled)
+                            .buttonStyle(.bordered)
```
tmolitor-stud-tu: Why did you change this to bordered? I believe this make a tap somewhere in the same row trigger the button, even if not tapped onto the button at all. I'm not sure if that's intended here.

Monal/Classes/WelcomeLogIn.swift
```diff
@@ -174,12 +174,11 @@ struct WelcomeLogIn: View {
...
                                         .frame(maxWidth: .infinity)
-                                        .padding(9.0)
-                                        .background(Color(UIColor.tertiarySystemFill))
-                                        .foregroundColor(buttonColor)
+                                        .padding(2.0)
                                         .clipShape(RoundedRectangle(cornerRadius: 10, style: .continuous))
                                 }
-                                .buttonStyle(BorderlessButtonStyle())
+                                .disabled(loginButtonDisabled)
+                                .buttonStyle(.bordered)
```
tmolitor-stud-tu: Why did you change this to bordered? I believe this make a tap somewhere in the same row trigger the button, even if not tapped onto the button at all. I'm not sure if that's intended here.

Monal/shareSheet-iOS/ShareViewController.m
```diff
@@ -51,8 +50,11 @@ +(void) initialize
 -(void) viewDidLoad
 {
     [super viewDidLoad];
+
+    UIColor* monalDarkGreen = [UIColor colorWithRed:20.0/255 green:138.0/255 blue:103.0/255 alpha:1.0f];
```
tmolitor-stud-tu: Why did you duplicate the color codes in here? Please don't do this, if possible.

(On the other hand: if we replace this whole UI with a SwiftUI rewrite soon, this might be okay)

tmolitor-stud-tu (changes requested):
Some small things to change

Monal/Classes/RegisterAccount.swift
lissine0: > I believe this make a tap somewhere in the same row trigger the button, even if not tapped onto the button at all.

I reversed this change.
But just as an FYI, this also happens now / before my PR, but it's not noticeable because there's very small horizontal space next to the button.

Monal/shareSheet-iOS/ShareViewController.m
lissine0: Resolved by adding `shareSheet` to the target membership of `Images.xcassets`

PR #1221, opened 2024-08-29
Fixes for Media Gallery

tmolitor-stud-tu:
Lets use this PR to do bookkepping of open issues in the media gallery.

The following list is not exhaustive, feel free to comment more issues and create pull requests to this branch @matthewrfennell @lissine0. This will allow @thevaidik so easily see what we changed on top of his code.

- [x] Wrong coding style in the objc part
- [x] Don't use NSLog but proper DDLogXXX
- [ ] If I tap the X icon in media viewer I don't get back to the gallery, but back to the contact details
- [ ] Share button of MediaViewer isn't displayed for video files, only for images
- [x] The thumbnail in the gallery does not indicate that this is a video at all
- [x] The thumbnails in the gallery are squished to a square aspect ratio
- [ ] The native video UI VERY BADLY interferes with the MediaViewer UI (pinch, zoom, show close and share buttons etc.)
- [x] Opening an image from the chat view doesn't allow the user to swipe left/right in the mediaviewer like when opening a media file from gallery.
- [ ] Images opened via media gallery should show in a fullscreen view, like when tapping onto an image in the chat
- [ ] Don't use an `extension` for UIImage to calculate a thumbnail (currently defined at the end of `SwiftHelpers.swift`)

### DIFF
--- a/Monal/Classes/ContactDetails.swift
+++ b/Monal/Classes/ContactDetails.swift
@@ -388,15 +388,16 @@ struct ContactDetails: View {
                 let displayName = contact.contactDisplayName as String
                 let sharedUrl = HelperTools.getSharedDocumentsURL(forPathComponents:[accountJid, displayName])
                 if UIApplication.shared.canOpenURL(sharedUrl) && FileManager.default.fileExists(atPath:sharedUrl.path) {
+                    NavigationLink(destination: LazyClosureView{MediaGalleryView(contact: contact.contactJid as String, accountID: contact.accountID)}) {
+                        Text("View Media Gallery")
+                    }
+
                     Button(action: {
                             UIApplication.shared.open(sharedUrl, options:[:])
                     }) {
                         Text("Show shared Media and Files")
                     }
                     .tint(Color.primary)
-                }
-                NavigationLink(destination: LazyClosureView{MediaGalleryView(contact: contact.contactJid as String, accountNo: contact.accountId)}) {
-                    Text("View Media Gallery")
                 }
                 
                 NavigationLink(destination: LazyClosureView(BackgroundSettings(contact:contact))) {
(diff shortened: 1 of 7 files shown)


lissine0:
Potential enhancements:

- [ ] Show the most recent media first:
Currently, the oldest media is shown at the top, and the newest at the bottom
two possible ways to do it:
    * flip the sorting so the most recent is shown at the top
    * load the MediaGallery view at the very bottom, and let users scroll up if they want to see older items
    (this is what the iOS Photos app does. But I'm not sure which method is preferred)
- [ ] Add the possibility to scroll down from the MediaViewer to close it. (This is already possible if you're viewing images from the Media Gallery, but not if you're viewing an image in full-screen from the chat directly)
- [ ] Currently, in the ContactDetails, there are two similar entries `show shared Media and Files` and `View Media Gallery`
The former opens the file manager, in a directory that contains file uploads for that chat (all files, not just media).
Users are tempted to press the `show shared Media and Files` because it's placed before the other option, and because it has easier wording.
Make the `View Media Gallery`option more prominent (for example by placing in on top of the other option and improving its wording)
- [ ] When a chat doesn't have any shared media yet, the `show shared Media and Files` button is hidden, but the `View Media Gallery` button is shown.
Consider hiding it as well.

Feature requests:
- [ ] When using macOS, It would be nice to be able to download the file we're viewing. Now we can only "Save to Photos".
- [ ] The iOS Photos app shows the date of when the photos were created / downloaded, overlayed on top of the media items, and it changes as we scroll up and down. It would be nice if this could be replicated. (showing only the month and year, instead of the full date, may make it easier)
- [ ] Add support for PDF documents (including showing a preview of the first page as a thumbnail, with a way to distinguish it from thumbnails of photos and videos, like a "pdf" overlayed on top of it)

tmolitor-stud-tu:
moved to issue

PR #1222, opened 2024-08-29
Delete old contacts view controller

matthewrfennell:


Monal/Classes/ActiveChatsViewController.m
```diff
@@ -1424,12 +1410,16 @@ -(void) showContacts:(id) sender { // function definition for @selector
 
 -(void) showContacts
 {
-    appendToViewQueue((^(PMKResolver resolve) {
-        // Only segue if at least one account is enabled
-        if(![self showAccountNumberWarningIfNeeded]);
-            [self performSegueWithIdentifier:@"showContacts" sender:self];
-        resolve(nil);
-    }));
+    if([self showAccountNumberWarningIfNeeded]) {
```
tmolitor-stud-tu, line 1413: this should still be `appendToViewQueue` imho

Monal/localization/Base.lproj/Main.storyboard
```diff
@@ -31,52 +31,6 @@
             </objects>
             <point key="canvasLocation" x="-2798" y="-1290"/>
         </scene>
-        <!--Contacts -->
```
tmolitor-stud-tu, line 33: sometimes changing the storyboard has unexpected side effects (especially if done semi automatic using the xcode gui). I'd rather prefer to not clean up the storyboard when migrating things to swiftui but simply delete the complete storyboard once the transition is complete. (or at least once only the active chats view is the last storyboard based view)

or....well....on second thought, maybe a proper cleanup isn't so bad at all...just go ahead then :)

tmolitor-stud-tu (changes requested):

PR #1223, opened 2024-08-29
Always display search bar in ContactsView

matthewrfennell:
When we open the contacts view, our main intent is to find a specific contact to chat with. As a result, the search bar should always be visible, so we can find the contact quickly.

Thank you lissine for catching this change in behaviour compared to the old version.

https://github.com/user-attachments/assets/b7cb87fc-a0d5-4740-894c-881e151317ea

https://github.com/user-attachments/assets/b2921ca7-bbe0-46c6-b877-542b0b24c3be


tmolitor-stud-tu:
Shouldn't the contact list always open scrolled to the top instead of the last scroll position? any thoughts? @lissine0 ?

PR #1224, opened 2024-08-31
Roster Group - Backend

matthewrfennell:
Implement the backend changes for #121.

I did some verification during debugging on the DB and MLContact.

I think the DB logic is a little complex. I considered writing some automated tests for this part, maybe interacting maybe with a fake DB. But this would take some time to explore and might be overkill. If this is something you'd be interested in me looking into, let me know, I'd be happy to give it a go.

# Testing

## Receive stanza

```
RECV Stanza:
<iq id='FC2DF395-1F08-49C3-987C-7308387D292D' to='user1@continuous.nonprod.fennell.dev/Monal-iOS.236ee368' xmlns='jabber:client' type='result'>
  <query xmlns='jabber:iq:roster'>
    …
    <item name='' subscription='both' jid='user2@continuous.nonprod.fennell.dev'>
      <group>Sharks</group>
    </item>
  </query></iq>
```

## DB State after persistence

```SQL
select bl.buddy_name, bg.*, rg.*
from buddylist bl
join buddy_groups bg on bl.buddy_id = bg.buddy_id
join roster_groups rg on bg.roster_group_id = rg.roster_group_id
```
buddy_name | buddy_id | roster_group_id | roster_group_id | account_id | group_name
-- | -- | -- | -- | -- | --
user2@continuous.nonprod.fennell.dev | 73 | 1 | 1 | 2 | Sharks

## Group added to MLContact

(see `_rosterGroups` field)

sharks-user

### DIFF
--- a/Monal/Classes/DataLayer.h
+++ b/Monal/Classes/DataLayer.h
@@ -282,6 +282,7 @@ extern NSString* const kMessageTypeFiletransfer;
 
 -(NSDictionary *) getSubscriptionForContact:(NSString*) contact andAccount:(NSNumber*) accountID;
 -(void) setSubscription:(NSString *)sub andAsk:(NSString*) ask forContact:(NSString*) contact andAccount:(NSNumber*) accountID;
+-(void) setGroups:(NSSet*) groups forContact:(NSString*) contact inAccount:(NSNumber*) accountNo;
 
 #pragma mark History Message Search
 /*
--- a/Monal/Classes/DataLayer.m
+++ b/Monal/Classes/DataLayer.m
@@ -479,16 +479,14 @@ static NSDateFormatter* dbFormatter;
 
         if([results count] == 0)
             return (NSMutableDictionary*)nil;
-        else
-        {
-            NSMutableDictionary* contact = [results[0] mutableCopy];
-            //correctly extract NSDate object or 1970, if last interaction is zero
-            contact[@"lastInteraction"] = nilWrapper([self lastInteractionOfJid:username forAccountID:accountID]);
-            //if we have this muc in our favorites table, this muc is "subscribed"
-            if([self.db executeScalar:@"SELECT room FROM muc_favorites WHERE room=? AND account_id=?;" andArguments:@[username, accountID]] != nil)
-                contact[@"subscription"] = @"both";
-            return contact;
-        }
+
+        NSMutableDictionary* contact = [results[0] mutableCopy];
+        //correctly extract NSDate object or 1970, if last interaction is zero
+        contact[@"lastInteraction"] = nilWrapper([self lastInteractionOfJid:username forAccountID:accountID]);
+        //if we have this muc in our favorites table, this muc is "subscribed"
+        if([self.db executeScalar:@"SELECT room FROM muc_favorites WHERE room=? AND account_id=?;" andArguments:@[username, accountID]] != nil)
+            contact[@"subscription"] = @"both";
+        return contact;
     }];
 }
 
--- a/Monal/Classes/DataLayer.m
+++ b/Monal/Classes/DataLayer.m
@@ -819,6 +819,38 @@ static NSDateFormatter* dbFormatter;
     }];
 }
 
+/**
+ Ensures that the given groups are the only ones persisted for the given contact
+
+ This function updates the groups in the database to match the groups passed to the function. This means (for instance) that passing an empty groups set will delete all the groups for a user.
+ */
+-(void) setGroups:(NSSet*) groups forContact:(NSString*) contact inAccount:(NSNumber*) accountNo
+{
+    if(groups == nil || contact == nil || accountNo == nil)
+        return;
+
+    NSSet* validGroups = [groups filteredSetUsingPredicate:[NSPredicate predicateWithFormat:@"SELF.length > 0"]];
+
+    if([validGroups count] < [groups count]) {
+        DDLogWarn(@"Refusing to persist group(s) with empty name for contact %@", contact);
+    }
+
+    DDLogVerbose(@"For contact %@ and account %@, intend to persist these groups: %@", contact, accountNo, validGroups);
+
+    NSString* deleteBuddyGroups = @"DELETE FROM buddy_groups \
+        WHERE buddy_id IN ( \
+            SELECT buddy_id FROM buddylist WHERE buddy_name=? AND account_id=? \
+        );";
+    NSString* saveBuddyGroups = @"INSERT OR IGNORE INTO buddy_groups ('buddy_id', 'group_name') \
+        SELECT buddy_id, ? FROM buddylist WHERE buddy_name=? AND account_id=?;";
+
+    [self.db voidWriteTransaction:^{
+        [self.db executeNonQuery:deleteBuddyGroups andArguments:@[contact, accountNo]];
+        for(NSString* group in groups) {
+            [self.db executeNonQuery:saveBuddyGroups andArguments:@[group, contact, accountNo]];
+        }
+    }];
+}
 
 
 #pragma mark Contact info
--- a/Monal/Classes/DataLayer.m
+++ b/Monal/Classes/DataLayer.m
@@ -461,7 +461,7 @@ static NSDateFormatter* dbFormatter;
         return nil;
 
     return [self.db idReadTransaction:^{
-        NSArray* results = [self.db executeReader:@"SELECT b.buddy_name, state, status, b.full_name, b.nick_name, Muc, muc_subject, muc_type, muc_nick, mentionOnly, b.account_id, 0 AS 'count', subscription, ask, IFNULL(pinned, 0) AS 'pinned', blocked, encrypt, muted, \
+        NSArray* results = [self.db executeReader:@"SELECT b.buddy_id, b.buddy_name, state, status, b.full_name, b.nick_name, Muc, muc_subject, muc_type, muc_nick, mentionOnly, b.account_id, 0 AS 'count', subscription, ask, IFNULL(pinned, 0) AS 'pinned', blocked, encrypt, muted, \
             CASE \
                 WHEN a.buddy_name IS NOT NULL THEN 1 \
                 ELSE 0 \
@@ -481,6 +481,7 @@ static NSDateFormatter* dbFormatter;
             return (NSMutableDictionary*)nil;
 
         NSMutableDictionary* contact = [results[0] mutableCopy];
+        [contact removeObjectForKey:@"buddy_id"];
         //correctly extract NSDate object or 1970, if last interaction is zero
         contact[@"lastInteraction"] = nilWrapper([self lastInteractionOfJid:username forAccountID:accountID]);
         //if we have this muc in our favorites table, this muc is "subscribed"
--- a/Monal/Classes/DataLayer.m
+++ b/Monal/Classes/DataLayer.m
@@ -481,7 +481,13 @@ static NSDateFormatter* dbFormatter;
             return (NSMutableDictionary*)nil;
 
         NSMutableDictionary* contact = [results[0] mutableCopy];
+        NSNumber* buddyId = contact[@"buddy_id"];
         [contact removeObjectForKey:@"buddy_id"];
+
+        NSArray* groupArray = [self.db executeScalarReader:@"SELECT DISTINCT group_name FROM buddy_groups WHERE buddy_id=?;" andArguments:@[buddyId]];
+        NSSet* groups = [NSSet setWithArray:groupArray];
+        [contact setValue:groups forKey:@"rosterGroups"];
+
         //correctly extract NSDate object or 1970, if last interaction is zero
         contact[@"lastInteraction"] = nilWrapper([self lastInteractionOfJid:username forAccountID:accountID]);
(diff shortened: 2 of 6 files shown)


Monal/Classes/DataLayer.h
```diff
@@ -283,6 +283,23 @@ extern NSString* const kMessageTypeFiletransfer;
 -(NSDictionary *) getSubscriptionForContact:(NSString*) contact andAccount:(NSNumber*) accountID;
 -(void) setSubscription:(NSString *)sub andAsk:(NSString*) ask forContact:(NSString*) contact andAccount:(NSNumber*) accountID;
 
+/**
```
tmolitor-stud-tu: please move the documentation into the method inside DataLayer.m

I usually look into the implementation to check what the method does, so having the documentation there helps a lot visibility wise.

I know that automatic tools generating documentation use the header file to generate it, but since we don't use such a tool (nor properly document every method), I think looking up the implementation and reading the comments provided there is more natural.

Monal/Classes/DataLayer.m
```diff
@@ -479,16 +479,24 @@ -(NSDictionary* _Nullable) contactDictionaryForUsername:(NSString*) username for
...
+        NSMutableDictionary* contact = [results[0] mutableCopy];
+        NSNumber* buddyId = contact[@"buddy_id"];
+        [contact removeObjectForKey:@"buddy_id"];
+
+        NSArray* groupArray = [self.db executeScalarReader:@"SELECT DISTINCT group_name \
+            FROM buddy_groups b JOIN roster_groups r \
+            ON b.roster_group_id = r.roster_group_id \
+            WHERE b.buddy_id = ?;" andArguments:@[buddyId]];
+        NSSet* groups = [NSSet setWithArray:groupArray];
+        [contact setValue:groups forKey:@"groups"];
```
tmolitor-stud-tu: Please use `rosterGroups` as key to properly distinguish this from mucs etc.

Monal/Classes/DataLayer.m
```diff
@@ -819,6 +827,101 @@ -(void) setSubscription:(NSString*)sub andAsk:(NSString*) ask forContact:(NSStri
     }];
 }
 
+// Callers should provide at least a read transaction
```
tmolitor-stud-tu: never write datalayer methods accessing the database without utilizing a proper transaction.

Our `MLSQLite.m` wrapper makes sure transactions can be nested, even though vanilla sqlite doesn't support this. The wrapper even checks for write transactions opened inside read transactions and throws an exception in this case.

Explicitly opening a transaction within every DataLayer method ensures that there is always a proper transaction, regardless of whether the method is called directly from the GUI code or from the receive queue when processing an incoming stanza or calling this method from inside a transaction in another datalayer method.

Monal/Classes/DataLayer.m
```diff
@@ -819,6 +827,101 @@ -(void) setSubscription:(NSString*)sub andAsk:(NSString*) ask forContact:(NSStri
...
+        WHERE R.group_name = ? \
+        AND B.buddy_id = ?";
+    for(NSString* group in groups)
+    {
+        [self.db executeNonQuery:buddyGroupsQuery andArguments:@[group, buddyId]];
+    }
+}
+
+// Callers should provide at least a write transaction
+-(void) deleteStaleRosterGroupEntriesWithNoMappingInBuddyGroups
```
tmolitor-stud-tu: Couldn't this cleanup be realized using an ON DELETE trigger? That seems cleaner to me than an explicit datalayer method that could easily be forgotten

Monal/Classes/DataLayer.m
```diff
@@ -819,6 +827,101 @@ -(void) setSubscription:(NSString*)sub andAsk:(NSString*) ask forContact:(NSStri
...
+{
+    NSString* existingDbGroupsQuery = @"SELECT rg.roster_group_id, rg.group_name \
+        FROM buddy_groups bg JOIN roster_groups rg \
+        ON bg.roster_group_id = rg.roster_group_id \
+        WHERE bg.buddy_id = ?;";
+    return [self.db executeReader:existingDbGroupsQuery andArguments:@[buddyId]];
+}
+
+// Callers should provide at least a write transaction
+-(void) deleteStaleBuddyGroupEntries:(NSSet*) staleGroupIds forBuddy:(NSNumber*) buddyId
```
tmolitor-stud-tu: Couldn't this cleanup be realized using an ON DELETE trigger? That seems cleaner to me than an explicit datalayer method that could easily be forgotten

Monal/Classes/DataLayer.m
```diff
@@ -819,6 +827,101 @@ -(void) setSubscription:(NSString*)sub andAsk:(NSString*) ask forContact:(NSStri
...
+    if(groups == nil || contact == nil || accountNo == nil)
+        return;
+
+    NSSet* validGroups = [groups filteredSetUsingPredicate:[NSPredicate predicateWithFormat:@"SELF.length > 0"]];
+
+    if([validGroups count] < [groups count]) {
+        DDLogWarn(@"Refusing to persist group(s) with empty name for contact %@", contact);
+    }
+
+    DDLogDebug(@"For contact %@ and account %@, intend to persist these groups: %@", contact, accountNo, validGroups);
```
tmolitor-stud-tu: I'm not sure if this log level is intended, maybe verbose instead of debug would match better (verbose is lower than debug, the order being: error, warning, info, debug, verbose)

Monal/Classes/MLContact.m
```diff
@@ -847,6 +848,7 @@ +(MLContact*) contactFromDictionary:(NSDictionary*) dic
     contact.isMuted = [[dic objectForKey:@"muted"] boolValue];
     // initial value comes from db, all other values get updated by our kMonalLastInteractionUpdatedNotice handler
     contact.lastInteractionTime = nilExtractor([dic objectForKey:@"lastInteraction"]);        //no default needed, already done in DataLayer
+    contact.rosterGroups = [dic objectForKey:@"groups"] ?: [NSSet new];
```
tmolitor-stud-tu: You'll have to add this new property to the serialize and unserialize methods, too (NSSecureCoding, NSCoding)

Monal/Classes/MLIQProcessor.m
```diff
@@ -310,9 +310,11 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
         return NO;
     }
     
-    NSArray* rosterList = [iqNode find:@"{jabber:iq:roster}query/item@@"];
-    for(NSMutableDictionary* contact in rosterList)
+    NSArray* rosterList = [iqNode find:@"{jabber:iq:roster}query/item"];
+    for(MLXMLNode* contactNode in rosterList)
     {
+        NSMutableDictionary* contact = [contactNode attributes];
```
tmolitor-stud-tu: I'm not entirely sure, but I think I'd prefer `[contactNode find:@"/@@"]`

tmolitor-stud-tu (changes requested):
Regarding the testcases. I'd love to have a testsuite in monal with tests for the xml parser, database methods etc.

But I think this is something to be postponed until the ui rewrite finishes.

Monal/Classes/DataLayer.m
matthewrfennell: I agree in the other case since it's just keeping the db in sync with itself.
But in this case we're deleting the rows from the db that are no longer in the IQ message.
So I think we have to pass buddy_group changes from the app layer.

tmolitor-stud-tu: Couldn't we simply always delete all groups for the given buddy and then set them again using the given NSSet? e.g. no cleanup but the same code for all updates whether adding new entries or deleting entries.

DELETE FROM [...]
INSERT INTO [...]

Monal/Classes/MLContact.m
```diff
@@ -777,6 +780,7 @@ -(void) updateWithContact:(MLContact*) contact
     updateIfPrimitiveNotEqual(self.isMuted, contact.isMuted);
     //don't update lastInteractionTime from contact, we dynamically update ourselves by handling kMonalLastInteractionUpdatedNotice
     //updateIfIdNotEqual(self.lastInteractionTime, contact.lastInteractionTime);
+    updateIfPrimitiveNotEqual(self.rosterGroups, contact.rosterGroups);
```
tmolitor-stud-tu: That's not a primitive data type but an objc class, use `updateIfIdNotEqual` for these

Monal/Classes/MLContact.m
```diff
@@ -847,6 +851,7 @@ +(MLContact*) contactFromDictionary:(NSDictionary*) dic
     contact.isMuted = [[dic objectForKey:@"muted"] boolValue];
     // initial value comes from db, all other values get updated by our kMonalLastInteractionUpdatedNotice handler
     contact.lastInteractionTime = nilExtractor([dic objectForKey:@"lastInteraction"]);        //no default needed, already done in DataLayer
+    contact.rosterGroups = [dic objectForKey:@"roster_groups"] ?: [NSSet new];
```
tmolitor-stud-tu: Since this dictionary solely comes from the corresponding datalayer method, I'd rather not set a default value here, but crash the app....this would catch errors in the datalayer method much better than using a default value masking possible errors.

what do you think?

Monal/Classes/DataLayer.m
```diff
@@ -819,6 +827,114 @@ -(void) setSubscription:(NSString*)sub andAsk:(NSString*) ask forContact:(NSStri
...
+    DDLogVerbose(@"For contact %@ and account %@, intend to persist these groups: %@", contact, accountNo, validGroups);
+
+    [self.db voidWriteTransaction:^{
+        NSNumber* buddyId = [self buddyIdForContact:contact inAccount:accountNo];
+        DDLogVerbose(@"buddyId for contact %@ and account %@ is: %@", contact, accountNo, buddyId);
+
+        NSArray* existingGroups = [self existingGroupsForBuddyWithId:buddyId];
+        DDLogVerbose(@"Found existing groups: %@ in db for buddyId %@", existingGroups, buddyId);
+
+        NSArray* staleGroups = [existingGroups filteredArrayUsingPredicate:[NSPredicate predicateWithFormat:@"!(SELF['group_name'] IN %@)", validGroups]];
```
tmolitor-stud-tu: use a db trigger, the "delete ad fill again" strategy on roster pushes from the server and proper JOIN and WHERE sqlite clauses to simplify this and ideally reduce it to a small set of 1-3 (longer) sqlite queries doing all the work.

tmolitor-stud-tu (changes requested):

Monal/Classes/DataLayerMigrations.m
```diff
@@ -1091,6 +1091,33 @@ FOREIGN KEY('account_id') REFERENCES 'account'('account_id') ON DELETE CASCADE \
             [db executeNonQuery:@"ALTER TABLE account DROP COLUMN 'supports_sasl2';"];
         }];
         
+        //create tables to allow for storage of roster groups
+        [self updateDB:db withDataLayer:dataLayer toVersion:6.407 withBlock:^{
+            [db executeNonQuery:@"CREATE TABLE 'roster_groups' ( \
+                'roster_group_id' INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT, \
```
matthewrfennell: SQLite [recommends against using](https://www.sqlite.org/autoinc.html) `AUTOINCREMENT` unless we specifically always need the primary key to be monotonically increasing (and never reuse values that were since deleted).

If I replace this with `'roster_group_id' INTEGER PRIMARY KEY`, then SQLite will [automatically alias this](https://www.sqlite.org/lang_createtable.html#rowid) to the `ROWID` which should be faster and allow us to reuse previous IDs.

tmolitor-stud-tu: well, speed isn't that crucial (I doubt we'll have group updates every few seconds).

But we could simplify the db setup by removing the roster_groups table (the one matching group names to group ids) and write the group names directly to the buddy_groups table (making the buddy_id, group _name tuple a primary key).

The additional space needed in this case should be negligible, but it will be way easier to maintain :)

Monal/Classes/MLContact.m
matthewrfennell: Agreed. I'd put the default value here so I wouldn't need to define it in every variant of `makeDummyContact`.

But since all the other default values are specified there, agreed it makes sense to move the `[NSSet new]` there and have a stricter check here.

Monal/Classes/DataLayer.m
```diff
@@ -819,6 +824,48 @@ -(void) setSubscription:(NSString*)sub andAsk:(NSString*) ask forContact:(NSStri
     }];
 }
 
+-(NSNumber*) buddyIdForContact:(NSString*) contact inAccount:(NSNumber*) accountNo
```
tmolitor-stud-tu: Is this still needed as extra method? Wouldn't it be cleaner to join on buddylist in the sql queries in setGroupsforContact:inAccount:?

tmolitor-stud-tu (changes requested):
Only one thing left: use the full powers of sqlite and join tables using WHERE clauses etc.

PR #1225, opened 2024-08-31
Always show back button on mac catalyst

matthewrfennell:
#1205 introduced a regression on MacOS where, once a user opens the contact details view, they cannot close it, since there is no back button and clicking outside of the modal does nothing (unlike on iOS).

This PR works around this by always showing the back button in `AddTopLevelNavigation`. The consequence of this is that we also show the back button in popovers (such as the `ContactsView` popover)

Screenshot 2024-08-31 at 15 59 53
Screenshot 2024-08-31 at 16 00 06


tmolitor-stud-tu:
Wouldn't it be possible to use `#if targetEnvironment(macCatalyst)` to only add the back button on macos but not on ios (iphone/ipad)?

tmolitor-stud-tu:
Okay, forget my comment. You exactly did what I had in mind :)

PR #1228, opened 2024-09-03
Rewrite the ServerDetails view in SwiftUI

lissine0:
Replicate the UI of the UIKit version.


Monal/Classes/MLXMLNode.h
```diff
@@ -33,11 +33,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 -(NSArray*) find:(NSString* _Nonnull) queryString, ... NS_FORMAT_FUNCTION(1, 2);
 -(id _Nullable) findFirst:(NSString* _Nonnull) queryString, ... NS_FORMAT_FUNCTION(1, 2);
+-(id _Nullable) findFirstWithNoVarArgs:(NSString* _Nonnull) queryString;
```
lissine0: I'm open to suggestions for better names to the added functions

tmolitor-stud-tu: `findFirst:`, `find:` and `check:` are all just wrappers for `-(NSArray*) find:(NSString* _Nonnull) queryString arguments:(va_list*) args` and that method should already be exported to swift as per https://developer.apple.com/documentation/swift/stringinterpolationprotocol

--> no need to change the objc side of things, just write 2 wrapper functions as a swift category of `MLXMLNode` (e.g. using the swift keyword `extension`).

tmolitor-stud-tu (changes requested):

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -735,7 +735,14 @@ class SwiftuiInterface : NSObject {
 #endif
         return host
     }
-    
+
+    @objc
+    func makeServerDetailsView(_ xmppAccount: xmpp) -> UIViewController {
```
tmolitor-stud-tu:
```suggestion
    func makeServerDetailsView(for: xmppAccount: xmpp) -> UIViewController {
```

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+    let status: Status
+}
+
+struct ServerDetailsEntry: View {
+    let title: String
+    let description: String
+    let status: Status
+    var color: Color {
+        switch status {
+        case .success:
```
tmolitor-stud-tu: Coding style: indent the cases inside the switch statement

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+            return .clear
+        case .warning:
+            return Color(.serverDetailsEntryWarning)
+        case .error:
+            return Color(.serverDetailsEntryError)
+        }
+    }
+    var body: some View {
+        VStack(alignment: .leading) {
+            Text(LocalizedStringKey(title))
```
tmolitor-stud-tu: I'm not sure if the system will manage to extract the localizable strings this way, use explicit `NSLocalizedString` to make sure that the title and description remain translatable.

Background: XCode has two methods for the extraction of localizable strings. One will compile the whole source tree and collect every usage of `LocalizedStringKey`, the other one only collecting the strings inside `Text()` (and a few others).

We can't use the first one, because it somehow throws an error. Feel free to try to fix this, though :)
See `scripts/updateLocalization.sh`, you can call it like this: `scripts/updateLocalization.sh NOCOMMIT YES` to try the compile variant (use `scripts/updateLocalization.sh NOCOMMIT` to try without compilation.

You'll have to install `bartycrouch` using homebrew first.

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+                }
+            }
+
+            Section(header: Text("These are the channel-binding types your server supports to detect attacks on the TLS layer (used one in blue, orange ones unsupported by Monal).")) {
+                ForEach(getChannelBindingEntryData(xmppAccount: xmppAccount, connection: connection)) { entryData in
+                    ServerDetailsEntry(entryData)
+                }
+            }
+
+        }.navigationTitle(connection.identity.domain)
```
tmolitor-stud-tu:
```suggestion
        }
        .navigationTitle(connection.identity.domain)
```

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+        for (jid, node) in (connection.conferenceServers as! [String: MLXMLNode]) {
+
+            let entry = node.findFirstWithNoVarArgs("identity@@") as! [String: String]
+
+            result.append(EntryData(title: "Server: \(jid)",
+                                     description: "\(entry["name"]!) (type \(entry["type"]!), category \(entry["category"]!))",
+                                     status: entry["type"] == "text" ? .success : .normal))
+        }
+        return result
+    }
```
tmolitor-stud-tu:
```suggestion
    guard connection.conferenceServers.count > 0 else {
        return [EntryData(title: "None",
                           description: "This server does not provide any MUC servers.",
                           status: .error)]
    }

    var result: [EntryData] = []
    for (jid, node) in (connection.conferenceServers as! [String: MLXMLNode]) {
        let entry = node.findFirstWithNoVarArgs("identity@@") as! [String: String]

        result.append(EntryData(title: "Server: \(jid)",
                                 description: "\(entry["name"]!) (type \(entry["type"]!), category \(entry["category"]!))",
                                 status: entry["type"] == "text" ? .success : .normal))
    }
    return result
```

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+}
+
+private func getStunTurnEntryData(connection: MLXMPPConnection) -> [EntryData] {
+    var result: [EntryData] = []
+
+    let stunTurnServers = connection.discoveredStunTurnServers as! [[String: String]]
+    for service in stunTurnServers {
+        var status = Status.normal
+        if (service["type"] == "stun" || service["type"] == "turn") {
```
tmolitor-stud-tu: wouldn't a switch statement with fallthrough cases be more elegant here?

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+        } else {
+            status = .error
+        }
+        result.append(EntryData(title: service["type"] ?? "error",
+                                 description: "\(service["host"]!):\(service["port"]!)",
+                                 status: status))
+    }
+
+    if result.isEmpty {
+    result.append(EntryData(title: "None",
```
tmolitor-stud-tu:
```suggestion
        result.append(EntryData(title: "None",
```

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+                            description: "This server does not provide any STUN / TURN services.",
+                            status: .error))
+    }
+
+    return result
+}
+
+private func getSRVEntryData(xmppAccount: xmpp) -> [EntryData] {
+    if xmppAccount.discoveredServersList.count == 0 {
```
tmolitor-stud-tu: Use a guard statement, too

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+            let prio = srvEntry["priority"] as! Int
+
+            var entryStatus = Status.normal
+
+            // 'connectServer()' has been renamed to 'connect()'
+            if (xmppAccount.connectionProperties.server.connect() == hostname &&
+                xmppAccount.connectionProperties.server.connectPort() == port &&
+                xmppAccount.connectionProperties.server.isDirectTLS() == isSecure) {
+                entryStatus = .success
+                foundCurrentConn = true
```
tmolitor-stud-tu:
```suggestion
            if (xmppAccount.connectionProperties.server.connect() == hostname &&
                    xmppAccount.connectionProperties.server.connectPort() == port &&
                    xmppAccount.connectionProperties.server.isDirectTLS() == isSecure) {
                entryStatus = .success
                foundCurrentConn = true
```

tmolitor-stud-tu: OR:
```suggestion
            if (xmppAccount.connectionProperties.server.connect() == hostname &&
            xmppAccount.connectionProperties.server.connectPort() == port &&
            xmppAccount.connectionProperties.server.isDirectTLS() == isSecure) {
                entryStatus = .success
                foundCurrentConn = true
```

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+}
+
+private func getTLSEntryData(connection: MLXMPPConnection) -> [EntryData] {
+    return [EntryData(title: "TLS 1.2",
+                       description: "Older, slower, but still secure TLS version",
+                       status: connection.tlsVersion == "1.2" ? .success : .normal),
+            EntryData(title: "TLS 1.3",
+                       description: "Newest TLS version which is faster than TLS 1.2",
+                       status: connection.tlsVersion == "1.3" ? .success : .normal)
+            ]
```
tmolitor-stud-tu:
```suggestion
    return [
        EntryData(
            title: "TLS 1.2",
            description: "Older, slower, but still secure TLS version",
            status: connection.tlsVersion == "1.2" ? .success : .normal
        ),
        EntryData(
            title: "TLS 1.3",
            description: "Newest TLS version which is faster than TLS 1.2",
            status: connection.tlsVersion == "1.3" ? .success : .normal
        ),
    ]
```

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+}
+
+private func getSASLEntryData(connection: MLXMPPConnection) -> [EntryData] {
+
+    if connection.saslMethods.count == 0 {
```
tmolitor-stud-tu: Guard statement

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+
+        var result: [EntryData] = []
+        let saslMethods = connection.saslMethods as! [String: Bool]
+
+        for method in saslMethods.keys.sorted() {
+            let used = saslMethods[method]!
+            let supported = (SCRAM.supportedMechanisms(includingChannelBinding: true) as! [String]).contains(method)
+            var description: String
+            switch method {
+            case "PLAIN":
```
tmolitor-stud-tu: Indendation

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,352 @@
...
+        return result
+    }
+
+}
+
+private func getChannelBindingEntryData(xmppAccount: xmpp, connection: MLXMPPConnection) -> [EntryData] {
+    if connection.channelBindingTypes.count == 0 {
+        return [EntryData(title: "None",
+                           description: "This server does not support any modern channel-binding to secure against MITM attacks on the TLS layer.",
+                           status: .error)]
```
tmolitor-stud-tu:
```suggestion
        return [
            EntryData(
                title: "None",
                description: "This server does not support any modern channel-binding to secure against MITM attacks on the TLS layer.",
                status: .error
            )
        ]
```

tmolitor-stud-tu: OR:
```suggestion
        return [EntryData(
            title: "None",
            description: "This server does not support any modern channel-binding to secure against MITM attacks on the TLS layer.",
            status: .error
        )]
```

tmolitor-stud-tu (changes requested):

Monal/Classes/ServerDetails.swift
lissine0: what about:
```
        if xmppAccount.connectionProperties.server.connect() == hostname &&
           xmppAccount.connectionProperties.server.connectPort() == port &&
           xmppAccount.connectionProperties.server.isDirectTLS() == isSecure {
            entryStatus = .success
            foundCurrentConn = true
        } else if !foundCurrentConn {
```
Also, I assume it is for readability here that `xmppAccount.connectionProperties` was used in here instead of `connection` (and consequently that connection wasn't passed as an argument to the function; I followed https://github.com/monal-im/Monal/blob/2ce574b/Monal/Classes/MLServerDetails.m#L219)

tmolitor-stud-tu: What about this one? That's my favorite :)
```
            if (xmppAccount.connectionProperties.server.connect() == hostname &&
                xmppAccount.connectionProperties.server.connectPort() == port &&
                xmppAccount.connectionProperties.server.isDirectTLS() == isSecure
            ) {
                entryStatus = .success
                foundCurrentConn = true
```

Regarding readability: I don't know why `xmppAccount.connectionProperties` was used instead of `connection`. Feel free to change that if you want  :)

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,368 @@
...
+    let link = "[\(linkText)](https://github.com/monal-im/Monal/wiki/Considerations-for-XMPP-server-admins)"
+    return ServerDetailsEntry(EntryData(title: serverName,
+                                        description: "\(description)\n\n\(link)",
+                                        status: .normal))
+}
+
+private func getXEPEntryData(connection: MLXMPPConnection) -> [EntryData] {
+
+    var result: [EntryData] = []
+    result.append(EntryData(title: "XEP-0163 Personal Eventing Protocol",
```
tmolitor-stud-tu: why use `result.append` instead of just listing every entry inside an array literal?

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,368 @@
...
+        ]
+    }
+
+    var result: [EntryData] = []
+    for (jid, node) in (connection.conferenceServers as! [String: MLXMLNode]) {
+        let entry = node.findFirstWithNoVarArgs("identity@@") as! [String: String]
+
+        result.append(EntryData(title: "Server: \(jid)",
+                                description: "\(entry["name"]!) (type \(entry["type"]!), category \(entry["category"]!))",
+                                status: entry["type"] == "text" ? .success : .normal))
```
tmolitor-stud-tu:
```suggestion
        result.append(EntryData(
            title: "Server: \(jid)",
            description: "\(entry["name"]!) (type \(entry["type"]!), category \(entry["category"]!))",
            status: entry["type"] == "text" ? .success : .normal
        ))
```

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,368 @@
...
+    let stunTurnServers = connection.discoveredStunTurnServers as! [[String: String]]
+    for service in stunTurnServers {
+        var status = Status.normal
+        switch(service["type"]) {
+            case "stun", "turn", "stuns", "turns":
+                status = .success
+            default:
+                status = .error
+        }
+        result.append(EntryData(title: service["type"] ?? "error",
```
tmolitor-stud-tu: same here (and multiple other places)

tmolitor-stud-tu (changes requested):

Monal/Classes/ServerDetails.swift
lissine0: Should I match it with the style I picked in https://github.com/monal-im/Monal/pull/1228#discussion_r1742676959 ? (when returning an array literal)

tmolitor-stud-tu: yes :)

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,416 @@
...
+        }
+        .navigationTitle(connection.identity.domain)
+        .listStyle(.grouped)
+    }
+}
+
+
+private func showServerVersionInfoView(connection: MLXMPPConnection) -> some View {
+    let serverVersion = connection.serverVersion
+
```
tmolitor-stud-tu, line 76: why that arbitrary empty line here?

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,416 @@
...
+    let description = String(format: NSLocalizedString("version %@%@", comment: "server details"), serverVersionString, serverPlatform)
+    let linkText = NSLocalizedString("Considerations for Server Administrators", comment: "server details")
+    let link = "[\(linkText)](https://github.com/monal-im/Monal/wiki/Considerations-for-XMPP-server-admins)"
+    return ServerDetailsEntry(EntryData(title: serverName,
+                                        description: "\(description)\n\n\(link)",
+                                        status: .normal))
+}
+
+private func getXEPEntryData(connection: MLXMPPConnection) -> [EntryData] {
+
```
tmolitor-stud-tu: why that empty line?

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,416 @@
...
+            xmppAccount.connectionProperties.server.isDirectTLS() == isSecure
+        ) {
+            entryStatus = .success
+            foundCurrentConn = true
+        } else if !foundCurrentConn {
+            // Set the status of all connections entries that failed to error
+            // discoveredServersList is sorted. Therfore all entries before foundCurrentConn == true have failed
+            entryStatus = .error
+        }
+        result.append(EntryData(title: "Server: \(hostname)",
```
tmolitor-stud-tu: coding style

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,416 @@
...
+        EntryData(
+            title: "TLS 1.3",
+            description: "Newest TLS version which is faster than TLS 1.2",
+            status: connection.tlsVersion == "1.3" ? .success : .normal
+        ),
+    ]
+}
+
+private func getSASLEntryData(connection: MLXMPPConnection) -> [EntryData] {
+
```
tmolitor-stud-tu, line 206: why that empty line?

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,416 @@
...
+        result.append(
+            EntryData(
+                title: "Method: \(method)",
+                description: description,
+                status: used ? .success : (!supported ? .warning : .normal)
+            )
+        )
+    }
+    return result
+
```
tmolitor-stud-tu, line 260: why that empty line?

tmolitor-stud-tu (changes requested):

Monal/Classes/MLXMLNode.h
lissine0: See https://github.com/monal-im/Monal/compare/7b60a85..70faba9

lissine0: Or would you rather we do the computation in objective-C and just use the result in Swift?
I think it depends whether you expect to need check / findFirst in Swift in the future.

Monal/Classes/ServerDetails.swift
lissine0: It's not so arbitrary, because all the following declarations depend on the first one.
But I'll remove it.

Monal/Classes/ServerDetails.swift
tmolitor-stud-tu: This has still to be solved. Why don't you just pass in `Text` views for title and description instead of strings? That would solve it without fixing the compilation problems I already mentioned.

Monal/Classes/MLXMLNode.h
```diff
@@ -33,11 +33,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 -(NSArray*) find:(NSString* _Nonnull) queryString, ... NS_FORMAT_FUNCTION(1, 2);
 -(id _Nullable) findFirst:(NSString* _Nonnull) queryString, ... NS_FORMAT_FUNCTION(1, 2);
+-(id _Nullable) findFirst:(NSString* _Nonnull) queryString :(va_list) args;
```
tmolitor-stud-tu: Didn't we settle for moving all things needing a mlxml query to computed properties and removing all those mlxmlnode swift bridging stuff?

tmolitor-stud-tu (changes requested):

Monal/Classes/MLXMLNode.h
tmolitor-stud-tu: I don't expect swift code to use the MLXMLNode at all. All processing should be kept in the objc land and only the UI should be in swift (or some small helpers not possible to write in objc)

tmolitor-stud-tu: Objc land should expose models backed by db etc. to be used by swift (like the MLContact, MLMessage or MLContactSoftwareVersionInfo models)

Monal/Classes/MLXMPPConnection.h
```diff
@@ -64,6 +64,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 -(id) initWithServer:(MLXMPPServer*) server andIdentity:(MLXMPPIdentity*) identity;
 
+-(BOOL) supportsRosterVersioning;
```
tmolitor-stud-tu: no, I talked about a computed readonly property (having only a getter method), not a plain method

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,418 @@
+//
+//  ServerDetails.swift
+//  Monal
+//
+//  Created by lissine on 3/9/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+enum Status {
```
tmolitor-stud-tu: please make that enum fileprivate, to avoid name clashes

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,418 @@
...
+    case success, normal, warning, error
+}
+
+struct EntryData: Identifiable {
```
tmolitor-stud-tu: please either make this fileprivate, too, or rename it to `ServerDetailsEntryData`

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,418 @@
...
+    let id = UUID()
+    let title: String
+    let description: String
+    let status: Status
+}
+
+struct ServerDetailsEntry: View {
+    let title: String
```
tmolitor-stud-tu: why do you replicate the contents of `EntryData` instead of making `EntryData` a `class` inheriting from `ObservableObject` and using it as `@ObservedObject var`

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,418 @@
...
+
+enum Status {
+    case success, normal, warning, error
+}
+
+struct EntryData: Identifiable {
+    let id = UUID()
+    let title: String
+    let description: String
+    let status: Status
```
tmolitor-stud-tu, line 17: add a computed property named `color` using the `Status` to calculate the color

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,418 @@
...
+}
+
+struct ServerDetailsEntry: View {
+    let title: String
+    let description: String
+    let status: Status
+    var color: Color {
+        switch status {
```
tmolitor-stud-tu, line 26: move that into the computed property above

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,418 @@
...
+                .font(.headline)
+            Text(LocalizedStringKey(description))
+                .font(.caption)
+        }
+        .listRowBackground(color)
+    }
+}
+
+struct ServerDetails: View {
+
```
tmolitor-stud-tu: why that empty line?

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,418 @@
...
+            title: NSLocalizedString("XEP-0474: SASL SCRAM Downgrade Protection", comment: ""),
+            description: NSLocalizedString("This specification provides a way to secure the SASL and SASL2 handshakes against method and channel-binding downgrades.", comment: ""),
+            status: connection.supportsSSDP ? .success : .error
+        ),
+    ]
+    return result
+}
+
+private func getMUCEntryData(connection: MLXMPPConnection) -> [EntryData] {
+    // Yes, conferenceServers is plural, but for now, our connectionProperties only store one single muc server (the first one encountered)
```
tmolitor-stud-tu: are you sure about that (only one single muc server)?

tmolitor-stud-tu: I've posted a screenshot in our dev muc showing the old objc interface listing multiple chatroom servers...

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,418 @@
...
+                title: String(format: NSLocalizedString("Server: %@", comment: ""), entry["jid"] ?? "error"),
+                description: String(format: NSLocalizedString("%@ (type '%@', category '%@')", comment: ""), entry["name"]!, entry["type"]!, entry["category"]!),
+                status: entry["type"] == "text" ? .success : .normal
+            )
+        )
+    }
+    return result
+}
+
+private func getStunTurnEntryData(connection: MLXMPPConnection) -> [EntryData] {
```
tmolitor-stud-tu: why don't you add all these functions to the `ServerDetails` view?

tmolitor-stud-tu (changes requested):

Monal/Classes/ServerDetails.swift
lissine0: I copied the comment from https://github.com/monal-im/Monal/blob/develop/Monal/Classes/MLServerDetails.m#L180
I didn't verify whether or not it was true 😅

Monal/Classes/MLXMPPConnection.h
lissine0: Shall I keep the `conferenceServersInfo` a plain method?
I think doing so may be better for clarity, since there is already a `conferenceServers` property. (especially that one is a dictionary and the other is an array of dictionaries)

Monal/Classes/ServerDetails.swift
lissine0: Hmm, why not `let entryData: EntryData` (and make it a `@State` if you need to)?

tmolitor-stud-tu: That would be possible, too...but structs have some limitations, classes don't have...therefore models should usually be classes

Monal/Classes/MLXMPPConnection.h
tmolitor-stud-tu: well, renaming `conferenceServers` to `conferenceServersDisco` and making `conferenceServersInfo` a computed property, too, would be even cleaner. what do you think?

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,413 @@
...
+                EntryData(
+                    title: NSLocalizedString("None", comment: ""),
+                    description: NSLocalizedString("This server does not have any SRV records in DNS.", comment: ""),
+                    status: .error
+                )
+            ]
+        }
+
+        var result: [EntryData] = []
+
```
tmolitor-stud-tu: Why that empty line between those 2 vars? Even the empty line below is strange, because these 2 vars belong to the for block after them.

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,413 @@
...
+                    title: NSLocalizedString("None", comment: ""),
+                    description: NSLocalizedString("This server does not support modern SASL2 authentication.", comment: ""),
+                    status: .error
+                )
+            ]
+        }
+
+        var result: [EntryData] = []
+        let saslMethods = connection.saslMethods as! [String: Bool]
+
```
tmolitor-stud-tu: Why that empty line? The two vars belong to the for block below them.

Monal/Classes/ServerDetails.swift
```diff
@@ -0,0 +1,413 @@
...
+                    description: NSLocalizedString("This server does not support any modern channel-binding to secure against MITM attacks on the TLS layer.", comment: ""),
+                    status: .error
+                )
+            ]
+        }
+
+        var result: [EntryData] = []
+        let channelBindingTypes = connection.channelBindingTypes as! [String: Bool]
+        let supportedChannelBindingTypes = xmppAccount.supportedChannelBindingTypes as! [String]
+
```
tmolitor-stud-tu: Here again.

tmolitor-stud-tu (changes requested):
These are the last change requests, I guess :)

Monal/Classes/MLXMPPConnection.h
lissine0: How about we keep `conferenceServers` as is, and we rename `conferenceServersInfo` to `conferenceServersIdentity` ?

lissine0: or `conferenceServerIdentities`

PR #1236, opened 2024-09-10
Rewrite the Blocked Users view in SwiftUI

lissine0:


Monal/Classes/BlockedUsers.swift
```diff
@@ -0,0 +1,104 @@
...
+//  BlockedUsers.swift
+//  Monal
+//
+//  Created by lissine on 10/9/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+struct BlockedUsers: View {
+    let xmppAccount: xmpp
+    static private let jidFaultyPattern = "^([^@]+@)?[^/\\n]+(\\..{2,})?(/.+)?$"
```
lissine0: Side question: I followed the existing naming convention, but why do you call it jidFaultyPattern instead of jidPattern?

Monal/Classes/BlockedUsers.swift
```diff
@@ -0,0 +1,104 @@
...
+                hideLoadingOverlay(overlay)
+            }
+
+        }
+        .toolbar {
+            ToolbarItem(placement: .navigationBarTrailing) {
+                Button(action: {
+                    guard blockingSupported else {
+                        showBlockingUnsupportedAlert = true
+                        return
```
lissine0: Would you like to also show this alert on BlockedUsers view appear?
Or would that cause problems if the discovery isn't done yet?

Monal/Classes/BlockedUsers.swift
tmolitor-stud-tu: That was @FriedrichAltheide ;) No idea, but in AddContactMenu.swift it's called jidFaultyPattern, too. Maybe some copy/paste.

You can rename it to jidPattern if you want :)

Monal/Classes/BlockedUsers.swift
tmolitor-stud-tu: I'd show a placeholder view if blocking isn't supported and remove that alert altogether. You can use this to check if the disco was already done and show a "loading..." overlay otherwise: `account.accountState >= kStateBound && account.connectionProperties.accountDiscoDone` (this is objc code).
you can use the `kMonalAccountDiscoDone` notification to remove the loading overlay once the disco was done and then either show the "not supported" placeholder or the normal ui.

tmolitor-stud-tu:
I think the proper fix for nullability coming from the db could be to change `MLSQLite` to don't add something to its return dictionary if it is of type `NSNull`. Accessing a non-existent key of a dictionary automatically returns `nil`.

And then check every db table definition for colums that could be `NULL` and check if the corresponding code in monal properly handles `nil` return values.

Only caveat: writing `NULL` to the db still needs `NSNull` because we need to tell the `MLSQLite` implementation what fields to set (even if it is `NULL`).
Changing the behavior for returning db fields to not return `NSNull` would make the interface for reading and writing to the db different.

@matthewrfennell what do you think about this? context: currently the `MLSQLite` db abstraction returns an `NSNull` if a column was `NULL`.

I'm unsure what would be the best here.

tmolitor-stud-tu:
this code in the db layer can be rewritten using the `nilWrapper()`

```
                parsedJid[@"node"] ? parsedJid[@"node"] : [NSNull null],
                parsedJid[@"host"] ? parsedJid[@"host"] : [NSNull null],
                parsedJid[@"resource"] ? parsedJid[@"resource"] : [NSNull null],
```

lissine0:
> I think the proper fix for nullability coming from the db could be to change `MLSQLite` to don't add something to its return dictionary if it is of type `NSNull`. Accessing a non-existent key of a dictionary automatically returns `nil`.

Note: I checked with the debugger, and the dictionary bridged by Swift didn't include any of the (key, value) pairs whose value is null

Monal/Classes/BlockedUsers.swift
lissine0: I was talking about AddContactMenu, WelcomeLogIn, and RegisterAccount. (faultyPattern introduced in e7ea977)
MLBlockedUsersTableViewController didn't use a regular expression. Instead, it used HelperTools.splitJid and checked that the host != nil (which was incorrect. Should've been a check that host != `@""`

Monal/Classes/XMPPEdit.m
```diff
@@ -892,15 +897,7 @@ -(void) tableView:(UITableView*) tableView accessoryButtonTappedForRowWithIndexP
 
 -(void) prepareForSegue:(UIStoryboardSegue*) segue sender:(id) sender
 {
-    if([segue.identifier isEqualToString:@"showBlockedUsers"])
-    {
-        xmpp* xmppAccount = [[MLXMPPManager sharedInstance] getEnabledAccountForID:self.accountID];
-        // force blocklist update
-        [xmppAccount fetchBlocklist];
```
lissine0, line 899: I didn't copy this line over because I thought it was unnecessary:
`fetchBlocklist`gets the blocklist from the server and stores it in the DB as a local cache.
The method is called on login after discovery of blocking support. [MLIQProcessors.m L537](https://github.com/monal-im/Monal/blob/develop/Monal/Classes/MLIQProcessor.m#L537)
Subsequent blocks / unblocks would result in the relevant entries being added to (or removed from) the local cache. This is done when processing incoming setIQs  [MLIQProcessor.m L115-L156](https://github.com/monal-im/Monal/blob/develop/Monal/Classes/MLIQProcessor.m#L115-L156)

What do you think? Would it be more reliable to reload the blocklist from the server every time we open the BlockedUsers view?

tmolitor-stud-tu: No, our local cache should always be correct, no need to fetch this whole thing again when opening the ui :)

tmolitor-stud-tu:
> > I think the proper fix for nullability coming from the db could be to change `MLSQLite` to don't add something to its return dictionary if it is of type `NSNull`. Accessing a non-existent key of a dictionary automatically returns `nil`.
> 
> Note: I checked with the debugger, and the dictionary bridged by Swift didn't include any of the (key, value) pairs whose value is null

But where is the `NSNull` returned by `blockedJidsForAccount:` coming from, if it's not the the db (via `MLSQLite`)?

lissine0:
> > Note: I checked with the debugger, and the dictionary bridged by Swift didn't include any of the (key, value) pairs whose value is null
> 
> But where is the `NSNull` returned by `blockedJidsForAccount:` coming from, if it's not the the db (via `MLSQLite`)?

Yes, the DB contains null values, which become NSNull in the array of dictionaries returned by `blockedJidsForAccount:`
But from my understanding, the Swift version of the dictionary doesn't include any (key,value) pairs where the value is null, because as you noted, accessing a non-existent key will return nil anyway.

tmolitor-stud-tu:
That sounds a bit strange. So to clarify: An `NSDictionary` containing `NSNull` values (which is an ordinary object, distinct from a `nil` value, which is a null pointer) gets rewritten to not contain those keys pointing to `NSNull` values when casted/converted to a Swift dictionary??

Some background (you may already know all of this):
ObjC collection objects (`NSArray`, `NSDictionary` etc.) don't allow `nil` values. Only objc objects are allowed. Thus `NSNull` can be used to mark something as "this should be nil".
Querying a specific key in an `NSDictionary` that is not present will return nil, thus instead of writing `NSNull` to such a dictionary, one can just NOT write key-value pairs if the value is `nil`. The downside is obviously that in this case the key won't be part of the dictionary, too. That might be acceptable or not, depending on the use case (for example enumerating all keys isn't possible in this case).

Throughout our Monal codebase there are many places where values are optional. In objc world this means they can be `nil` or an object of a specific type. To always allow adding those safely to dictionaries etc. (for example as argument to a `MLHandler` that internally collects all arguments in two NSDictionaries), I introduced two macros: `nilWrapper()` and `nilExtractor()`. The wrapper will transparently return a `NSNull` instance, if the argument was nil and the unmodified argument otherwise. While the extractor will return the unmodified argument unless it is `NSNull`, in which case it will return `nil`.

matthewrfennell:
While I was looking at the null question above, I noticed a couple of other small things. Sorry it's taken some time to review.

I noticed that the text in the pop-up is present even after it gets dismissed

https://github.com/user-attachments/assets/1c568745-a417-488f-888f-4450a801ebd6

(side note: the colour of the default text and cancel button starts out wrong, I mention this as its the first time I've seen this happen in a SwiftUI view instead of an Obj-C one)

matthewrfennell:
It's possible to add the same blocked user multiple times. If I remove one, it removes all the matching rows at once

https://github.com/user-attachments/assets/ec6e4724-8d0f-487b-aa66-3e3c79107a72

matthewrfennell:
As a wishlist item: it would be cool if changes to the list are animated

tmolitor-stud-tu:
@matthewrfennell could you check if all your remarks were fixed now?

lissine0:
> @matthewrfennell could you check if all your remarks were fixed now?

The blocked addresses' duplication is still not fixed; the duplication is in the DB (due to SQLite considering two entries with the same accountID, node and host, but with NULL in the resource, as satisfying the UNIQUE constraint)
The most simple fix is to restructure the blockListCache like I suggested, so it has only account_id and blocked_jid

The other two remarks were fixed.

tmolitor-stud-tu:
The table should be `UNIQUE('account_id','node','host','resource')` and as far as I know that should hold even with `NULL` values.

lissine0:
> The table should be `UNIQUE('account_id','node','host','resource')` and as far as I know that should hold even with `NULL` values.

You can open sqlite3 in a terminal and execute the following script to test:
```
CREATE TABLE 'blocklistCache' (
                'account_id' INT NOT NULL,
                'node' TEXT,
                'host' TEXT NOT NULL,
                'resource' TEXT,
                UNIQUE('account_id','node','host','resource'),
                CHECK(LENGTH(host) > 0)
);

INSERT OR IGNORE INTO blocklistCache(account_id, node, host, resource) VALUES(1, 'juliet', 'example.org', NULL);
SELECT * FROM blocklistCache;
--Output: 1|juliet|example.org|

INSERT OR IGNORE INTO blocklistCache(account_id, node, host, resource) VALUES(1, 'juliet', 'example.org', NULL);
SELECT * FROM blocklistCache;
--Output: 1|juliet|example.org|
--        1|juliet|example.org|

INSERT OR IGNORE INTO blocklistCache(account_id, node, host, resource) VALUES(1, 'juliet', 'example.org', 'balcony');
SELECT * FROM blocklistCache;
--Output: 1|juliet|example.org|
--        1|juliet|example.org|
--        1|juliet|example.org|baclony

INSERT OR IGNORE INTO blocklistCache(account_id, node, host, resource) VALUES(1, 'juliet', 'example.org', 'balcony');
SELECT * FROM blocklistCache;
--Output: 1|juliet|example.org|
--        1|juliet|example.org|
--        1|juliet|example.org|baclony

```

tmolitor-stud-tu:
Well, I think that's a strange design decision, but documented over here: https://www.sqlite.org/nulls.html

According to this article, `SELECT DISTINCT [...]` should still work as expected.

According to the XEP, blocking can be done for a domain, a user or a device (e.g. a full jid also covering the resource).

You can modify the database table and design as you like as long as everything possible according to the XEP will remain possible.

matthewrfennell:
> @matthewrfennell could you check if all your remarks were fixed now?

Most are fixed or known about already, I see two left on my simulator:

* Text stays in the pop-up box after it's been dismissed (either by cancel or block)
* List removals are animated now, but not additions

lissine0:
> Most are fixed or known about already, I see two left on my simulator:
> 
>     * Text stays in the pop-up box after it's been dismissed (either by cancel or block)
> 
>     * List removals are animated now, but not additions

@matthewrfennell are you sure you updated to the version I (force-) pushed last week?
I have neither issue.


https://github.com/user-attachments/assets/cc507b16-348c-4b04-8007-9fb65b939003


https://github.com/user-attachments/assets/cbd5377e-c211-4629-90f4-957c8b7dbb95

lissine0:
The last version I pushed fixed the duplication.
But I discovered a bug, which also exists before my PR, so it's not related to my changes:
If you open Contact Details _from a chat view_, then block the contact, the button would change from `Block Contact` to `Unblock Contact`.
If you then go back to the contact's chat view, and reopen its contactDetails, the button will be reset to `Block Contact`
Note: there are no changes on the blocklist when that happens. Also, this problem doesn't happen if you access ContactDetails from the contactList, block contact, close it and open it again.
Nonetheless, I don't think the bug should block this PR, as it doesn't seem related.

matthewrfennell:
> @matthewrfennell are you sure you updated to the version I (force-) pushed last week? I have neither issue.

You're absolutely right, sorry about that, my mistake

tmolitor-stud-tu:
The bug seems to be related to the `DataLayer` method `-(NSDictionary* _Nullable) contactDictionaryForUsername:(NSString*) username forAccount:(NSNumber*) accountID` returning a `blocked` dictionary item that's not related to our `blocklistCache` database and that column should be removed from our `buddylist` table.
The method `-(uint8_t) isBlockedContact:(MLContact*) contact` should be the only source for the blocked state and is already used by `MLContact` method `-(void) handleBlockListRefresh:(NSNotification*) notification`.

I've created an issue for this over here: https://github.com/monal-im/Monal/issues/1254
Would be nice if you could fix that in a new pr, @lissine0

PR #1237, opened 2024-09-14
Improve button colors in ContactDetails

lissine0:
Change the `Add Contact` button to primary color.
Make the notification and encryption button text match their label's color.

Note: I kept the color of `Notify only when mentioned` as primary color.

Feel free to change the colors if you wish.

### DIFF
--- a/Monal/Classes/ContactDetails.swift
+++ b/Monal/Classes/ContactDetails.swift
@@ -275,25 +275,24 @@ struct ContactDetails: View {
                                 contact.isMuc ? Text("Notifications disabled") : Text("Contact is muted")
                             } icon: {
                                 Image(systemName: "bell.slash.fill")
-                                    .foregroundColor(.red)
                             }
+                            .foregroundStyle(Color.red)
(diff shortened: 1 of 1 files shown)


lissine0:
Or, we could opt to color the text only when its icon is red.

Monal/Classes/ContactDetails.swift
```diff
@@ -271,29 +271,34 @@ struct ContactDetails: View {
                         }
                     } label: {
                         if contact.isMuted  {
-                            Label {
-                                contact.isMuc ? Text("Notifications disabled") : Text("Contact is muted")
-                            } icon: {
-                                Image(systemName: "bell.slash.fill")
-                                    .foregroundColor(.red)
+                            Group {
```
tmolitor-stud-tu: why the `Group` wrapper? couldn't the `foregroundStyle` modifier be applied directly to the label?

tmolitor-stud-tu (changes requested):

PR #1238, opened 2024-09-15
Implement XEP-0157: Contact Addresses for XMPP Services

lissine0:
yax.im
conversations.im

### DIFF
--- a/Monal/Classes/MLIQProcessor.m
+++ b/Monal/Classes/MLIQProcessor.m
@@ -548,6 +548,20 @@ $$class_handler(handleServerDiscoInfo, $$ID(xmpp*, account), $$ID(XMPPIQ*, iqNod
     //query external services to learn stun/turn servers
     if([features containsObject:@"urn:xmpp:extdisco:2"])
         [account queryExternalServicesOn:iqNode.fromUser];
+
+    //get the server's contact addresses (XEP-0157)
+    XMPPDataForm* dataForm = [iqNode findFirst:@"{http://jabber.org/protocol/disco#info}query/\\{http://jabber.org/network/serverinfo}result\\"];
+    NSMutableDictionary<NSString*, NSArray*>* resultDictionary = [NSMutableDictionary dictionary];
+    for (NSString* fieldName in dataForm.allKeys)
+    {
+        if ([fieldName hasSuffix:@"-addresses"])
+        {
+            NSArray* addresses = [dataForm getField:fieldName][@"allValues"];
+            if (addresses != nil && addresses.count > 0)
+                resultDictionary[fieldName] = addresses;
+        }
+    }
+    account.connectionProperties.serverContactAddresses = [resultDictionary copy];
 $$
 
 $$class_handler(handleServiceDiscoInfo, $$ID(xmpp*, account), $$ID(XMPPIQ*, iqNode))
--- a/Monal/Classes/MLXMPPConnection.h
+++ b/Monal/Classes/MLXMPPConnection.h
@@ -33,6 +33,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong) MLXMLNode* serverFeatures;
 @property (nonatomic, strong) NSSet* accountDiscoFeatures;
 @property (nonatomic, strong) NSSet* serverDiscoFeatures;
+@property (nonatomic, strong) NSDictionary<NSString*, NSArray*>* serverContactAddresses;
 
 @property (nonatomic, strong) NSMutableArray* discoveredServices;
 @property (nonatomic, strong) NSMutableArray* discoveredStunTurnServers;
--- a/Monal/Classes/MLXMPPConnection.m
+++ b/Monal/Classes/MLXMPPConnection.m
@@ -26,6 +26,7 @@
     self.serverFeatures = [MLXMLNode new];
     self.accountDiscoFeatures = [NSSet new];
     self.serverDiscoFeatures = [NSSet new];
+    self.serverContactAddresses = [NSDictionary new];
     self.conferenceServers = [NSMutableDictionary new];
     self.discoveredServices = [NSMutableArray new];
     self.discoveredStunTurnServers = [NSMutableArray new];
--- a/Monal/Classes/ServerDetails.swift
+++ b/Monal/Classes/ServerDetails.swift
@@ -334,8 +334,7 @@ struct ServerDetails: View {
 
         var result: [EntryData] = []
         let saslMethods = connection.saslMethods as! [String: Bool]
-        for method in saslMethods.keys.sorted() {
-            let used = saslMethods[method]!
+        for (method, used) in saslMethods.sortedByKey() {
             let supported = (SCRAM.supportedMechanisms(includingChannelBinding: true) as! [String]).contains(method)
             var description: String
             switch method {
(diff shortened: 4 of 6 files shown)


Monal/Classes/MLIQProcessor.m
```diff
@@ -548,6 +548,25 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
...
+        {
+            NSMutableArray* addresses = [NSMutableArray array];
+            for (MLXMLNode* valueElement in field.children)
+            {
+                [addresses addObject:valueElement.data];
+            }
+            resultDictionary[fieldName] = addresses;
+        }
+    }
+    account.connectionProperties.serverContactAddresses = resultDictionary;
```
lissine0: Is this fine? or is [ copy]  /  [ mutableCopy] more appropriate?

Monal/Classes/MLIQProcessor.m
```diff
@@ -548,6 +548,25 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
     //query external services to learn stun/turn servers
     if([features containsObject:@"urn:xmpp:extdisco:2"])
         [account queryExternalServicesOn:iqNode.fromUser];
+
+    //get the server's contact addresses (XEP-0157)
```
lissine0, line 552: I considered moving this section to a separate handler, but I think there is no need to send two separate IQs that are identical. See [xmpp.m L3959-L3961](https://github.com/monal-im/Monal/blob/develop/Monal/Classes/xmpp.m#L3959-L3961)

Monal/Classes/MLIQProcessor.m
```diff
@@ -548,6 +548,25 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
...
+    XMPPDataForm* dataForm = [iqNode findFirst:@"{http://jabber.org/protocol/disco#info}query/\\{http://jabber.org/network/serverinfo}result\\"];
+    NSMutableDictionary* resultDictionary = [NSMutableDictionary dictionary];
+    NSString* fieldName;
```
lissine0: Side question: would it be more efficient if NSMutableString is used here?

Monal/Classes/MLIQProcessor.m
```diff
@@ -548,6 +548,25 @@ +(BOOL) processRosterWithAccount:(xmpp*) account andIqNode:(XMPPIQ*) iqNode
...
+    for (MLXMLNode* field in dataForm.children)
```
tmolitor-stud-tu: Read the data-forms part of my xml query article and check out the dictionary and array interfaces in XMPPDataForm.h as well.

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: having this non-mutable would be better...

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: use NSMutableString where?

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: no, sending two iqs would be just stupid, your solution to add it here is great :)

tmolitor-stud-tu (changes requested):

Monal/Classes/MLIQProcessor.m
lissine0: `NSMutableString* fieldName;` instead of `NSString* fieldName;`

Monal/Classes/MLIQProcessor.m
lissine0: So do you mean just
`account.connectionProperties.serverContactAddresses = [resultDictionary copy];`

Or do you also mean changing `MLXMPPConnection.serverContactAddresses`'s type from `NSMutableDictionary*` to `NSDictionary*`

tmolitor-stud-tu: both, copy and change to NSDictionary

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: but the field name isn't mutated at all, why use NSMutableString? You'd even have to mutably copy the NSString over into an NSMutableString as opposed to just using the NSString directly as is.

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: This still needs to be addressed...

tmolitor-stud-tu:
quick heads up: this is still missing the dataforms rewrite to be mergeable...

lissine0:
Just as an FYI:

- the `%fieldName` extraction command doesn't work: it doesn't have any effect.
For example, `[iqNode findFirst:@"{http://jabber.org/protocol/disco#info}query/\\{http://jabber.org/network/serverinfo}result%%admin-addresses\\"];` returns the whole dataForm, as if `%%admin-addresses` wasn't there.
On the other hand, XMPPDataForm's `getField:(NSString*) name` works; and the `@fieldName` extraction command works.

 And

- XMPPDataForm's `objectAtIndexedSubscript:(NSInteger) idx` doesn't work.
In fact trying `[dataForm objectAtIndexedSubscript:0];` causes an exception: `Exception  NSException *   "*** -[__NSArray0 objectAtIndex:]: index 0 beyond bounds for empty array"`, eventhough the server does have published contact addresses.
XMPPDataForm.h does have a [comment](https://github.com/monal-im/Monal/blob/8f0a624/Monal/Classes/XMPPDataForm.h#L38) above the method signature, saying the interface is incomplete, so I assume this is expected.

tmolitor-stud-tu:
Why did you double the `%`? You were qurying for the field `%admin-addresses` which indeed isn't there.

But why use the `%` operator at all? That will return a dictionary with additional information, but you'll only need the field value, so the `@` operator should be enough...

tmolitor-stud-tu:
`objectAtIndexedSubscript` will only work for data-forms having multiple item elements encapsulating the form fields, see [example 8 of XEP-0004](https://xmpp.org/extensions/xep-0004.html#example-8).

tmolitor-stud-tu:
No, the "incomplete interface" comment means that it does not implement everything an `NSArray` or `NSDictionary` is capable of.

lissine0:
> Why did you double the `%`? You were qurying for the field `%admin-addresses` which indeed isn't there.

%% is needed to escape the % inside the string, otherwise it'll be looking for a conversion format (%@, %d etc.)
In fact, including a single % character results in a compilation error.

> But why use the `%` operator at all?

I don't need it in this case. I was just trying different things.

tmolitor-stud-tu:
ah, this is a interference between the `NS_FORMAT_FUNCTION` annotation (triggering compiler errors for incorrectly used format specifiers) of the `find:` method and the query implementation not allowing a format specifier at that position but using it as an "ordinary" extraction operator...

I'll fix that and use another operator instead of `%`, but the `@` should still work just fine.

lissine0:
I tested eb680c6f1 and the `&` extraction command works correctly

PR #1246, opened 2024-09-21
Integrate "Add Account (advanced)" in WelcomeLogIn.swift

lissine0:

### DIFF
--- a/Monal/Classes/DataLayer.m
+++ b/Monal/Classes/DataLayer.m
@@ -247,9 +247,11 @@ static NSDateFormatter* dbFormatter;
         NSString* query = @"UPDATE account SET server=?, other_port=?, username=?, resource=?, domain=?, enabled=?, directTLS=?, rosterName=?, statusMessage=?, needs_password_migration=? WHERE account_id=?;";
         NSString* server = (NSString*)[dictionary objectForKey:kServer];
         NSString* port = (NSString*)[dictionary objectForKey:kPort];
+        if ([port isEqual:@""])
+            port = nil;
         NSArray* params = @[
-            server == nil ? @"" : server,
-            port == nil ? @"5222" : port,
+            nilDefault(server, @""),
+            nilDefault(port, @"5222"),
             ((NSString*)[dictionary objectForKey:kUsername]),
             ((NSString*)[dictionary objectForKey:kResource]),
             ((NSString*)[dictionary objectForKey:kDomain]),
@@ -272,9 +274,11 @@ static NSDateFormatter* dbFormatter;
         NSString* query = @"INSERT INTO account (server, other_port, resource, domain, enabled, directTLS, username, rosterName, statusMessage, plain_activated) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?);";
         NSString* server = (NSString*) [dictionary objectForKey:kServer];
         NSString* port = (NSString*)[dictionary objectForKey:kPort];
+        if ([port isEqual:@""])
+            port = nil;
         NSArray* params = @[
-            server == nil ? @"" : server,
-            port == nil ? @"5222" : port,
+            nilDefault(server, @""),
+            nilDefault(port, @"5222"),
             ((NSString *)[dictionary objectForKey:kResource]),
             ((NSString *)[dictionary objectForKey:kDomain]),
             [dictionary objectForKey:kEnabled] ,
--- a/Monal/Classes/MLXMPPManager.m
+++ b/Monal/Classes/MLXMPPManager.m
@@ -716,6 +716,26 @@ static const int pingFreqencyMinutes = 5;       //about the same Conversations u
 
 -(NSNumber*) login:(NSString*) jid password:(NSString*) password
 {
+    NSArray* elements = [jid componentsSeparatedByString:@"@"];
+    MLAssert([elements count] > 1, @"Got invalid jid", (@{@"jid": nilWrapper(jid), @"elements": elements}));
+    NSString* domain = ((NSString*)[elements objectAtIndex:1]).lowercaseString;
+
+    //we don't want to set kPlainActivated (not even according to our preload list) and default to plain_activated=false,
+    //because the error message will warn the user and direct them to the advanced account creation menu to activate PLAIN
+    //if they still want to connect to this server
+    //only exception: yax.im --> we don't want to suggest a server during account creation that has a scary warning
+    //when logging in using another device afterwards
+    //TODO: to be removed once yax.im and quicksy.im supports SASL2 and SSDP!!
+    //TODO: use preload list and allow PLAIN for all others once enough domains are on this list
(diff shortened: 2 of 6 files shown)


tmolitor-stud-tu:
I don't like this. the welcome login screen should be easy to use for non-techies and advanced options should be hidden in a separate screen for tech savy people to use.

lissine0:
> the welcome login screen should be easy to use for non-techies and advanced options should be hidden in a separate screen for tech savy people to use.

That's exactly how it is in this PR. The screenshots above are what's shown when `Add Account (advanced)` is selected from the settings. Normal `Add Account` is still the same as before.

What this PR does, is basing `Add Account (advanced)` on `Add Account` instead of `Account Edit`
It feels more natural this way.

tmolitor-stud-tu:
Ah okay, that's okay. But, for the advanced view, could you remove the logo and change the explanatory text at the top to something explaining that this is an advanced expert account creation menu?

tmolitor-stud-tu:
We should make the direct tls switch only accessible (e.g. enabled) if the hostname or port fields contain something...

lissine0:
> Ah okay, that's okay. But, for the advanced view, could you remove the logo and change the explanatory text at the top to something explaining that this is an advanced expert account creation menu?

Which one of these looks do you prefer? this:




Or:


<br></br>

> We should make the direct tls switch only accessible (e.g. enabled) if the hostname or port fields contain something...

Would you like to also switch the toggle to `false` if the user types something in hostname / port, then enables directTLS, then clears the hostname / port?

tmolitor-stud-tu:
I'd prefer the third one. @FriedrichAltheide @matthewrfennell what do you prefer?

> Would you like to also switch the toggle to false if the user types something in hostname / port, then enables directTLS, then clears the hostname / port?

I'd say yes.

matthewrfennell:
> I'd prefer the third one. @FriedrichAltheide @matthewrfennell what do you prefer?

I agree

tmolitor-stud-tu:
okay, third one it is :)

Monal/Classes/MLXMPPManager.m
```diff
@@ -715,6 +715,26 @@ -(void) sendChatState:(BOOL) isTyping toContact:(MLContact*) contact
...
+    //if they still want to connect to this server
+    //only exception: yax.im --> we don't want to suggest a server during account creation that has a scary warning
+    //when logging in using another device afterwards
+    //TODO: to be removed once yax.im and quicksy.im supports SASL2 and SSDP!!
+    //TODO: use preload list and allow PLAIN for all others once enough domains are on this list
+    //allow plain for all servers not on preload list, since prosody with SASL2 wasn't even released yet
+    BOOL defaultPlainActivated = YES;
+    BOOL plainActivated = ([domain isEqualToString:@"yax.im"] || [domain isEqualToString:@"quicksy.im"]) ? YES : defaultPlainActivated;
+
+return [self login:jid password:password hardcodedServer:nil hardcodedPort:nil forceDirectTLS:NO allowPlainAuth:plainActivated];
```
tmolitor-stud-tu: wrong indentation.

Monal/Classes/WelcomeLogIn.swift
```diff
@@ -154,7 +174,40 @@ struct WelcomeLogIn: View {
...
+
+                            if advancedMode {
+                                TextField("Optional Hardcoded Hostname", text: $hardcodedServer)
+                                    .textInputAutocapitalization(.never)
+                                    .autocorrectionDisabled()
+                                    .keyboardType(.URL)
+                                    .addClearButton(isEditing:  hardcodedServer.count > 0, text: $hardcodedServer)
+                                    .listRowSeparator(.hidden)
+
+                                TextField("Optional Port", text: $hardcodedPort)
```
tmolitor-stud-tu: should be "Optional Hardcoded Port" (the "hardcoded" is missing)

Monal/Classes/WelcomeLogIn.swift
```diff
@@ -154,7 +174,40 @@ struct WelcomeLogIn: View {
...
+                                    Text("Always use direct TLS, not STARTTLS")
+                                }
+                                .disabled(hardcodedServer.isEmpty && hardcodedPort.isEmpty)
+                                .onChange(of: [hardcodedServer, hardcodedPort]) { _ in
+                                    if hardcodedServer == "" && hardcodedPort == "" {
+                                        forceDirectTLS = false
+                                    }
+                                }
+
+                                Toggle(isOn: $allowPlainAuth){
```
tmolitor-stud-tu: could you add a list of domains for which this toggle isn't allowed to be turned on? that list should currently only contain "conversations.im", but more might be added.

tmolitor-stud-tu (changes requested):
Almost mergeable :)

Monal/Classes/WelcomeLogIn.swift
tmolitor-stud-tu: @lissine0  this seems still to be missing, right?

tmolitor-stud-tu (changes requested):

lissine0:
I made some other changes aside from those you asked for. Here's a demo of them:

https://github.com/user-attachments/assets/11b97e74-cb73-4d77-8ae4-3484cb669747

The reason why I did this is: the hardcoded port is only used in connection if a hardcoded server is set.
And I think the forceDirectTLS setting is also only used if there's a hardcoded server.
Thus I made the UI changes to make the users aware that those settings only take effect when setting a hardcoded server.

Also, in the old UI, 5222 is set by default as hardcoded port, even when doing a normal account adding. That can confuse users into thinking that StartTLS is used because of that value.
Thus I made 5222 the default only when a hardcoded server is set. Otherwise an empty string is used.
(We can store an empty string in a DB column of type Int because of a quirk in SQLite: the column's type is just a hint, and it doesn't prevent us from from storing a string in an int column)

And BTW, you were correct, the old add account (advanced) did indeed disallow plain auth for conversations.im
(XMPPEdit.m L342)

Monal/Classes/DataLayer.m
```diff
@@ -247,9 +247,11 @@ -(BOOL) updateAccounWithDictionary:(NSDictionary*) dictionary
         NSString* query = @"UPDATE account SET server=?, other_port=?, username=?, resource=?, domain=?, enabled=?, directTLS=?, rosterName=?, statusMessage=?, needs_password_migration=? WHERE account_id=?;";
         NSString* server = (NSString*)[dictionary objectForKey:kServer];
         NSString* port = (NSString*)[dictionary objectForKey:kPort];
+        if ([port isEqual:@""])
+            port = nil;
         NSArray* params = @[
-            server == nil ? @"" : server,
-            port == nil ? @"5222" : port,
+            nilDefault(server, @""),
```
tmolitor-stud-tu, line 253: I like that you are using the `nilDefault` macro :)

tmolitor-stud-tu (approved):

PR #1249, opened 2024-09-24
Fix crash when opening settings on iOS 18

matthewrfennell:
If you open the settings page on iOS 18.0, the app will crash. On iOS 17.4, it works fine. The crash happens because UIKit complains that the account cell is initialised twice.

```
CRASH(NSInternalInconsistencyException): View was already initialized: <MLSwitchCell: 0x10c03f000; baseClass = UITableViewCell ...>
```

We initialise the cell for the first time in
`MLSettingsTableViewController:tableView`:

```objective-c
MLSwitchCell* cell = [tableView dequeueReusableCellWithIdentifier:@"AccountCell" forIndexPath:indexPath];
```

Just below, we call `AccountListController:initContactCell`, which initialises the cell for the second time:

```objective-c
cell = [cell initWithStyle:UITableViewCellStyleSubtitle reuseIdentifier:@"AccountCell"];
```

We seem to do this so we can mostly resuse the styling defined in the nib file for AccountCell, but additionally add the subtitle styling that lets us show the time that the account was last connected. This approach is recommended here: https://stackoverflow.com/a/40809039

However, this seems incorrect since it leads to the cell being initialised twice. Probably Apple were not asserting this condition before, but are now.

So, only for the account cell, manually initialise the cell.

This changes the cells styling slightly. However, it still looks fine, and the settings screens will be rewritten in SwiftUI soon anyway, so I don't think it is worth spending lots of time on. But - if you disagree, I'm happy to do it :)

Before:

settings-contact-before

After:

settings-contact-after

### DIFF
--- a/Monal/Classes/AccountListController.m
+++ b/Monal/Classes/AccountListController.m
@@ -76,7 +76,6 @@
 -(void) initContactCell:(MLSwitchCell*) cell forAccNo:(NSUInteger) accNo
 {
     [cell initTapCell:@"\n\n"];
-    cell = [cell initWithStyle:UITableViewCellStyleSubtitle reuseIdentifier:@"AccountCell"];
     NSDictionary* account = [self.accountList objectAtIndex:accNo];
     MLAssert(account != nil, ([NSString stringWithFormat:@"Expected non nil account in row %lu", (unsigned long)accNo]));
     if([(NSString*)[account objectForKey:@"domain"] length] > 0) {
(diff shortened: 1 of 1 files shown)


tmolitor-stud-tu:
I did not check this on a real device, but according to @FriedrichAltheide, the crash only happens in the simulator.

But since I don't know if and when Apple might harden their API and also crash on a real device, we should fix that.

According to @FriedrichAltheide your usage of that reusable cell is not conforming to the API documentation and could lead to problems.

I think we should therefore remove the "connected since" stuff to fix that and only include this text again when doing the swiftui rewrite.

matthewrfennell:
Thanks both for the review! Here's how it looks now

settings-page

PR #1255, opened 2024-09-30
Properly initialize a contact's blocked state

### LINKED ISSUE #1254: [Bug]: Blocked state not properly synchronized in contact details
If you open Contact Details _from a chat view_, then block the contact, the button would change from `Block Contact` to `Unblock Contact`.
If you then go back to the contact's chat view, and reopen its contactDetails, the button will be reset to `Block Contact`
Note: there are no changes on the blocklist when that happens. Also, this problem doesn't happen if you access ContactDetails from the contactList, block contact, close it and open it again.

This bug seems to be related to the `DataLayer` method `-(NSDictionary* _Nullable) contactDictionaryForUsername:(NSString*) username forAccount:(NSNumber*) accountID` returning a `blocked` dictionary item that's not related to our `blocklistCache` database and that column should be removed from our `buddylist` table.
The method `-(uint8_t) isBlockedContact:(MLContact*) contact` should be the only source for the blocked state and is already used by `MLContact` method `-(void) handleBlockListRefresh:(NSNotification*) notification`.

lissine0:
Fixes #1254

The changes I _force-pushed_ are not strictly necessary. But I think they make it clearer that we don't expect a `@"blocked"` key in the dictionary we get from DataLayer
This works because an uninitialized BOOL is set to false by default.

### DIFF
--- a/Monal/Classes/MLContact.m
+++ b/Monal/Classes/MLContact.m
@@ -94,7 +94,6 @@ static NSMutableDictionary* _singletonCache;
             //@"muc_nick": nil,
             @"Muc": @NO,
             @"pinned": @NO,
-            @"blocked": @NO,
             @"encrypt": @NO,
             @"muted": @NO,
             @"status": @"",
@@ -119,7 +118,6 @@ static NSMutableDictionary* _singletonCache;
             @"muc_type": kMucTypeGroup,
             @"Muc": @YES,
             @"pinned": @NO,
-            @"blocked": @NO,
             @"encrypt": @NO,
             @"muted": @NO,
             @"status": @"",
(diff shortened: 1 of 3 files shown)


tmolitor-stud-tu:
This still has a bug: the dummy contact could still be blocked (whole domain or by bare jid), even if it is a dummy contact not being in our buddylist.

Your code does not handle that correctly.

Monal/Classes/MLContact.m
```diff
@@ -233,7 +228,12 @@ +(MLContact*) createContactFromDatabaseWithJid:(NSString*) jid andAccountID:(NSN
         }];
     }
     else
-        return [self contactFromDictionary:contactDict];
+    {
```
tmolitor-stud-tu, line 231: the curly braces are not needed here, because this is a single instruction.

tmolitor-stud-tu (changes requested):
Small changes required, other than that, this is now mergeable :)

PR #1261, opened 2024-10-13
Use UIKitWorkaround for own OmemoKeysView

lissine0:
This makes OmemoKeysView's top bar visible on iPad when accessed from AccountEdit

Before this commit:
before

After this commit:
after
(Some things were redacted from the screenshots for privacy)

This commit should be included in https://github.com/monal-im/Monal/pull/1258


tmolitor-stud-tu:
Thanks for fixing this :)

The commit has first do go into a new betarelease before it will be released, the beta and stable branches are locked down to only accept prs which let the code "bubble down".

PR #1283, opened 2024-11-02
[Draft]: Rewrite the Account Settings view in SwiftUI

### LINKED ISSUE #745: [Bug]: Settings / Account menu - Make clear to confirm changes before leaving
### What Monal Release channel are you using?

AppStore

### iOS system version

14.8

### iOS Monal version

5.0.1

### macOS system version

_No response_

### macOS Monal version

_No response_

### Used XMPP server (domain)

yax.im

### Which XMPP-Server software are you using?

Prosody

### XMPP Server Software Version

_No response_

### How many accounts are you using in Monal?

2

### What happened?

1. Go to Settings
2. Go to your account
3. Deactivate the account toggle
4. Leave the menu
5. Nothing changed

--> Monal should ask you if you want to confirm your changes - or just do it on change

### Anything else?

6. Monal should tell the 'offline' user why they cannot write messages anymore

### Related Issues

- [X] I have cross-checked this overview https://github.com/monal-im/Monal/issues/322 as well as filtered for related labels https://github.com/monal-im/Monal/labels

### XEP-Check

- [X] I have checked that at least XEP-198, XEP-0280, XEP-0352, XEP-0357 and XEP-0163 are activated on my server and shown as 'green' under Settings --> Account--> (i) in advanced settings

### Notifications-Menu

- [X] I have checked that all checkboxes are ticked under Settings --> Notifications

lissine0:
Most Notable changes:
- Get rid of the Save button, and dynamically update the AccountSettings view after enabling / disabling the account.
Fixes #745
- Move the Login credentials / settings to a separate view

Todo:
- [ ] Show the current account status (connected / connecting... / disabled).
This is necessary because we currently don't show anything to the user when they enable disable the account. Some ideas on how to display it:
    - An entry from the settings' accountList
    - A colored circle in the corner of the avatar. Though this alone is insufficient, because it would be inacessible for the visually impaired (e.g. colorblind users). It would need to be in addition to something else in text.
    - Show the status in the form
        - in a separate list entry
        - in the list entry containing the toggle
        - in the form section header
- [ ] Currently, changes to the Display Name and to the Status Message only get sent to the server on submit (i.e. on clicking return on the iOS keyboard / pressing enter on a physical keyboard)
Improve this situation, by: 
    - Sending changes to the server when the Textfield isn't in focus anymore, instead of on submit.
    Or
    - Showing a "save" icon. (inspired by Conversations, when you edit a group's name / topic). This can be placed like the "clear" button on Textfields 
    Or
    - when clicking the Display Name or Status Message fields, open an alert with a TextField, a cancel button and a save button.
 - [ ] Find a SwiftUI-alternative for MBProgressHUD: Something to indicate success when the account is removed or deleted or its chat history cleared. Some possibilities:
    - Use a success alert and dismiss it programmatically after a short time
    - Use a normal success alert that the user dismisses manually
    - Use a SwiftUI library that's similar to MBProgressHUD e.g. https://github.com/JonasGessner/JGProgressHUD-SwiftUI
- [ ] Improve the presentation of the avatar (shadows etc.) and move the whole thing to a separate file, AvatarPicker.swift, that would also be used by ContactDetails.swift
- [ ] Send notifications using MLNotificationQueue instead of NotificationCenter. Thilo also recommends that we post notifications from the model class.
- [ ] Thilo recommends that we use MLContact with ObservableKVOWrapper to dynamically get avatar updates, instead of observing for notifications
- [ ] Fix the following bugs:
    - [ ] If you put a wrong password in Login Credentials and save, the state of the enabled toggle will be out of sync with the account state.
- [ ] Test the PR thoroughly
- [ ] Ask for the account password before doing an in-band account deletion. (new feature)
- [ ] Add a Link to AccountSettings in ContactDetails for note to self chats
- [ ] Adapt the following commit to the new account settings view https://github.com/monal-im/Monal/pull/1465/commits/6985941d30ac99ac981179cfda895437500e4b7e (Display the splitView placeholder when clearing an account's history)
- [ ] Allow changing the avatar PEP node's access model. See https://xmpp.org/extensions/xep-0060.html#accessmodels and  https://github.com/monal-im/Monal/blob/b8849a6ab/Monal/Classes/xmpp.m#L5717. (The default access model used by Monal prevents your avatar from being visible in public channels)

Note: I may want to use `showPromisingLoadingOverlay` when changing the status message / display name.

Some screenshots from the initial version of this PR:


tmolitor-stud-tu:
I'm not sure if you already did this (I've not checked the code): the ui for the display and change of the avatar image should be the same as the code used for the avatar image in groupchats you own. --> that code should be refactored into a new view to not be duplicated between these two.

tmolitor-stud-tu:
I'd move the change password button to the login credentials view

tmolitor-stud-tu:
The status message should be multiline, just like editing the group/channel description for mucs (the code should be refactured to use the same view for both).

lissine0:
> the ui for the display and change of the avatar image should be the same as the code used for the avatar image in groupchats you own. --> that code should be refactored into a new view to not be duplicated between these two.

That is already mentioned in the todo list above:
> Improve the presentation of the avatar (shadows etc.) and move the whole thing to a separate file, AvatarPicker.swift, that would also be used by ContactDetails.swift

tmolitor-stud-tu:
> when clicking the Display Name or Status Message fields, open an alert with a TextField, a cancel button and a save button.

I like that extra alert most :)

We should write a generic view/view modifier etc. for this text input alert and use that everywhere such an alert is displayed


> Find a SwiftUI-alternative for MBProgressHUD: Something to indicate success when the account is removed or deleted or its chat history cleared. Some possibilities:

We already have that: `LoadingOverlay`, see the reconnect button in our debug menu on how to use a timer to display it...but for account removal I'd prefer using proper promises to truly indicate that the server processed the request...

tmolitor-stud-tu:
>  If you put a wrong password in Login Credentials and save, the state of the enabled toggle will be out of sync with the account state.

Should be fixed by using xmpp.m as model class, too (e.g. use `ObservableKVOWrapper` and the `accountState` property or a new property reflecting the enabled state as bound value of that toggle (it might be necessary to write a wrapping binding to translate states or trigger some `MLXMPPManager` method on enable/disable...or implement these actions/translations as writable computed property in xmpp.m)

tmolitor-stud-tu:
> Currently, changes to the Display Name and to the Status Message only get sent to the server on submit (i.e. on clicking return on the iOS keyboard / pressing enter on a physical keyboard)

I had the same problem, you can check out my solution in https://github.com/monal-im/Monal/blob/develop/Monal/Classes/ContactDetails.swift#L356

Basically `MLContact.m` starts a timer in `setNickNameView` and that timer is reset on every new char. on timeout, the change is submitted to the lower xmpp layer.

PR #1317, opened 2024-11-30
Rewrite the Settings view in SwiftUI

lissine0:
I was going to link the UIKit AccountEdit view from the SwiftUI settings, so alpha users can use this PR, but it turned out to be ugly:

https://github.com/user-attachments/assets/111afc30-bce1-487f-b145-11a40c42d2b7

As a result, the UIKit views rewritten in this PR are still being used. This should be the case until the account settings are in SwiftUI.
In order to test this PR, you need to revert https://github.com/lissine0/Monal/commit/60839b8c

Notes:

- The UIKit version of the settings is presented in a split view on iPads: the settings view on the left and the currently selected subview on the right. I didn't port that behavior. It currently looks on iPads like it does on macOS.
- The UIKit version of the settings has in-app modals for accessing the mail app and the app store. I didn't port that behavior, and relied instead on normal links (`mailto:` and `itms-apps://`) which should open the relevant apps _externally_ (I did not test this because it's not possible on the Simulator, but it should work).

Please indicate if you want these things added to the SwiftUI rewrite. I think the second one makes sense, as the modal is nicer than opening an app externally, but I'm not sure about the split view.

### DIFF
--- a/Monal/Classes/AccountList.swift
+++ b/Monal/Classes/AccountList.swift
@@ -1,0 +1,111 @@
+//
+//  AccountList.swift
+//  Monal
+//
+//  Created by lissine on 30/11/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+private class Account: Identifiable {
+    let accountID: NSNumber
+    let username: String
+    let domain: String
+    var enabled: Bool
+    var jid: String {
+        return username+"@"+domain
+    }
+    var connected: Bool {
+        return MLXMPPManager.sharedInstance().isAccount(forIdConnected: self.accountID)
+    }
+    var connectedTime: Date {
+        return MLXMPPManager.sharedInstance().connectedTime(for: self.accountID)
+    }
+    var avatar: UIImage {
+        return MLImageManager.sharedInstance().getIconFor(MLContact.createContact(fromJid: self.jid, andAccountID: self.accountID)) ?? UIImage(named: "noicon")!
+    }
+    // Conformance to the Identifiable protocol
+    var id: NSNumber {
+        return accountID
+    }
+
+    init(account: [String: Any]) {
+        self.accountID = account["account_id"] as! NSNumber
+        self.username = account["username"] as! String
+        self.domain = account["domain"] as! String
+        self.enabled = account["enabled"] as! Bool
+    }
+}
+
+private struct AccountEntry: View {
+    let account: Account
+    let uptimeFormatter: DateFormatter
+
+    init(account: Account) {
+        self.account = account
+        self.uptimeFormatter = DateFormatter()
+        uptimeFormatter.dateStyle = .short
+        uptimeFormatter.timeStyle = .short
+        uptimeFormatter.doesRelativeDateFormatting = true
+    }
+
+    var connectionStatusString: String {
+        if account.enabled && account.connected {
+            return String(format: NSLocalizedString("Connected since: %@", comment: ""), uptimeFormatter.string(from: account.connectedTime))
+        } else if account.enabled && !account.connected {
+            return NSLocalizedString("Connecting...", comment: "")
+        } else {
+            return NSLocalizedString("Account disabled", comment: "")
+        }
+    }
+
+    var body: some View {
+        HStack {
+            Image(uiImage: account.avatar)
+                .resizable()
+                .scaledToFit()
+                .frame(width: 40, height: 40)
+                .padding(.leading, -3)
+                .padding(.trailing, 4)
+            VStack {
+                Text(account.jid)
+                    .frame(maxWidth: .infinity, alignment: .leading)
+
+                Text(self.connectionStatusString)
+                    .font(.footnote)
+                    .frame(maxWidth: .infinity, alignment: .leading)
+            }
+            Spacer()
+            Image(systemName: account.enabled ? (account.connected ? "checkmark.circle.fill" : "checkmark.circle") : "circle")
+                .foregroundStyle(Color.accentColor)
+        }
+    }
+}
+
+struct AccountList: View {
+    @State private var accounts: [Account] = getAccountList()
+
+    private static func getAccountList() -> [Account] {
+        return (DataLayer.sharedInstance().accountList() as! [[String: Any]]).map { Account(account: $0) }
+    }
+    private func refreshAccountList() {
+        self.accounts = AccountList.getAccountList()
+    }
+
+    var body: some View {
+        List {
+            ForEach(accounts) { account in
+                NavigationLink {
+                    LazyClosureView(EmptyView())
+                } label: {
+                    AccountEntry(account: account)
+                }
+            }
+        }
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kMonalAccountStatusChanged)).receive(on: RunLoop.main)) { notification in
+            DispatchQueue.main.async {
+                DDLogVerbose("Refreshing the account list in the Settings view")
+                refreshAccountList()
+            }
+        }
+    }
+}
--- a/Monal/Classes/Settings.swift
+++ b/Monal/Classes/Settings.swift
@@ -1,0 +1,108 @@
+//
+//  Settings.swift
+//  Monal
+//
+//  Created by lissine on 30/11/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+struct Settings: View {
+    @State private var tappedVersionInfo = 0
+    @State private var showDebugEntry = HelperTools.defaultsDB().bool(forKey: "showLogInSettings")
+    var delegate: SheetDismisserProtocol
+    var body: some View {
+        Form {
+            Section(header: Text("")) {
+                AccountList()
+#if !IS_QUICKSY
+                NavigationLink(destination: LazyClosureView(WelcomeLogIn(delegate: delegate))) {
+                    Text("Add Account")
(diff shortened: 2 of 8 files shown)


tmolitor-stud-tu:
That's really ugly and quite strange. I'd be curious what causes this distortion.

As for your PR itself: I just tried it and I'm okay with the `mailto:` and `itms-apps://`, but the split-view is broken now: opening the settings on ipad or iphone big enough to be in split-view will open the settings but don't show any back button or other means to close the settings again.

Even though I liked to have a split-view for our settings, I'm okay with removing that and only having a single-view in that case. But closing the settings should always be possible, split-view or not.

lissine0:
> opening the settings on ipad or iphone big enough to be in split-view will open the settings but don't show any back button or other means to close the settings again.

On iPads (at least on the iPad 6th generation simulator), you can close the modal by clicking the area outside it, or by swiping it down.
On big-screen iPhones, this problem exists when in landscape mode;
It turns out that this was happening for all views using AddTopLevelNavigation.
I pushed a commit that fixes it.

tmolitor-stud-tu:
I don't like it if swiping down/tapping outside is the only method to close things. there should always be a button to do the same.

Monal/Classes/AccountList.swift
```diff
@@ -0,0 +1,111 @@
...
+        return username+"@"+domain
+    }
+    var connected: Bool {
+        return MLXMPPManager.sharedInstance().isAccount(forIdConnected: self.accountID)
+    }
+    var connectedTime: Date {
+        return MLXMPPManager.sharedInstance().connectedTime(for: self.accountID)
+    }
+    var avatar: UIImage {
+        return MLImageManager.sharedInstance().getIconFor(MLContact.createContact(fromJid: self.jid, andAccountID: self.accountID)) ?? UIImage(named: "noicon")!
```
tmolitor-stud-tu, line 24: You should use the `avatar` property of `MLContact` to get auto-updates (wrap the `MLContact` in `ObservableKVOWrapper`.

Monal/Classes/AccountList.swift
```diff
@@ -0,0 +1,111 @@
...
+                .resizable()
+                .scaledToFit()
+                .frame(width: 40, height: 40)
+                .padding(.leading, -3)
+                .padding(.trailing, 4)
+            VStack {
+                Text(account.jid)
+                    .frame(maxWidth: .infinity, alignment: .leading)
+
+                Text(self.connectionStatusString)
```
tmolitor-stud-tu, line 73: see above

Monal/Classes/AccountList.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    init(account: Account) {
+        self.account = account
+        self.uptimeFormatter = DateFormatter()
+        uptimeFormatter.dateStyle = .short
+        uptimeFormatter.timeStyle = .short
+        uptimeFormatter.doesRelativeDateFormatting = true
+    }
+
+    var connectionStatusString: String {
```
tmolitor-stud-tu, line 51: use the `accountState` of our `xmpp` object (in a `ObservableKVOWrapper`) and a switch statement to stringify the values of the `xmppState` enum.

Monal/Classes/AccountList.swift
```diff
@@ -0,0 +1,111 @@
...
+    let domain: String
+    var enabled: Bool
+    var jid: String {
+        return username+"@"+domain
+    }
+    var connected: Bool {
+        return MLXMPPManager.sharedInstance().isAccount(forIdConnected: self.accountID)
+    }
+    var connectedTime: Date {
+        return MLXMPPManager.sharedInstance().connectedTime(for: self.accountID)
```
tmolitor-stud-tu, line 21: use `connectedTime` property of our `xmpp` class, wrapped by a `ObservableKVOWrapper`.

tmolitor-stud-tu: the `connectedTimeFor:` method of `MLXMPPManager` can even be removed afterwards (no other code is using it if I don't miss something).

Monal/Classes/Settings.swift
```diff
@@ -0,0 +1,108 @@
...
+                NavigationLink(destination: LazyClosureView(WelcomeLogIn(delegate: delegate))) {
+                    Text("Add Account")
+                }
+                NavigationLink(destination: LazyClosureView(WelcomeLogIn(advancedMode: true, delegate: delegate))) {
+                    Text("Add Account (advanced)")
+                }
+#endif
+            }
+            Section(header: Text("App")) {
+                NavigationLink(destination: LazyClosureView(GeneralSettings())) {
```
tmolitor-stud-tu, line 27: The general settings should now be embedded in this settings rather than being a submenu (I only used a submenu because the settings weren't yet ported to swiftui).

Monal/Classes/Settings.swift
```diff
@@ -0,0 +1,108 @@
...
+                }
+#endif
+
+                // Version button
+                Button(action: {
+                    // Copy the version string to the clipboard
+                    UIPasteboard.general.setValue(HelperTools.appBuildVersionInfo(for: MLVersionType.IQ), forPasteboardType: UTType.utf8PlainText.identifier)
+#if !DEBUG
+                    tappedVersionInfo += 1
+                    if tappedVersionInfo > 16 {
```
tmolitor-stud-tu, line 89:
```suggestion
                    if tappedVersionInfo >= 16 {
```

Monal/Classes/Settings.swift
```diff
@@ -0,0 +1,108 @@
...
+//  Settings.swift
+//  Monal
+//
+//  Created by lissine on 30/11/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+struct Settings: View {
+    @State private var tappedVersionInfo = 0
+    @State private var showDebugEntry = HelperTools.defaultsDB().bool(forKey: "showLogInSettings")
```
tmolitor-stud-tu, line 11: Use a new class with `@defaultsDB` annotated properties to hold and autoupdate these settings, see the implementation of `GeneralSettingsDefaultsDB` in `GeneralSettings.swift`

Monal/Classes/Settings.swift
```diff
@@ -0,0 +1,108 @@
...
+#endif
+
+                // Version button
+                Button(action: {
+                    // Copy the version string to the clipboard
+                    UIPasteboard.general.setValue(HelperTools.appBuildVersionInfo(for: MLVersionType.IQ), forPasteboardType: UTType.utf8PlainText.identifier)
+#if !DEBUG
+                    tappedVersionInfo += 1
+                    if tappedVersionInfo > 16 {
+                        HelperTools.defaultsDB().set(true, forKey: "showLogInSettings")
```
tmolitor-stud-tu, line 90: You don't need this, if you use the `@defaultsDB` annotation (see comment above).

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -811,6 +810,15 @@ class SwiftuiInterface : NSObject {
         return host
     }
 
+    @objc
+    func makeSettingsView() -> UIViewController {
```
tmolitor-stud-tu, line 814: Since `makeSettingsView` does not need any special argument, you should add the settings view to our `makeView` function below.

Monal/Classes/AccountList.swift
```diff
@@ -0,0 +1,111 @@
...
+        List {
+            ForEach(accounts) { account in
+                NavigationLink {
+                    LazyClosureView(EmptyView())
+                } label: {
+                    AccountEntry(account: account)
+                }
+            }
+        }
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kMonalAccountStatusChanged)).receive(on: RunLoop.main)) { notification in
```
tmolitor-stud-tu, line 104: This won't be needed anymore, once the `xmpp` class is a proper data model class used in conjunction with the `ObservableKVOWrapper`.

tmolitor-stud-tu (changes requested):
My main point is: I'd like the `xmpp` class to be a data model just like the `MLContact` class is and then be used in conjunction with the `ObservableKVOWrapper` to get an auto-updating ui.

tmolitor-stud-tu:
username/domain are part of the connectionProperties property and can be used with the KVO wrapper: `xmppAccount.connectionProperties.identity.user` and `xmppAccount.connectionProperties.identity.domain`

enabled is not, that's right, but I'd like to make the xmpp class a observable data model just like MLContact is...so we should add an enabled property to the xmpp class and rewrite our code so that every other code in monal uses that property instead of asking the database directly and changing that property gets written to our database automatically...
--> half done in https://github.com/lissine0/Monal/commit/5828f36d9

to properly implement the `enable` property, the semantics/code of MLXMPPManager functions like `connectAccountWithDictionary` etc. must be changed...

lissine0:
> That's really ugly and quite strange. I'd be curious what causes this distortion.

I think it's using the [.plain ListStyle](https://developer.apple.com/documentation/swiftui/liststyle/plain)
When using that ListStyle, the header of the current section remains visible at the top of the screen as you scroll down.

tmolitor-stud-tu:
> I think it's using the [.plain ListStyle](https://developer.apple.com/documentation/swiftui/liststyle/plain) When using that ListStyle, the header of the current section remains visible at the top of the screen as you scroll down.

Well, that should be easily fixable, shouldn't it? :)

PR #1324, opened 2024-12-06
changes done

SidheshwarSarangal:


SidheshwarSarangal:
Changes done include changing the view chat file

tmolitor-stud-tu:
What's this all about? You did not include any description, most of your PR removes code and comments that are there for a reason and last but not least, you just propose some non-standard markup implementation not being compatible with the implementation of other clients: [XEP-0393](https://xmpp.org/extensions/xep-0393.html).

This it not mergeable.

PR #1325, opened 2024-12-06
Fix compile time errors on current develop

matthewrfennell:
Two changes:

* Adding a missing `%@` to the format string
* Making the array mutable as there is a call to `removeObjectAtIndex` lower down


tmolitor-stud-tu (approved):
thanks for fixing this, I didn't test my changes because they were small, but apparently most of them had bugs :/

PR #1337, opened 2024-12-18
Fix compile error in PasswordMigration.swift

matthewrfennell:


tmolitor-stud-tu:
Thanks, I already fixed it, but forgot to commit it. It's amended to 7d7fcc35a3d0c6d2a67e486dd41e5c4fed5b0ccc now.

PR #1338, opened 2024-12-18
Promise registry

matthewrfennell:
To test this, I:

- Added a plugin to my server that delays its response and ultimately returns an error or success.
- Triggered the update from the app, swiped it away to put it in the background, and waited for the app to suspend.
- Then, I triggered a notification from CloudKit to wake up the appex.
- Then, while the appex was awake, the server gave its response.
- Then, I swiped back to the app and saw the promise get consumed successfully:

https://github.com/user-attachments/assets/19a4f510-0c16-4247-ac8c-26dac996fbe7

Todo:

- [x] Test all scenarios
- [x] Consider creating a protocol (not sure if it's possible to pass a protocol to a handler - yes it is, a protocol can be used just like any other types)
- [x] Display only the error description instead of a debug description of the error in case of timeout / server error
- [x] Add comments to `MLPromise`
- [x] Split up commits / better commit messages

### DIFF
--- a/Monal/Classes/ContactDetails.swift
+++ b/Monal/Classes/ContactDetails.swift
@@ -144,11 +144,11 @@ struct ContactDetails: View {
                                             Text("Yes"),
                                             action: {
                                                 showPromisingLoadingOverlay(overlay, headlineView:Text("Removing avatar..."), descriptionView:Text("")) {
-                                                    promisifyMucAction(account:account, mucJid:contact.contactJid) {
-                                                        self.account.mucProcessor.publishAvatar(nil, forMuc: contact.contactJid)
-                                                    }
+                                                    self.account.mucProcessor.publishAvatar(nil, forMuc: contact.contactJid)
                                                 }.catch { error in
-                                                    errorAlert(title: Text("Error removing avatar!"), message: Text("\(String(describing:error))"))
+                                                    let nsError = error as NSError
+                                                    let description: String = nsError.userInfo[NSLocalizedDescriptionKey] as? String ?? "Could not remove avatar. Please try again."
+                                                    errorAlert(title: Text("Error removing avatar!"), message: Text(description))
                                                     hideLoadingOverlay(overlay)
                                                 }
                                             }
@@ -664,11 +664,11 @@ struct ContactDetails: View {
                 inputImage = nil
             }) { (image, cropRect, angle) in
                 showPromisingLoadingOverlay(overlay, headlineView:Text("Uploading avatar..."), descriptionView:Text("")) {
-                    promisifyMucAction(account:account, mucJid:contact.contactJid) {
-                        self.account.mucProcessor.publishAvatar(image, forMuc: contact.contactJid)
-                    }
+                    self.account.mucProcessor.publishAvatar(image, forMuc: contact.contactJid)
                 }.catch { error in
-                    errorAlert(title: Text("Error changing avatar!"), message: Text("\(String(describing:error))"))
+                    let nsError = error as NSError
+                    let description: String = nsError.userInfo[NSLocalizedDescriptionKey] as? String ?? "Could not remove avatar. Please try again."
+                    errorAlert(title: Text("Error changing avatar!"), message: Text(description))
                     hideLoadingOverlay(overlay)
                 }
             }
--- a/Monal/Classes/MLPromise.h
+++ b/Monal/Classes/MLPromise.h
@@ -1,0 +1,37 @@
+//
+//  MLPromise.h
+//  Monal
+//
+//  Created by Matthew Fennell on 29/09/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+#import "MLConstants.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ In Monal, we use the handler framework to create "serializable callbacks" so that processing can be handed off between the main app and app extension.
+ (See https://github.com/monal-im/Monal/wiki/Handler-Framework/ for more information).
+ Meanwhile, in SwiftUI, we use AnyPromises (aka PMKResolvers) to update the state of the UI as a result of an asynchronous action.
+
+ If an async action that should trigger a UI update gets started from the main app, but then the app is put into the background, the handler for that action will get called from the app extension. The UI should reflect the new state when the app is reopened.
+ Critically, the main app and app extension are separate processes and do not share memory, so we need a way to co-ordinate them.
+
+ This class handle this co-ordination via the database. Any function that creates a handler to respond to the response from the server, but returns an AnyPromise to the UI, could create an MLPromise, call toAnyPromise on it to return an AnyPromise to the UI, then pass the MLPromise instance to the handler. Then, the handler can fulfill or reject the promise, and the MLPromise will take care of updating the UI.
+
+ Meanwhile, the MLPromise takes care of:
+ * Co-ordinating whether the promise has been fulfilled, and its value, between the main app and app extension (via the database)
+ * Checking when the main app reopens whether the app extension had fulfilled a promise in the meantime, and then calling the AnyPromise for the UI
+ */
+@interface MLPromise : NSObject<NSSecureCoding>
+
+@property(nonatomic, strong) NSUUID* uuid;
+
+-(void) fulfill:(id _Nullable) arg;
+-(void) reject:(NSError*) error;
+-(AnyPromise*) toAnyPromise;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Monal/Classes/MLPromise.m
+++ b/Monal/Classes/MLPromise.m
@@ -1,0 +1,168 @@
+//
+//  MLPromise.m
+//  monalxmpp
+//
+//  Created by Matthew Fennell on 29/09/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+#import "DataLayer.h"
+#import "HelperTools.h"
+#import "MLPromise.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MLPromise()
+
+-(void) resolve:(id _Nullable) argument;
+
+@property(nonatomic, strong) AnyPromise* anyPromise;
+@property(nonatomic, strong) id resolvedArgument;
+@property(nonatomic, assign) BOOL isResolved;
+
+@end
+
+@implementation MLPromise
+
+static NSMutableDictionary* _resolvers;
+
++(void) initialize
+{
+    _resolvers = [NSMutableDictionary new];
+}
+
+-(instancetype) init
+{
+    self.uuid = [NSUUID UUID];
+    self.isResolved = false;
+
+    [self serialize];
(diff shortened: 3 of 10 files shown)


Monal/Classes/MLPromise.h
```diff
@@ -0,0 +1,26 @@
...
+#import "MLConstants.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MLPromise : NSObject <NSSecureCoding>
+
+-(NSUUID*) uuid;
+-(void) fulfill:(id _Nullable) arg;
+-(void) reject:(NSError*) error;
+-(AnyPromise*) toPromise;
```
tmolitor-stud-tu: Please rename to `toAnyPromise` or to `toPmkPromise` or `toPMKPromise`

Monal/Classes/MLPromise.h
```diff
@@ -0,0 +1,26 @@
...
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+#import "MLConstants.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MLPromise : NSObject <NSSecureCoding>
+
+-(NSUUID*) uuid;
```
tmolitor-stud-tu: that should be a property

Monal/Classes/MLPromise.h
```diff
@@ -0,0 +1,26 @@
...
+-(void) fulfill:(id _Nullable) arg;
+-(void) reject:(NSError*) error;
+-(AnyPromise*) toPromise;
+-(void) attemptConsume;
```
tmolitor-stud-tu: this is an internal api, don't declare it in the `MLPromise.h` header file but in an interface in the file that uses this internal api (see the MLVoipProcessor and MLCall interactions).

Monal/Classes/MLPromise.h
```diff
@@ -0,0 +1,26 @@
...
+
+-(void) serialize:(NSNotification*) notification;
```
tmolitor-stud-tu: This is internal api, too, right?

Monal/Classes/MLPromise.m
```diff
@@ -0,0 +1,166 @@
...
+        return self.promise;
+    }
+
+    self.promise = [AnyPromise promiseWithResolverBlock:^(PMKResolver resolve) {
+        [resolvers setObject:resolve forKey:self.uuid];
+        DDLogDebug(@"Adding resolver %@ with uuid %@ to resolvers map", resolve, self.uuid);
+        DDLogDebug(@"Resolvers map is now: %@", resolvers);
+    }];
+
+    return self.promise;
```
tmolitor-stud-tu: maybe rename it to `pmkPromise` or `anyPromise` to reflect the fact that this is a PromiseKit promise, not an MLPromise.

Monal/Classes/PasswordMigration.swift
```diff
@@ -28,6 +28,23 @@ struct PasswordMigration: View {
         DDLogInfo("Migration needed: \(String(describing:self.needingMigration))")
     }
     
+    func updatePassword(_ password: String) {
```
tmolitor-stud-tu: I already fixed that :)

Monal/Classes/MLPromise.m
```diff
@@ -0,0 +1,166 @@
...
+{
+    resolvers = [NSMutableDictionary new];
+}
+
+-(instancetype) init
+{
+    self.uuid = [NSUUID UUID];
+    self.isResolved = false;
+
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(serialize:) name:kMonalFrozen object:nil];
```
tmolitor-stud-tu: No, please don't do this: the `kMonalFrozen` is emitted just a few milliseconds before the app is actually suspended by iOS. Classes hooking this should only do last clean up things like releasing a file lock (that would crash the app with an 0xdeadl00c error otherwise). Likewise holding the db lock will trigger the same crash. Even logging is already deactivated when this notification is emitted.

Just serialize the `MLPromise` every time, it gets changed (use setters (and getters) or key-value-observers to trigger a serialize).

Monal/Classes/MLPromise.m
```diff
@@ -0,0 +1,166 @@
...
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(deserialize:) name:kMonalUnfrozen object:nil];
```
tmolitor-stud-tu: that one is fine though :)

Monal/Classes/MLPromise.m
```diff
@@ -0,0 +1,166 @@
...
+    {
+        DDLogError(@"Tried to consume promise %@ with uuid %@ when there is no resolver available", self, self.uuid);
+        return;
+    }
+
+    DDLogDebug(@"Resolving promise %@ with uuid %@ and argument %@", self, self.uuid, self.resolvedArgument);
+    resolve(self.resolvedArgument);
+
+    [resolvers removeObjectForKey:self.uuid];
+    DDLogDebug(@"Removed resolver with uuid %@ from resolvers map", self.uuid);
```
tmolitor-stud-tu: Maybe use `DDLogVerbose` instead of `DDLogDebug` (verbose is an even lower loglevel than the debug one)

tmolitor-stud-tu (changes requested):

Monal/Classes/MLPromise.m
```diff
@@ -0,0 +1,175 @@
...
+    DDLogVerbose(@"Deallocated promise %@ with uuid %@", self, self.uuid);
+}
+
+-(void) serialize
+{
+    [[DataLayer sharedInstance] addPromise:self];
+    DDLogVerbose(@"Serialized promise %@ with uuid %@", self, self.uuid);
+}
+
+-(void) deserialize:(NSNotification*) notification
```
tmolitor-stud-tu: this proxy is not needed. you can give a function without argument to `addObserver` as well (e.g. thje `-(void) deserialize` below).

Monal/Classes/MLPromise.m
```diff
@@ -0,0 +1,175 @@
...
+@property(nonatomic, strong) id resolvedArgument;
+@property(nonatomic, assign) BOOL isResolved;
+
+@end
+
+@implementation MLSerializablePromise
+
+@synthesize uuid;
+
+static NSMutableDictionary* resolvers;
```
tmolitor-stud-tu:
```suggestion
static NSMutableDictionary* _resolvers;
```

We are using underscores for global static vars (at least in all parts I managed to to touch/rewrite yet).

tmolitor-stud-tu (changes requested):
Looks really great! Only minor nitpicks.

Monal/Classes/ContactDetails.swift
```diff
@@ -144,11 +144,11 @@ struct ContactDetails: View {
...
                                             action: {
                                                 showPromisingLoadingOverlay(overlay, headlineView:Text("Removing avatar..."), descriptionView:Text("")) {
-                                                    promisifyMucAction(account:account, mucJid:contact.contactJid) {
-                                                        self.account.mucProcessor.publishAvatar(nil, forMuc: contact.contactJid)
-                                                    }
+                                                    self.account.mucProcessor.publishAvatar(nil, forMuc: contact.contactJid)
                                                 }.catch { error in
-                                                    errorAlert(title: Text("Error removing avatar!"), message: Text("\(String(describing:error))"))
+                                                    let nsError = error as NSError
+                                                    let description: String = nsError.userInfo[NSLocalizedDescriptionKey] as? String ?? "Could not remove avatar. Please try again."
```
tmolitor-stud-tu, line 150:
```suggestion
                                                    let description: String = nsError.userInfo[NSLocalizedDescriptionKey] as? String ?? NSLocalizedString("Could not remove avatar. Please try again.", comment:"")
```

Monal/Classes/ContactDetails.swift
```diff
@@ -664,11 +664,11 @@ struct ContactDetails: View {
...
             }) { (image, cropRect, angle) in
                 showPromisingLoadingOverlay(overlay, headlineView:Text("Uploading avatar..."), descriptionView:Text("")) {
-                    promisifyMucAction(account:account, mucJid:contact.contactJid) {
-                        self.account.mucProcessor.publishAvatar(image, forMuc: contact.contactJid)
-                    }
+                    self.account.mucProcessor.publishAvatar(image, forMuc: contact.contactJid)
                 }.catch { error in
-                    errorAlert(title: Text("Error changing avatar!"), message: Text("\(String(describing:error))"))
+                    let nsError = error as NSError
+                    let description: String = nsError.userInfo[NSLocalizedDescriptionKey] as? String ?? "Could not remove avatar. Please try again."
```
tmolitor-stud-tu, line 670:
```suggestion
                    let description: String = nsError.userInfo[NSLocalizedDescriptionKey] as? String ?? NSLocalizedString("Could not remove avatar. Please try again.", comment:"")
```

tmolitor-stud-tu, line 670: Shouldn't this error extraction be refactored into a `HelperTools` method taking an `NSError` and a fallback string, returning an error string?

We would still need to wrap the fallback string in `NSLocalizedString()`. 

Even better: write a ObjcC macro and a Swift macro doing this. That macro can include the call to `NSLocalizedString()`, thus abstracting it away, too :)

The ObjC macro would be this:
```objc
#include "metamacros.h"
#define errorString(error, ...) metamacro_if_eq(0, metamacro_argcount(__VA_ARGS__))( emptyDefault(((NSError*)error).userInfo[NSLocalizedDescriptionKey], @"", NSLocalizedString(@"Unknown error!", @"")) )( emptyDefault(((NSError*)error).userInfo[NSLocalizedDescriptionKey], @"", NSLocalizedString(metamacro_head(__VA_ARGS__), @"")) )
```
Swift macro resources, that should be helpful:
https://www.avanderlee.com/swift/macros/

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/macros/

https://github.com/krzysztofzablocki/Swift-Macros?tab=readme-ov-file

This can be a follow up PR, but don't forget that (we'll need that NSError to string conversion **a lot** when we use promises)

tmolitor-stud-tu: This can be a follow up PR, but don't forget that (we'll need that NSError to string conversion **a lot** when we use promises)

Monal/Classes/MLPromise.h
```diff
@@ -0,0 +1,37 @@
+//
+//  MLPromise.h
+//  Monal
+//
+//  Created by Matthew Fennell on 29/09/2024.
+//  Copyright © 2024 monal-im.org. All rights reserved.
+//
+
+#import "MLConstants.h"
```
tmolitor-stud-tu, line 9: Don't import MLConstants in a header file if you don't need it there (you can always import it in the .m file). Importing it in the header can cause strange compiler errors once using this header file somewhere.

General rule of thumb: don't import header files in other header files if this can be avoided (by using an `@class ...` forward declaration for example).

Monal/Classes/MLPromise.h
```diff
@@ -0,0 +1,37 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
```
tmolitor-stud-tu, line 13: I like that you added some detailed documentation :)

tmolitor-stud-tu (approved):
Small changes, I'll fix them myself :)

PR #1348, opened 2024-12-28
Refactor stale promise consumption

matthewrfennell:
This WIP PR contains refactors to allow an error popup on consumption of stale, rejected promises.

See https://github.com/matthewrfennell/Monal/pull/1/commits/3d8c5d2fe4a7108305b684318948ecb5e7b81fb8 for how I was thinking of implementing that error posting (based off this branch).

Feedback welcome, I've left as WIP as there are still some things I want to check (left comments for myself to review later)

### DIFF
--- a/Monal/Classes/MLMucProcessor.m
+++ b/Monal/Classes/MLMucProcessor.m
@@ -1311,7 +1311,7 @@ $$
 $$instance_handler(handleAvatarPublishResultInvalidation, account.mucProcessor, $$ID(xmpp*, account), $$ID(NSString*, room), $$ID(MLPromise*, promise))
     NSString* errorString = [NSString stringWithFormat:NSLocalizedString(@"Publishing avatar for muc '%@' returned timeout", @""), room];
     NSError* error = [NSError errorWithDomain:@"Monal" code:0 userInfo:@{NSLocalizedDescriptionKey: errorString}];
-    [promise reject:error];
+    [promise rejectWithError:error andNode:nil forAccountWithID:account.accountID];
 $$
 
 $$instance_handler(handleAvatarPublishResult, account.mucProcessor, $$ID(xmpp*, account), $$ID(XMPPIQ*, iqNode), $$ID(MLPromise*, promise))
@@ -1320,7 +1320,7 @@ $$instance_handler(handleAvatarPublishResult, account.mucProcessor, $$ID(xmpp*, 
         DDLogError(@"Publishing avatar for muc '%@' returned error: %@", iqNode.fromUser, [iqNode findFirst:@"error"]);
         NSString* errorString = [NSString stringWithFormat:NSLocalizedString(@"Failed to publish avatar image for group/channel %@", @""), iqNode.fromUser];
         NSError* error = [NSError errorWithDomain:@"Monal" code:0 userInfo:@{NSLocalizedDescriptionKey: errorString}];
-        [promise reject:error];
+        [promise rejectWithError:error andNode:iqNode forAccountWithID:account.accountID];
         return;
     }
     DDLogInfo(@"Successfully published avatar for muc: %@", iqNode.fromUser);
--- a/Monal/Classes/MLPromise.h
+++ b/Monal/Classes/MLPromise.h
@@ -27,10 +27,14 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, strong) NSUUID* uuid;
 
 -(void) fulfill:(id _Nullable) arg;
--(void) reject:(NSError*) error;
+-(void) rejectWithError:(NSError*) error andNode:(XMPPStanza* _Nullable) node forAccountWithID:(NSNumber*) accountID;
 -(AnyPromise*) toAnyPromise;
 
 +(void) removeStalePromises;
+
+@end
+
+@interface MLPromiseRejection : NSObject<NSSecureCoding>
 
 @end
 
--- a/Monal/Classes/MLPromise.h
+++ b/Monal/Classes/MLPromise.h
@@ -30,7 +30,7 @@ NS_ASSUME_NONNULL_BEGIN
 -(void) rejectWithError:(NSError*) error andNode:(XMPPStanza* _Nullable) node forAccountWithID:(NSNumber*) accountID;
 -(AnyPromise*) toAnyPromise;
 
-+(void) removeStalePromises;
++(void) consumeStalePromises;
 
 @end
 
--- a/Monal/Classes/MLPromise.m
+++ b/Monal/Classes/MLPromise.m
@@ -20,10 +20,54 @@ typedef NS_ENUM(NSInteger, PromiseResolutionState) {
     PromiseResolutionStateRejected,
 };
 
+@interface MLPromiseRejection()
+
+@property(nonatomic, strong) NSString* errorDescription;
+@property(nonatomic, strong) XMPPStanza* node;
+@property(nonatomic, strong) NSNumber* accountID;
+
+-(instancetype) initWithErrorDescription:(NSString*) description andNode:(XMPPStanza*) node forAccountWithID:(NSNumber*) accountID;
+
+@end
+
+@implementation MLPromiseRejection
+
+-(instancetype) initWithErrorDescription:(NSString*) errorDescription andNode:(XMPPStanza*) node forAccountWithID:(NSNumber*) accountID
+{
+    self = [self init];
+    self.errorDescription = errorDescription;
+    self.node = node;
+    self.accountID = accountID;
+    return self;
+}
+
+-(nullable instancetype) initWithCoder:(NSCoder*) coder
+{
+    self.errorDescription = [coder decodeObjectForKey:@"errorDescription"];
+    self.node = [coder decodeObjectForKey:@"node"];
+    self.accountID = [coder decodeObjectForKey:@"accountID"];
+    return self;
+}
+
+-(void) encodeWithCoder:(NSCoder*) coder
+{
+    [coder encodeObject:self.errorDescription forKey:@"errorDescription"];
+    [coder encodeObject:self.node forKey:@"node"];
+    [coder encodeObject:self.accountID forKey:@"accountID"];
+}
+
++(BOOL) supportsSecureCoding
+{
+    return YES;
+}
+
+@end
+
 @interface MLPromise()
 
 @property(nonatomic, strong) AnyPromise* anyPromise;
 @property(nonatomic, strong) id resolvedArgument;
+@property(nonatomic, strong) MLPromiseRejection* rejection;
 @property(nonatomic) PromiseResolutionState state;
 
 @end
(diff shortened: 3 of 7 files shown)


Monal/Classes/MLPromise.m
```diff
@@ -14,13 +14,62 @@
...
+}
+
+@end
+
+@interface MLPromise()
 
 @property(nonatomic, strong) AnyPromise* anyPromise;
 @property(nonatomic, strong) id resolvedArgument;
-@property(nonatomic, assign) BOOL isResolved;
+@property(nonatomic, strong) MLPromiseRejection* rejection;
```
matthewrfennell, line 70: It doesn't feel very clean to have a dedicated `MLPromiseRejection` that will be `nil` in the vast majority of cases - consider creating a union to cleanly separate the fulfilled and rejected cases.

Monal/Classes/MLPromise.m
```diff
@@ -136,26 +192,29 @@ -(void) attemptConsume
         return;
     }
 
-    if(!self.isResolved)
+    if(self.state == PromiseResolutionStateUnresolved)
```
matthewrfennell, line 195: We previously deleted all stale promises from the DB, but now they will only be deleted if they are resolved. I need to think if this is the right behaviour or not.

Monal/Classes/MLPromise.h
```diff
@@ -27,10 +27,14 @@ NS_ASSUME_NONNULL_BEGIN
...
--(void) reject:(NSError*) error;
+-(void) rejectWithError:(NSError*) error andNode:(XMPPStanza* _Nullable) node forAccountWithID:(NSNumber*) accountID;
 -(AnyPromise*) toAnyPromise;
 
-+(void) removeStalePromises;
++(void) consumeStalePromises;
+
+@end
+
+@interface MLPromiseRejection : NSObject<NSSecureCoding>
```
matthewrfennell, line 37: I don't like exposing this in the header, but it's necessary to add the entry in `HelperTools` `unserializeData:`

tmolitor-stud-tu: Couldn't you just forward declare this with `@class MLPromiseRejection;` in `HelperTools.m`?

Monal/Classes/MLPromise.m
```diff
@@ -14,13 +14,62 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
-@interface MLPromise()
+typedef NS_ENUM(NSInteger, PromiseResolutionState) {
+    PromiseResolutionStateUnresolved,
+    PromiseResolutionStateFulfilled,
+    PromiseResolutionStateRejected,
+};
+
+@interface MLPromiseRejection()
```
tmolitor-stud-tu, line 23: I'd add a readonly computed property returning the xmpp object for the stored accountID. Similar to the one in MLCall.m.

Monal/Classes/MLPromise.m
tmolitor-stud-tu: Yes, a union or tagged enum would be a good fit here :)

Monal/Classes/MLPromise.m
```diff
@@ -80,31 +132,35 @@ -(void) deserialize
...
 }
 
 -(void) fulfill:(id _Nullable) argument
 {
-    [self resolve:argument];
+    [self resolveWithArgument:argument andState:PromiseResolutionStateFulfilled];
 }
 
--(void) reject:(NSError*) error
+-(void) rejectWithError:(NSError*) error andNode:(XMPPStanza* _Nullable) node forAccountWithID:(NSNumber*) accountID
```
tmolitor-stud-tu, line 160: I think we should provide both a simple `-(void) reject:(NSError*) error` method and a more complicated `-(void) rejectWithError:(NSError*) error andNode:(XMPPStanza* _Nullable) node forAccountWithID:(NSNumber*) accountID`

The simple method could just set the accountID to `nil` or `@0` and the HelperTools method would have to be modifed to use something like "Error" or "General Error" as title instead of the jid, if the `accountID` wan't provided.

Monal/Classes/MLPromise.m
tmolitor-stud-tu: Problem is: this could still lead to stale promises: promises never resolved or rejected (for example all iq handler that don't have a matching invalidation handler that rejects the promise).

There are multiple solutions to this:
1. Always implicitly pass a promise to the handler (maybe named `_promise`) and create a generic invalidation handler to be specified if no special invalidation is to be used. That generic invalidation would reject the promise given in the `_promise` paramter. The promise passed in could be auto-generated with some macro magic if no var named `_promise` is in scope when the `$newHandler` macro is used.
2. Just silently remove unresolved promises from the db on app start without displaying an error. This way you'd have to write an explicit invalidation handler if you want to display an error on invalidation. No macro magic needed.

Case 2 has a downside, though: IQs that don't get resolved (or time out after 30 seconds, which automatically generates a fake IQ error response) before the XEP-0198 smacks session times out and a completely new xmpp session is created, won't generate any error. Even if the main app isn't killed (that'll leave the ui in a "spinning" state forever).
That downside even is a unresolved problem in our current codebase: The promise would neither be resolved nor rejected in that case. We'd have 

I think solution 2 is better nonetheless.
We could fix this by:
1. generate fake error IQs in the smacks error case, too.
2. write invalidation handlers accompanying all iq handlers in our codebase (that are *a lot*)
3. automatically call a generic invalidation handler if no invalidation handler was specified and that generic invalidatiuon handler could iterate over all handler arguments, pick out all MLPromises and reject them with some generic error message. That's possible, because calling an invalidation handler makes sure, the normal handler will never called in the future and thus all MLPromises are free to be resolved without any conflict (if not already resolved of course).

I prefer solution 3.

Monal/Classes/MLPromise.m
```diff
@@ -136,26 +192,29 @@ -(void) attemptConsume
...
 
-    [_resolvers removeObjectForKey:self.uuid];
-    DDLogVerbose(@"Removed resolver with uuid %@ from resolvers map", self.uuid);
-    DDLogVerbose(@"Resolvers map is now: %@", _resolvers);
+        [_resolvers removeObjectForKey:self.uuid];
+        DDLogVerbose(@"Removed resolver with uuid %@ from resolvers map", self.uuid);
+        DDLogVerbose(@"Resolvers map is now: %@", _resolvers);
+    }
 
     [[DataLayer sharedInstance] removePromise:self];
```
tmolitor-stud-tu, line 219: I talked about removing all promises from the db (even stale ones) in some comment above and now realize, that already happens :)

Monal/Classes/MonalAppDelegate.m
```diff
@@ -395,7 +395,7 @@ -(BOOL) application:(UIApplication*) application willFinishLaunchingWithOptions:
     });
     
     // Remove stale promises left in the DB that weren't consumed last time we ran the app
-    [MLPromise removeStalePromises];
+    [MLPromise consumeStalePromises];
```
tmolitor-stud-tu, line 398: I think we'll have to extract a list/set of all pending stale promises *after* making sure the appex is not running anymore (the last act in the will launch method), but consume these stale promises only after creating all xmpp objects later in the app startup cycle.

We must extract all stale promises that early in the app startup cycle to not introduce race conditions with the startup process that could introduce new MLPromises that aren't stale before or while all xmpp objects are created.

tmolitor-stud-tu: I'd not wait for lissine and implement it in such a way, that all accountIDs, for which no enabled account (and thus an xmpp object) could be found, simply show a generic error title instead of the jid (see some comment above). That can simply improved to always show the jid later on after lissine's rewrite.

But even then not giving an xmpp object and using a generic notification title for the error notification in this case would be a useful feature I'd like to have independent of our MLPromise framework.

tmolitor-stud-tu (changes requested):
Added lots of comments :)

Monal/Classes/MLPromise.h
matthewrfennell: It isn't enough in this case unfortunately - I get this compile error in `HelperTools.m`

> Receiver 'MLPromiseRejection' for class message is a forward declaration

tmolitor-stud-tu: Okay, then you could just add
```
@interface MLPromiseRejection : NSObject<NSSecureCoding>
@end
```
to the header and 
```
@interface MLPromiseRejection() : NSObject<NSSecureCoding>
[...]
@end
```
containing all your actual property/method declarations to your `MLPromise.m`. That way you'll only expose the fact that this class exists, but none of its (internal) methods :)

Monal/Classes/MLPromise.m
tmolitor-stud-tu: maybe I prefer solution 1 (together with 3?).

Monal/Classes/MLPromise.m
matthewrfennell: Are there situations where we want to reject the promise without passing an account?

My understanding of handlers is that each one should always be logically tied to an account, and from what I can tell, handlers are always called with an account (even if some handlers do not currently have a parameter for it).

tmolitor-stud-tu: Yes there are. The promise framework isn't meant only for IQ handlers. All sorts of handlers could be tied to an `MLPromise`. And not even every handler has an associated account (though almost all do have that).

If it would simplify things, we could require an `MLPromise` to always be tied to an account. Handlers without accounts aren't that frequent and restricting them to not be able to return an `MLPromise` backed promise might be okay.

Monal/Classes/MLMucProcessor.m
```diff
@@ -1379,24 +1303,24 @@ -(AnyPromise*) publishAvatar:(UIImage* _Nullable) image forMuc:(NSString*) room
...
     return [promise toAnyPromise];
 }
 
-$$instance_handler(handleAvatarPublishResultInvalidation, account.mucProcessor, $$ID(xmpp*, account), $$ID(NSString*, room), $$PROMISE(promise))
-    NSString* errorString = [NSString stringWithFormat:NSLocalizedString(@"Failed to publish avatar image for group/channel %@: please try again", @""), room];
+$$instance_handler(handleAvatarPublishResultInvalidation, account.mucProcessor, $$ID(xmpp*, account), $$ID(NSString*, room), $$ID(MLPromise*, promise))
+    NSString* errorString = [NSString stringWithFormat:NSLocalizedString(@"Publishing avatar for muc '%@' returned timeout", @""), room];
     NSError* error = [NSError errorWithDomain:@"Monal" code:0 userInfo:@{NSLocalizedDescriptionKey: errorString}];
-    [promise reject:error];
+    [promise rejectWithError:error andNode:nil forAccountWithID:account.accountID];
```
matthewrfennell: Swap the order around (as node will be the only optional arg) and then introduce a helper method without the `andNode` arg.

PR #1354, opened 2025-01-03
Remove unused modulemap from monalxmpp

matthewrfennell:
# Change

This modulemap is not actually used in the built target, and is causing conflicts that hamper ObjC <-> Swift interop. We do not need a custom module map; the auto-generated one is sufficient for us. Therefore:

* Remove the modulemap. It is already not used in the built target, and is just causing conflicts at compile time.
* Remove the `SWIFT_INCLUDE_PATHS` override, since there is no longer a modulemap there.

# What is the modulemap for?

Modules (via `@import`/`import`) don't import things textually (like `#include`/`#import` did), but make the compiler see the imports as logical units.  If multiple files import a module, it is only built once and they all get the same built module.

The consequence is that the compiler must know the mapping of module name to header files. This is provided by the modulemap.

# Build settings

`MODULEMAP_FILE`: by default, a modulemap will be automatically generated for your target during the build (using the target's umbrella header). Setting this will override that default modulemap

`DEFINES_MODULE`: tells Xcode your framework uses modules. This is necessary to allow the target to be imported as a module

`SWIT_INCLUDE_PATHS`: get searched by the compiler, looking for additional modules

# Why did it break?

We had an interesting combination of build flags:

* `DEFINES_MODULE` was already true, so we were using modules
* `MODULEMAP_FILE` was not set, so we actually still used the deafult modulemap
* `SWIFT_INCLUDE_PATHS` was set to monalxmpp, so the compiler saw the custom modulemap during compilation (even though it wasn't used)

This explains why `import monalxmpp` actually worked in the past even though it wasn't included in the modulemap - we were never actually using our custom map!

If you set `MODULEMAP_FILE` to our custom one, the build fails saying monalxmpp was not defined (which makes sense, as `monalxmpp` is not defined anywhere in our map).

*But* - having `SWIFT_INCLUDE_PATHS` point to `monalxmpp/` means that the custom module map is still picked up in the frontend of the compilation - even though it is not actually used in the built artifact.

These conflicts ultimately resulted in the ObjC not having the full definition; I was able to replicate this behaviour in a minimal fresh project, by combining the above 3 build flags.


tmolitor-stud-tu:
Thanks for this really thorough analysis!!!

tmolitor-stud-tu:
cherry-picked

PR #1362, opened 2025-01-04
Improve ObjC/Swift interop in monalxmpp target

matthewrfennell:
This PR fixes a few compile time errors seen when adding new Swift classes to `monalxmpp`:

- Ensure that all headers declared within the `monalxmpp.h` umbrella header are set to public visibility in Xcode, and part of the `monalxmpp` target in Xcode. This is so that they can be imported from Swift
- Make sure that all headers transitively imported by public headers within `monalxmpp` are also set to public (since they are also part of the public interface)
- Use framework imports (`<>` instead of `""`), allowing the compiler to check whether the framework is fully self-contained (to satisfy the `CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER` check)
- Declare a dependency on `monalxmpp` for the share sheet and notification service, as they also use the framework


tmolitor-stud-tu:
A question about your wording: I thought framework style imports are `@import xxx`, not `#import <xxx.h>` (which I used to call "absolute imports" as opposed to `#import "xxx.h"` which I called "relative imports").

matthewrfennell:
Agreed, it looks like my wording was not completely precise! By the above I mean "absolute imports".

PR #1372, opened 2025-01-16
Add Github workflow to run tests

matthewrfennell:
Here is an example run:

https://github.com/matthewrfennell/Monal/actions/runs/12838210714/job/35803392375

Next steps after this (in future PRs?)

- [x] Add github workflow to run tests
- [x] Restrict permissions of all workflows
- [x] Run `xcbeautify` on all build workflows
- [ ] Integrate tests in MonalAppDelegate.m into test framework
- [ ] (eventually) fix UI tests


tmolitor-stud-tu:
It would be great if the name, result and duration of our test could be extracted from this huge blob of xcodebuild output and printed in a separate workflow step. Maybe even with unicode checkmarks or crossmarks in front.

If you want some examples how to forward data between build steps, you could check out our stable build script and the changelog extraction in there.

No merge blocker, though :)

tmolitor-stud-tu:
merged via cherry-pick

PR #1374, opened 2025-01-20
Minimise workflow permissions

matthewrfennell:
I established the permissions needed using a combination of these two docs:

* https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token
* https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents

I've marked this as WIP for now as I haven't tested these yet.

To test: I was thinking of simplifying the runners so they only contain the bits that interact with GitHub, and then running them on a standard `macos-latest` runner on my fork.

That should provide at least some assurance that we won't suddenly get permission denied errors :)

### DIFF
--- a/.github/workflows/create-stable-pr.yml
+++ b/.github/workflows/create-stable-pr.yml
@@ -5,9 +5,13 @@ on:
     branches: [ beta ]
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
   create-pull-request:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
     steps:
       - name: Checkout Beta Branch
         uses: actions/checkout@v4
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -5,6 +5,8 @@ name: update-translations
 on:
   workflow_dispatch:
 
+permissions: {}
+
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This workflow contains a single job called "build"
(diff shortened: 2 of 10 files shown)


matthewrfennell:
Ok, testing went well! To test the permissions were not overly strict, I modified the jobs to let them run on my fork (removing the parts that would try to access fastlane, secrets etc and therefore fail).

But, I left the interactions with GitHub intact, and everything worked with the restricted permissions.

Here are the successful job runs:

| Action                   | Logs                                                                              |
| ------------------------ | --------------------------------------------------------------------------------- |
| xcode-swift-dependencies | https://github.com/matthewrfennell/Monal/actions/runs/12877097594/job/35901052554 |
| publish-stable-release   | https://github.com/matthewrfennell/Monal/actions/runs/12877972669/job/35903235713
| develop-push             | https://github.com/matthewrfennell/Monal/actions/runs/12877468430/job/35901954862 |
| create-stable-pr         | https://github.com/matthewrfennell/Monal/actions/runs/12877517667/job/35902085985 |
| stable.build-push        | https://github.com/matthewrfennell/Monal/actions/runs/12877772006/job/35902736658 |

.github/workflows/create-stable-pr.yml
```diff
@@ -5,9 +5,14 @@ on:
     branches: [ beta ]
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
   create-pull-request:
     runs-on: ubuntu-latest
+    permissions:
+      issues: write
```
tmolitor-stud-tu: why that? doesn't `pull-requests: write` suffice?

.github/workflows/update-translations.yml
```diff
@@ -5,6 +5,8 @@ name: update-translations
 on:
   workflow_dispatch:
 
+permissions: {}
+
```
tmolitor-stud-tu, line 9: this needs push permissions to our translation repos (push done in our shell script). Are you sure that this works without those?

tmolitor-stud-tu (changes requested):
I noticed only 2 possible issues

.github/workflows/create-stable-pr.yml
matthewrfennell: I'd added it as I saw `github.rest.issues.addLabels` in `Add Label to Pull Request` and presumed it went through one of the issues endpoints.

However, I just tested, and it worked without this permission. So agreed, we can remove it

.github/workflows/update-translations.yml
matthewrfennell: Ah, quite right. Thanks for catching it!

tmolitor-stud-tu (approved):

PR #1376, opened 2025-01-23
Run xcbeautify in all workflows

matthewrfennell:
The Github Actions renderer is not useful in these cases since these scripts are not run against PRs. Therefore: just run standard `xcbeautify` with no extra arguments.

I tested, to see if the Github Actions renderer would add the comments to the commit view in Github, but it did not show anything.

I also tested the while loop in updateLocalization to make sure both `pipefail` and the piped command continue to work as expected.


tmolitor-stud-tu (approved):
Great, thanks! :)

PR #1377, opened 2025-01-23
Remove invalid XML characters before sending a message

lissine0:


Monal/Classes/MLXMPPManager.m
```diff
@@ -615,6 +615,16 @@ -(void) disconnectAll
     DDLogVerbose(@"manager disconnecAll done");
 }
 
+-(void) invalidateAllAccountStates
+{
+    NSArray* allAccounts = [[DataLayer sharedInstance] accountList];
+    for(NSDictionary* account in allAccounts)
+    {
+        NSMutableDictionary* dic = [[DataLayer sharedInstance] readStateForAccount:[account objectForKey:kAccountID]];
```
tmolitor-stud-tu: I wanted you to use `nil` as input, e.g. `[xmpp invalidateState:nil];`. Using `invalidateState` with a dictionary containing the state will NOT remove unsent message stanzas from state.

You  should also rename `invalidateAllAccountStates` to `resetAllAccountStates`, because using `nil` will do a full reset like if the account was freshly created.

Invalidating the state won't suffice, because afterwards, Monal will still have that message stanza containing invalid xml chars in its state and try to resend it. Our reset button should clear even that.

Why does invalidate not clean everything? - The invalidate method gets called if the format of the state is changed after an app upgrade (indicated by increasing its internal version number). We want to reset almost everything in that case, but not drop unsent message stanzas because that would result in a potential message loss on app upgrade.

tmolitor-stud-tu (changes requested):
Only one comment. I may be not specific enough on what that debug button should accomplish, I explained it now in more detail. Thanks again for helping out that quick!!

Monal/Classes/MLXMPPManager.m
```diff
@@ -615,6 +615,15 @@ -(void) disconnectAll
...
 }
 
+-(void) resetAllAccountStates
+{
+    NSArray* allAccounts = [[DataLayer sharedInstance] accountList];
+    NSMutableDictionary* dic = [xmpp invalidateState:nil];
+    for(NSDictionary* account in allAccounts)
+    {
+        [[DataLayer sharedInstance] persistState:dic forAccount:[account objectForKey:kAccountID]];
+    }
```
tmolitor-stud-tu:
```suggestion
    for(NSDictionary* account in allAccounts)
        [[DataLayer sharedInstance] persistState:[xmpp invalidateState:nil] forAccount:account[kAccountID]];
```

tmolitor-stud-tu (changes requested):

tmolitor-stud-tu (approved):

PR #1388, opened 2025-01-26
Avoid template injection in Github workflows

matthewrfennell:
GitHub workflow variables (`${{ ... }}`) use string replacement - simply replacing the variable with its value before the step gets executed.

Depending on where the input for these variables comes from, this makes it possible to pass malicious shell code that gets executed in the step.

It is safer to first pass the variables through an environment variable, and then from the environment variable to the script. This is because the shell will itself substitute the value, avoiding executing any code contained within it.

### DIFF
--- a/.github/workflows/create-stable-pr.yml
+++ b/.github/workflows/create-stable-pr.yml
@@ -71,17 +71,21 @@ jobs:
             const existingPR = pullRequests.find(pr => pr.labels.some(label => label.name === 'automated-pr'));
             console.log(`Existing PR: `, existingPR);
             if(existingPR)
-              return existingPR.number;
+              core.setOutput('prNumber', existingPR.number);
             else
-              return null;
+              core.setOutput('prNumber', null);
       - name: Create or Update Pull Request
         id: create_or_update_pr
         uses: actions/github-script@v7
+        env:
+          PR_NUMBER: ${{ steps.find_pr.outputs.prNumber }}
+          PR_TITLE: ${{ steps.get_commits.outputs.buildVersion }}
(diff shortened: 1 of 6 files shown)


.github/workflows/create-stable-pr.yml
```diff
@@ -71,17 +71,21 @@ jobs:
             const existingPR = pullRequests.find(pr => pr.labels.some(label => label.name === 'automated-pr'));
             console.log(`Existing PR: `, existingPR);
             if(existingPR)
-              return existingPR.number;
+              core.setOutput('prNumber', existingPR.number);
```
matthewrfennell, line 74: Returning output values from a `github-script`, and having them passed via environment variables to another `github-script` via `${{ ... .result }}` caused excessive quoting.

For instance, returning `null` would get treated in the resulting script as `"null"`, and returning `6.4.10` would get treated as `"6.4.10"`.

This caused conditional checks depending on these values to fail. This does not happen if you instead pass a named output variable as I've done here.

| Job | Bad example (return from script) | Good example (`core.setOutput(...)`) |
| - | - | - |
| create-stable-pr | https://github.com/matthewrfennell/Monal/actions/runs/12976231224/job/36188233230#step:6:183 | https://github.com/matthewrfennell/Monal/actions/runs/12976652561/job/36189174250#step:6:173 |
| pr-semver-title | https://github.com/matthewrfennell/Monal/actions/runs/12976540750/job/36188916701#step:4:12 | https://github.com/matthewrfennell/Monal/actions/runs/12976579382/job/36189005489#step:4:12

tmolitor-stud-tu: ah, thanks for fixing this as well! :)

PR #1406, opened 2025-04-04
Fixes for the integration of the new SwiftUI views in the app

lissine0:
This allows the navigation bar to be displayed on iPads

Note: this commit is only needed on the develop branch, as the new SwiftUI views are not in beta / stable yet.

Before this PR:
Change Password view
Blocked Users view
After this PR:
Change Password view
Blocked Users view

### DIFF
--- a/Monal/Classes/SwiftuiHelpers.swift
+++ b/Monal/Classes/SwiftuiHelpers.swift
@@ -736,18 +736,23 @@ class SwiftuiInterface : NSObject {
     @objc
     func makeOwnOmemoKeyView(_ ownContact: MLContact?) -> UIViewController {
         let host = UIHostingController(rootView:AnyView(EmptyView()))
-        if(ownContact == nil) {
-            host.rootView = AnyView(UIKitWorkaround(OmemoKeysView(omemoKeys: OmemoKeysForChat(viewContact: nil))))
-        } else {
-            host.rootView = AnyView(UIKitWorkaround(OmemoKeysView(omemoKeys: OmemoKeysForChat(viewContact: ObservableKVOWrapper<MLContact>(ownContact!)))))
+
+        @ViewBuilder
+        var omemoKeysView: some View {
+            if(ownContact == nil) {
+                OmemoKeysView(omemoKeys: OmemoKeysForChat(viewContact: nil))
+            } else {
+                OmemoKeysView(omemoKeys: OmemoKeysForChat(viewContact: ObservableKVOWrapper<MLContact>(ownContact!)))
+            }
         }
+        host.rootView = AnyView(UIKitWorkaround(omemoKeysView))
         return host
     }
 
     @objc
     func makeChangePasswordView(for accountID: NSNumber) -> UIViewController {
         let host = UIHostingController(rootView:AnyView(EmptyView()))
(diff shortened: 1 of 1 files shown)


tmolitor-stud-tu:
This is marked as draft, presumably because you did not yet check if that causes problems on iPhone?

lissine0:
I marked this as a draft, because I'm going to add a commit that fixes the same problem for the SwiftUI chatview on iPad / Mac, but I didn't decide how to do it yet.

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -736,18 +736,23 @@ class SwiftuiInterface : NSObject {
     @objc
     func makeOwnOmemoKeyView(_ ownContact: MLContact?) -> UIViewController {
         let host = UIHostingController(rootView:AnyView(EmptyView()))
-        if(ownContact == nil) {
-            host.rootView = AnyView(UIKitWorkaround(OmemoKeysView(omemoKeys: OmemoKeysForChat(viewContact: nil))))
-        } else {
-            host.rootView = AnyView(UIKitWorkaround(OmemoKeysView(omemoKeys: OmemoKeysForChat(viewContact: ObservableKVOWrapper<MLContact>(ownContact!)))))
+
```
tmolitor-stud-tu, line 760: Isn't this whole part a rewrite of the old code? Other than using a view builder, I see no difference between the old and the new code. But maybe I'm missing something important here?

Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -693,10 +693,27 @@ public extension UIViewController {
 // Interfaces between ObjectiveC/Storyboards and SwiftUI
 @objc
 class SwiftuiInterface : NSObject {
+    @StateObject private var sizeClass: ObservableKVOWrapper<SizeClassWrapper>
+    override init() {
```
tmolitor-stud-tu, line 697: are you sure this isn't missing a call to `super.init()` at the end?

tmolitor-stud-tu (changes requested):
I added two questions I'd like you to answer before I merge this.

Other than that: the commit "UIKitWorkaround: Fix displaying the navigation bar on large iphones in landscape mode" is really great, I like it :)

Monal/Classes/SwiftuiHelpers.swift
lissine0: This part is just adding the view builder, because I thought it's nice :D
I can remove this chunk if you want.

tmolitor-stud-tu: Well, the old code was shorter, but there is no need to change it. I was just curious :D

Monal/Classes/SwiftuiHelpers.swift
lissine0: You're right.
For some reason, the compiler didn't complain about its absence.
Fixed now.

tmolitor-stud-tu: Maybe the compiler calls it implicitly, but I don't know. Better to make it explicit, thanks! :)

PR #1408, opened 2025-04-04
Detect device migrations more reliably

### LINKED ISSUE #1401: Fix device uuid issues on macos
On macos, the apple-provided device uuid isn't as stable as the docs promise, leading to omemo deviceid (and other) resets.
Use this "this device only" keychain solution over here to get a stable deviceid: https://overflow.adminforge.de/questions/6993325/uidevice-uniqueidentifier-deprecated-what-to-do-now

TODO: check if that one even works for devices migrated using apples "hold new device next to your old device to automatically migrate everything to the new one".

lissine0:
I tested that this prevents generating new Omemo keys on reboot on macOS.
But I didn't test if the Omemo keys would be regenerated when an actual device migration happens.
Fixes #1401

### DIFF
--- a/Monal/Classes/DataLayerMigrations.m
+++ b/Monal/Classes/DataLayerMigrations.m
@@ -969,10 +969,8 @@
         if(![[HelperTools defaultsDB] boolForKey:@"isSandboxAPNS"])
         {
             NSString* stored_id = (NSString*)[db executeScalar:@"SELECT value FROM flags WHERE name='device_id';"];
-            NSString* current_id = UIDevice.currentDevice.identifierForVendor.UUIDString;
-            if(current_id == nil)
-                DDLogWarn(@"Deviceid is nil, not checking for deviceid change");
-            else if(![current_id isEqualToString:stored_id])
+            NSString* current_id = [[HelperTools deviceUUID] UUIDString];
+            if(![current_id isEqualToString:stored_id])
             {
                 DDLogWarn(@"Device id has changed (%@ --> %@), invalidating state AND omemo identity keys!", stored_id, current_id);
                 //invalidate account state because the app was migrated to a new device
@@ -989,10 +987,10 @@
                 [db executeNonQuery:@"DELETE FROM signalPreKey;"];
                 [db executeNonQuery:@"DELETE FROM signalSignedPreKey;"];
                 //update device id in db
-                [db executeNonQuery:@"UPDATE flags SET value=? WHERE name='device_id';" andArguments:@[UIDevice.currentDevice.identifierForVendor.UUIDString]];
+                [db executeNonQuery:@"UPDATE flags SET value=? WHERE name='device_id';" andArguments:@[current_id]];
             }
         }
-        
+
         //turn foreign keys on again
         //needed for sqlite >= 3.26.0 (see https://sqlite.org/lang_altertable.html point 2)
         [db executeNonQuery:@"PRAGMA legacy_alter_table=off;"];
--- a/Monal/Classes/HelperTools.m
+++ b/Monal/Classes/HelperTools.m
@@ -65,6 +65,7 @@ extern int64_t kscrs_getNextCrashReport(char* crashReportPathBuffer);
 @import CoreImage.CIFilterBuiltins;
 @import UIKit;
 @import AVFoundation;
+@import SAMKeychain;
 @import UniformTypeIdentifiers;
 @import QuickLookThumbnailing;
 
@@ -676,8 +677,8 @@ static void notification_center_logging(CFNotificationCenterRef center, void* ob
 +(NSDictionary<NSString*, NSString*>*) getInvalidPushServers
 {
     return @{
-        @"ios13push.monal.im": nilWrapper([[[UIDevice currentDevice] identifierForVendor] UUIDString]),
-        @"push.monal.im": nilWrapper([[[UIDevice currentDevice] identifierForVendor] UUIDString]),
+        @"ios13push.monal.im": nilWrapper([[HelperTools deviceUUID] UUIDString]),
+        @"push.monal.im": nilWrapper([[HelperTools deviceUUID] UUIDString]),
         @"us.prod.push.monal-im.org": nilWrapper(nil),
     };
 }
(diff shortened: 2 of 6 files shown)


Monal/Classes/DataLayerMigrations.m
```diff
@@ -963,18 +965,28 @@ FOREIGN KEY('account_id') REFERENCES 'account'('account_id') ON DELETE CASCADE \
...
-            if(current_id == nil)
-                DDLogWarn(@"Deviceid is nil, not checking for deviceid change");
-            else if(![current_id isEqualToString:stored_id])
+            // We save the device UUID once in a special keychain, which is bound to the device.
+            // (It can't be migrated to other devices)
+            // Thus, if we can't read the device UUID from the keychain, it means the app was
+            // migrated to a new device.
+
+            NSError* error;
+            NSString* deviceID __unused = [SAMKeychain passwordForService:kMonalDeviceUUIDKeychainName account:kAppGroup error:&error];
```
lissine0: Since we're storing only one thing in the keychain, I didn't know what to use as `account`. So I picked the kAppGroup (which can be `"group.monalalpha"` or `"group.quicksy"` or  `"group.monal"`)
Alternatively, we can use a constant string as `account`.

lissine0:
Semi-related: in https://github.com/monal-im/Monal/blob/4f43f66f1632/Monal/Classes/xmpp.m#L3093, is that usage of the potentially-instable device UUID okay? Or should we rather use the UUID stored in the keychain?

Monal/Classes/DataLayerMigrations.m
tmolitor-stud-tu: I think a constant string would be more suitable. The keychain is already scoped by app/appgroup.

tmolitor-stud-tu:
I think we should add a new HelperTools method, generating this stable id and use that HelperTools method everywhere `[[UIDevice currentDevice] identifierForVendor]` is used.

Monal/Classes/DataLayer.m
```diff
@@ -2555,4 +2555,13 @@ -(MLPromise*) getPromise:(MLPromise*) promise
...
 
+// clean up signal store and generate new omemo keys (but don't change trust settings!)
+-(void) cleanupSignalStore
+{
+    [self.db executeNonQuery:@"DELETE FROM signalContactSession;"];
+    [self.db executeNonQuery:@"DELETE FROM signalIdentity;"];
+    [self.db executeNonQuery:@"DELETE FROM signalPreKey;"];
+    [self.db executeNonQuery:@"DELETE FROM signalSignedPreKey;"];
+}
+
```
lissine0: I was going to add this method to `MLSignalStore`. But that class doesn't provide a singleton, and neither does `MLOMEMO`. In fact, `MLSignalStore` is tied to an `accountID` / `accountJid`

Monal/Classes/DataLayerMigrations.m
```diff
@@ -963,36 +963,21 @@ FOREIGN KEY('account_id') REFERENCES 'account'('account_id') ON DELETE CASCADE \
...
-                [db executeNonQuery:@"DELETE FROM signalPreKey;"];
-                [db executeNonQuery:@"DELETE FROM signalSignedPreKey;"];
-                //update device id in db
-                [db executeNonQuery:@"UPDATE flags SET value=? WHERE name='device_id';" andArguments:@[UIDevice.currentDevice.identifierForVendor.UUIDString]];
-            }
-        }
-        
+            // The method called here checks for device migrations and handles them
+            // (state invalidation, Omemo keys regeneration etc.)
+            [HelperTools deviceUUID];
```
tmolitor-stud-tu: that's not what I wanted.
I simply wanted the `[[UIDevice currentDevice] identifierForVendor]` replaced by a `HelperTools` method *only* returning that UUID from a keychain (and falling back to `[[UIDevice currentDevice] identifierForVendor]` if the keychain is empty, for backwards compatibility).

That method should not do any migrations etc. Also doing these migrations is a violation of the design principle of separation of concerns.

Monal/Classes/DataLayer.m
tmolitor-stud-tu: yes, that's right. But I don't think we need this method, once the `HelperTools` method for getting the uuid doesn't implement any migrations and the migration is done in `DataLayerMigrations.m` like before.

tmolitor-stud-tu (changes requested):
Please fix that. The result will be a smaller changeset, too :)

tmolitor-stud-tu (changes requested):
This looks much better now, but do we really need to check if the keychain is empty here? using the new `HelperTools` method and the old approach of using the `flags` table to store the old uuid and compare these two should do the same, no?

it would only be different, if `identifierForVendor` would return *the same* identifier after a migration, but imho that doesn't happen.

To recall: we try to fix an issue where the opposite is happening: the uuid is changing *without* a migration, and our new `HelperTools` method should already fix that, right?

I just want to keep the changeset as small as possible without introducing different behavior between the usage of our new `HelperTools` method in the `DataLayerMigrations` and other parts of our project (like `sendSMSRequest` or `getInvalidPushServers` etc.)

lissine0:
Please make sure to review https://github.com/monal-im/Monal/compare/b8142054a680f84ecc6986bbf012278400efa94e..526b386918a9a841ba00d4392b35eba69f43786b before merging.
edit: it should be okay.

Monal/Classes/DataLayerMigrations.m
```diff
@@ -989,10 +989,16 @@ FOREIGN KEY('account_id') REFERENCES 'account'('account_id') ON DELETE CASCADE \
                 [db executeNonQuery:@"DELETE FROM signalPreKey;"];
                 [db executeNonQuery:@"DELETE FROM signalSignedPreKey;"];
                 //update device id in db
-                [db executeNonQuery:@"UPDATE flags SET value=? WHERE name='device_id';" andArguments:@[UIDevice.currentDevice.identifierForVendor.UUIDString]];
+                [db executeNonQuery:@"UPDATE flags SET value=? WHERE name='device_id';" andArguments:@[current_id]];
+                //save the current device UUID in the special keychain (fixes issue #1401)
+                NSError* deviceUUIDSavingError;
+                [SAMKeychain setAccessibilityType:kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly];
```
tmolitor-stud-tu: saving the device id in the keychain should also be part of our HelperTools method.

Put simple: The HelperTools method should simply provide an interface having the same API as the `UIDevice.currentDevice.identifierForVendor` implementation, but providing stable IDs using the keychain hack while switching IDs when migrating the app to a new device.

tmolitor-stud-tu (changes requested):

Monal/Classes/HelperTools.m
```diff
@@ -2790,6 +2791,28 @@ +(NSString*) appBuildVersionInfoFor:(MLVersionType) type
...
+{
+    NSError* error;
+    NSString* deviceUUID = [SAMKeychain passwordForService:kMonalDeviceUUIDKeychainName account:kDeviceUUIDKeychainAccount error:&error];
+    if(error)
+    {
+        //the keychain is empty (due to a device migration for example)
+        NSUUID* newDeviceUUID = nilDefault([[UIDevice currentDevice] identifierForVendor], [NSUUID UUID]);
+        //save the new device UUID in the special keychain
+        NSError* deviceUUIDSavingError;
+        [SAMKeychain setAccessibilityType:kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly];
```
tmolitor-stud-tu, line 2807: Please add a short comment why the "this device only" accessibility type is necessary, possibly citing the stackoverflow post mentioning this workaround.

Monal/Classes/HelperTools.m
```diff
@@ -2790,6 +2791,28 @@ +(NSString*) appBuildVersionInfoFor:(MLVersionType) type
...
+// Similar to UIDevice.currentDevice.identifierForVendor, but provides stable IDs
+// Fixes issue #1401
++(NSUUID*) deviceUUID
+{
+    NSError* error;
+    NSString* deviceUUID = [SAMKeychain passwordForService:kMonalDeviceUUIDKeychainName account:kDeviceUUIDKeychainAccount error:&error];
+    if(error)
+    {
+        //the keychain is empty (due to a device migration for example)
+        NSUUID* newDeviceUUID = nilDefault([[UIDevice currentDevice] identifierForVendor], [NSUUID UUID]);
```
tmolitor-stud-tu, line 2803: great use of `nilDefault()` :)

Monal/Classes/DataLayerMigrations.m
```diff
@@ -12,6 +12,8 @@
 #import <monalxmpp/HelperTools.h>
 #import <monalxmpp/MLImageManager.h>
 
+@import SAMKeychain;
```
tmolitor-stud-tu: This can be removed now

Monal/Classes/DataLayerMigrations.m
```diff
@@ -969,10 +971,8 @@ FOREIGN KEY('account_id') REFERENCES 'account'('account_id') ON DELETE CASCADE \
         if(![[HelperTools defaultsDB] boolForKey:@"isSandboxAPNS"])
         {
             NSString* stored_id = (NSString*)[db executeScalar:@"SELECT value FROM flags WHERE name='device_id';"];
-            NSString* current_id = UIDevice.currentDevice.identifierForVendor.UUIDString;
```
tmolitor-stud-tu, line 972: great simplification :)

tmolitor-stud-tu (changes requested):
some very small things to be changed, thanks for fixing this longstanding bug! :)

PR #1410, opened 2025-04-18
Use the built-in ExyteChat Message view, instead of a custom one.

lissine0:


Monal/Classes/MLMessage.m
```diff
@@ -218,6 +218,16 @@ -(NSString*) id
     return [NSString stringWithFormat:@"%@|%@", self.accountID, self.messageDBId];
 }
 
+-(NSString*) senderID
+{
+    if(self.isMuc && self.participantJid.length == 0)
+        return self.contactDisplayName;
```
lissine0, line 224: If you're okay with adding `occupant-id` as a property to `MLMessage`, we can use the `occupant-id` as `senderID` in public rooms, and only fallback to nicks if occupant-ids are not available (for example if mam is disabled in the room)

In practice, this means that if a participant leaves a room after sending a message, and another participant joins with the same nick and immediately sends a message, the two message bubbles won't be styled as coming from the same user.

tmolitor-stud-tu: sure :) the `MLMessage` model class is only a rough sketch, not nearly as complete and refined as the `MLContact` one. Feel free to improve it as needed :)

You can use code similar to here, to decide which id to use, line 317 to line 377 in `MLMessageProcessor.m`: https://github.com/monal-im/Monal/blob/develop/Monal/Classes/MLMessageProcessor.m#L317

tmolitor-stud-tu:
I'm not sure, if using the original message bubbles is a good idea.

We need different background colors for different encryption levels, the checkmarks for "sent" and "received" states need to be inside the message bubble because users using background images in a chat should still be able to distinguish those checkmarks from the background image. And the markers like "Today" etc. should be readable with background images, too.

@matthewrfennell What do you think?

matthewrfennell:
Personally I think we should try and get these changes upstreamed instead of using our own builder (of course using the custom builder if we fail with that).

So far they have been very receptive to changes, and since these are not Monal-specific, I think we'd have a good chance of getting them accepted.

tmolitor-stud-tu:
I've read our todo list of doom and they indeed look very receptive. if that continues, bringing everything upstream might indeed be the best solution :)

PR #1412, opened 2025-04-24
New ChatView: Add support for sending and receiving messages

lissine0:
The PR also includes some other small fixes

### DIFF
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -335,6 +335,7 @@ struct ChatView: View {
         }
         .toolbarRole(.editor)       //make sure to never show the title of the previous view in the back bar button
         .addLoadingOverlay(overlay)
+        .navigationBarTitleDisplayMode(.inline)
         .onAppear {
             checkOmemoSupport(withAlert:false)
             
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -197,7 +197,10 @@ struct ChatView: View {
     
     var body: some View {
         ExyteChatView(messages: messages, chatType: .conversation, replyMode: .quote) { draft in
-            print("sending draft: \(String(describing:draft))")
+            guard let newMLMessage = MLXMPPManager.sharedInstance().sendMessageAndAddToHistory(message: draft.text, havingType: kMessageTypeText, toContact: self.contact.obj, isEncrypted: self.contact.isEncrypted, uploadInfo: nil) else {
+                return
+            }
+            messages.append(ChatViewMessage(newMLMessage))
         }
         .showNetworkConnectionProblem(false)
 //         .enableLoadMore(pageSize: 3) { message in
@@ -378,6 +381,19 @@ struct ChatView: View {
                         checkOmemoSupport(withAlert:true)
                     }
                 }
+            }
+        }
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("kMonalNewMessageNotice")).receive(on: RunLoop.main)) { notification in
+            DDLogVerbose("chat view got new message notice \(String(describing:notification.userInfo))")
+
+            guard let message = notification.userInfo?["message"] as? MLMessage else {
+                DDLogError("Notification without message");
+                return
+            }
+            if message.isEqual(to: self.contact.obj) {
+                // Do not insert based on delay timestamp because that
+                // would make it possible to fake history entries.
+                messages.append(ChatViewMessage(message))
             }
         }
     }
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -344,9 +344,12 @@ struct ChatView: View {
             checkOmemoSupport(withAlert:false)
             
             //TODO: load messages from db
-            let dbMessages = DataLayer.sharedInstance().messages(forContact:contact.contactJid, forAccount:contact.accountID) as! [MLMessage]
-            for msg in dbMessages {
-                messages.append(ChatViewMessage(msg))
+            // In some scenarios, onAppear can be executed more than once without the state variables being cleared.
+            if messages.isEmpty {
+                let dbMessages = DataLayer.sharedInstance().messages(forContact:contact.contactJid, forAccount:contact.accountID) as! [MLMessage]
+                for msg in dbMessages {
+                    messages.append(ChatViewMessage(msg))
+                }
             }
 //             messages = [
 //                 ExyteChat.Message(
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -244,7 +244,7 @@ struct ChatView: View {
                                     Text(contact.contactDisplayName as String)
                                         .fontWeight(.semibold)
                                         .font(.headline)
-                                        .foregroundColor(.black)
+                                        .foregroundColor(.primary)
                                     
                                     if (contact.isTyping as Bool) {
                                         Text("Typing...")
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -199,6 +199,7 @@ struct ChatView: View {
         ExyteChatView(messages: messages, chatType: .conversation, replyMode: .quote) { draft in
             print("sending draft: \(String(describing:draft))")
         }
+        .showNetworkConnectionProblem(false)
 //         .enableLoadMore(pageSize: 3) { message in
 //             print("load more messages before: \(String(describing:message))")
 //         }
--- a/Monal/Classes/MLXMPPManager.h
+++ b/Monal/Classes/MLXMPPManager.h
@@ -12,6 +12,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @class xmpp;
 @class MLContact;
+@class MLMessage;
 
 /**
  A singleton to control all of the active XMPP connections
@@ -104,9 +105,10 @@ NS_ASSUME_NONNULL_BEGIN
 -(void) updatePassword:(NSString*) password forAccount:(NSNumber*) accountID;
 
 /**
-Sends a message to a specified contact in account. Calls completion handler on success or failure.
+ Sends a message to a specified contact in account.
+ It returns an MLMessage object (corresponding to the sent message) on success, and nil on failure.
  */
--(void) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo withCompletionHandler:(void (^ _Nullable)(BOOL success, NSString* messageId)) completion;
+-(MLMessage* _Nullable) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo NS_SWIFT_NAME(sendMessageAndAddToHistory(message:havingType:toContact:isEncrypted:uploadInfo:));
 -(void)sendMessage:(NSString*) message toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted isUpload:(BOOL) isUpload messageId:(NSString*) messageId withCompletionHandler:(void (^ _Nullable)(BOOL success, NSString* messageId)) completion;
 -(void) sendChatState:(BOOL) isTyping toContact:(MLContact*) contact;
 
--- a/Monal/Classes/MLXMPPManager.m
+++ b/Monal/Classes/MLXMPPManager.m
@@ -674,7 +674,7 @@ static const int pingFreqencyMinutes = 5;       //about the same Conversations u
 }
 
 #pragma mark -  XMPP commands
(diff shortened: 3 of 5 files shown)


Monal/Classes/ChatView.swift
```diff
@@ -195,10 +195,38 @@ struct ChatView: View {
 #endif
 }
     
+    // Only for messages going out
+    private func addMessage(message: String, messageID: String, messageType: String, mimeType: String? = nil, size: NSNumber? = nil) {
```
tmolitor-stud-tu: Why don't you name it "addOutgoingMessage" if it is only for outgoing ones?

Monal/Classes/ChatView.swift
```diff
@@ -195,10 +195,38 @@ struct ChatView: View {
...
+            messages.append(ChatViewMessage(newMLMessage))
+            // Create and donate interaction to allow for share suggestions
+            MLNotificationManager.sharedInstance().donateInteraction(forOutgoingDBId: messageHistoryID)
+    }
+
     var body: some View {
         ExyteChatView(messages: messages, chatType: .conversation, replyMode: .quote) { draft in
-            print("sending draft: \(String(describing:draft))")
+            let newMessageID: String = UUID().uuidString
+            self.addMessage(message: draft.text, messageID: newMessageID, messageType: kMessageTypeText)
```
tmolitor-stud-tu: I think this is a good opportunity to add the complete code of `addMessage` to `MLXMPPManager.sharedInstance().sendMessage()` and let it return an `MLMessage` instance.

tmolitor-stud-tu: Additionally: this should be done in one db transaction (adding to history db AND sending the message), you can use the `DataLayer` method `-(void) createTransaction:(monal_void_block_t) block;` to accomplish this.

Monal/Classes/ChatView.swift
```diff
@@ -378,6 +407,19 @@ struct ChatView: View {
                 }
             }
         }
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("kMonalNewMessageNotice")).receive(on: RunLoop.main)) { notification in
+            DDLogVerbose("chat view got new message notice \(notification.userInfo)")
```
tmolitor-stud-tu: This produces the warning `expression implicitly coerced from '[AnyHashable : Any]?' to 'Any'
`. Use this instead:

```suggestion
            DDLogVerbose("chat view got new message notice \(String(describing:notification.userInfo))")
```

Monal/Classes/ChatView.swift
```diff
@@ -378,6 +407,19 @@ struct ChatView: View {
...
+
+            guard let message = notification.userInfo?["message"] as? MLMessage else {
+                DDLogError("Notification without message");
+                return
+            }
+            if message.isEqual(to: self.contact.obj) {
+                // Do not insert based on delay timestamp because that
```
tmolitor-stud-tu, line 397: Thanks for adding this comment as a reminder! :)

tmolitor-stud-tu (changes requested):

Monal/Classes/ChatView.swift
lissine0: > I think this is a good opportunity to add the complete code of `addMessage` to `MLXMPPManager.sharedInstance().sendMessage()` and let it return an `MLMessage` instance.

I found an existing method in MLXMPPManager (`sendMessageAndAddToHistory`) that does this. I modified it so it returns MLMessage.

> Additionally: this should be done in one db transaction (adding to history db AND sending the message)

Do you still want this? The way `sendMessageAndAddToHistory` is written, the message is never sent if adding it to history db failed.
But in that scenario, no feedback is given to the user; the message simply never shows up in the ChatView. Do failing transactions send a notification to the user?

tmolitor-stud-tu: Yes. If the app crashes/gets killed between adding the history entry and sending the message, the message would never be send, but still displayed in the chat (without the sent/received checkmarks ever showing up and without an error (i) either).

Adding these to a single transaction makes sure, that either both, the message and the history entry, are added to the db or nothing at all, preventing this type of inconsistency.

Monal/Classes/MLXMPPManager.h
```diff
@@ -7,6 +7,7 @@
 //
 
 #import <Foundation/Foundation.h>
+#import <monalxmpp/MLMessage.h>
```
tmolitor-stud-tu: Use a `@class` forward declaration rather than pulling in the complete `MLMessage` definition and all header file this header might itself import.

Always only add imports to header files if absolutely necessary, to prevent dependency loops forming and improve compilation speed.

Monal/Classes/MLXMPPManager.h
```diff
@@ -106,7 +107,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
 Sends a message to a specified contact in account. Calls completion handler on success or failure.
  */
--(void) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo withCompletionHandler:(void (^ _Nullable)(BOOL success, NSString* messageId)) completion;
+-(MLMessage* _Nullable) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo withCompletionHandler:(void (^ _Nullable)(BOOL success, NSString* messageId)) completion;
```
tmolitor-stud-tu: Please use `NS_SWIFT_NAME` to rename this function to something better than `sendMessageAnd(addToHistory: ...)` Maybe `sendMessageAndAddToHistory(`

Monal/Classes/MLXMPPManager.m
```diff
@@ -699,20 +699,37 @@ -(void) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) me
     // Send message
     if(messageDBId != nil)
     {
-        DDLogInfo(@"Message added to history with id %ld, now sending...", (long)[messageDBId intValue]);
+        DDLogVerbose(@"Message added to history with id %ld", (long)[messageDBId intValue]);
+
+        // MLMessage object that will be returned by the method
+        MLMessage* newMLMessage = [[DataLayer sharedInstance] messageForHistoryID:messageDBId];
```
tmolitor-stud-tu, line 705: If we are in a single db transaction, this should NEVER return `nil` after inserting the message in the same transaction.

Monal/Classes/MLXMPPManager.m
```diff
@@ -674,7 +674,7 @@ -(NSString*) getPasswordForAccount:(NSNumber*) accountID
 }
 
 #pragma mark -  XMPP commands
--(void) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo withCompletionHandler:(void (^ _Nullable)(BOOL success, NSString* messageId)) completion
+-(MLMessage*) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo withCompletionHandler:(void (^ _Nullable)(BOOL success, NSString* messageId)) completion
```
tmolitor-stud-tu: We should move away from these completions and return a promise instead. This promise can return the `MLMessage` object. Existing code can use this `MLMessage` to extract the db id. And the `then()` versus `catch()` handler to deduce the `success` boolean.

Promises can even be used in async code from swift by using `returnedPromise.toTypedPromise().asyncOnMainActor()`

tmolitor-stud-tu (changes requested):

Monal/Classes/MLXMPPManager.m
lissine0: Why not simply remove the completions, and test if the returned `MLMessage` is nil to deduce the `success` boolean? (and use the `MLMessage` to extract the db id)

tmolitor-stud-tu: You are right, I thought the completion was due to some async code dispatching to some queue, but the code is synchronous all the way down. The only thing not being synchronous ist the dispatch to the receive queue in `xmpp.m`: `-(void) send:(MLXMLNode*) stanza withSmacks:(BOOL) withSmacks`.

But that dispatch is okay. We could make that method return a promise and wait for it, but I don't know what the benefit of that would be. The only thing I wanted, was to do the history upgrade and the sending (adding the stanza to the smacks queue and persisting that to the db) in one single db transaction, but that dispatch to the receive queue (which is there for a reason) makes using a single transaction overly complicated.

--> return a `MLMessage` from `sendMessageAndAddToHistory`, fix the code using that method and don't use a dedicated transaction int that method.

Monal/Classes/MLXMPPManager.m
tmolitor-stud-tu: No single transaction anymore, see my other comment

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Unfortunately using a single db transaction turns out to be more complicated than I initially thought, so now I opt for no dedicated transaction.

Monal/Classes/MLXMPPManager.h
```diff
@@ -104,9 +105,10 @@ NS_ASSUME_NONNULL_BEGIN
 -(void) updatePassword:(NSString*) password forAccount:(NSNumber*) accountID;
 
 /**
-Sends a message to a specified contact in account. Calls completion handler on success or failure.
+ Sends a message to a specified contact in account.
+ It returns an MLMessage object (corresponding to the sent message) on success, and nil on failure.
  */
--(void) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo withCompletionHandler:(void (^ _Nullable)(BOOL success, NSString* messageId)) completion;
+-(MLMessage* _Nullable) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo NS_SWIFT_NAME(sendMessageAndAddToHistory(message:havingType:toContact:isEncrypted:uploadInfo:));
```
tmolitor-stud-tu, line 111: Normally the NS_SWIFT_NAME is added on top of the prototype, not to its end.

```suggestion
NS_SWIFT_NAME(sendMessageAndAddToHistory(message:havingType:toContact:isEncrypted:uploadInfo:))
-(MLMessage* _Nullable) sendMessageAndAddToHistory:(NSString*) message havingType:(NSString*) messageType toContact:(MLContact*) contact isEncrypted:(BOOL) encrypted uploadInfo:(NSDictionary* _Nullable) uploadInfo;
```

Monal/Classes/MonalAppDelegate.m
```diff
@@ -1725,29 +1727,33 @@ -(void) sendAllOutboxes
...
+                NSDictionary* mapping = @{
+                    @"text": kMessageTypeText,
+                    @"url": kMessageTypeUrl,
+                    @"geo": kMessageTypeGeo,
+                    @"image": kMessageTypeFiletransfer,
+                    @"file": kMessageTypeFiletransfer,
+                    @"contact":kMessageTypeFiletransfer,
+                    @"audiovisual":kMessageTypeFiletransfer,
+                };
+                NSString* messageType = mapping[payload[@"type"]];
```
tmolitor-stud-tu: This will crash with a nil error, if `payload` doesn't have a key named "type" whereas the old code did not crash but reached the unreachable in the else path, providing a more informative error message. You can use `nilDefault` to fix this, though.

```suggestion
                NSString* messageType = mapping[nilDefault(payload[@"type"], @"<type key not present>")];
```

Monal/Classes/MonalAppDelegate.m
```diff
@@ -1725,29 +1727,33 @@ -(void) sendAllOutboxes
...
+                        DDLogInfo(@"SHARESHEET_SEND_DATA success=NO, account=%@", account.accountID);
+
+                    cleanup(payload);
                 }
-                else if([payload[@"type"] isEqualToString:@"image"] || [payload[@"type"] isEqualToString:@"file"] || [payload[@"type"] isEqualToString:@"contact"] || [payload[@"type"] isEqualToString:@"audiovisual"])
+                else
                 {
+                    //the payload type is either "image", "file", "contact" or "audiovisual"
                     DDLogInfo(@"Got %@ upload: %@", payload[@"type"], payload[@"data"]);
                     [self.activeChats.currentChatView showUploadHUD];
```
tmolitor-stud-tu, line 1758: Reminder: this one will not work because `self.activeChats.currentChatView` points to the old `UIKit` chatview. See this todo: https://github.com/monal-im/Monal/blob/develop/Monal/Classes/ActiveChatsViewController.m#L1554

tmolitor-stud-tu (changes requested):
Only minor changes

Monal/Classes/MLXMPPManager.h
tmolitor-stud-tu: Oh, forget it, that was the KSCrash project where I'm currently trying to get some changes upstream. Our codebase uses the same "convention" as you did (add it to the end). "convention", because there is only one instance where `NS_SWIFT_NAME` is used :D

lissine0: Apple puts NS_SWIFT_NAME before for class names, and after for everything else
https://developer.apple.com/documentation/swift/renaming-objective-c-apis-for-swift

PR #1421, opened 2025-04-29
New ChatView: Update the unread counter

lissine0:


Monal/Classes/ChatView.swift
```diff
@@ -374,6 +377,15 @@ struct ChatView: View {
...
 //             ]
         }
+        .onDisappear {
+            // When the split view is active, selecting different chats results in the old
+            // view's onDisappear executing after the new view's onAppear, thus erroneously
+            // making currentContact always nil.
+            // This if statement ensures currentContact has the correct value.
+            if MLNotificationManager.sharedInstance().currentContact == self.contact.obj {
+                MLNotificationManager.sharedInstance().currentContact = nil
+            }
```
lissine0, line 387: Do we need a way to ensure these two instruction are atomically executed together?
The onAppear of the other (new) view is executed in the same thread as this onDisappear (of the old view), so I think a critical section not needed?

tmolitor-stud-tu: you could async dispatch all changes to the `currentContact` to the main thread, to be sure.

tmolitor-stud-tu: but then again, I think onAppear and onDisappear are always called from the main thread, so no need to do a second (async) dispatch :)

tmolitor-stud-tu (approved):

PR #1424, opened 2025-05-06
Display media attachments in the new ChatView

lissine0:


Monal/Classes/HelperTools.m
```diff
@@ -324,7 +324,7 @@ void swizzle(Class c, SEL orig, SEL new)
 
 static void notification_center_logging(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
 {
-    DDLogDebug(@"NSNotification %@ with %@: %@", name, object, userInfo);
+    DDLogDebug(@"NSNotification %@ with %p: %@", name, object, userInfo);
```
lissine0: I'm not sure what a proper fix would be.
The notification objects of at least these notifications can't be displayed (trying to display them causes a crash): 
`FPM_PlayableRangeChanged`, `FPM_StreamLikelyToKeepUp`, `MentorPrerollComplete`, `MentorStoppingDueToCompletion`, `MentorPausingDueToHighWaterLevel` and `MentorResumingAfterHighWaterLevel`
These are all received when ExyteChat is playing audio or a video
They seem to originate from Apple's CoreMedia

lissine0: I also get the same crash when using the UIKit chat view on the iOS 18.2 simulator, when an audio file is in the chat.
The notification in this case was `FPM_PlayableRangeChanged`

Monal/Classes/ChatView.swift
```diff
@@ -415,6 +416,20 @@ struct ChatView: View {
         .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("kMonalRefresh")).receive(on: RunLoop.main)) { notification in
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
         }
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("kMonalMessageFiletransferUpdateNotice")).receive(on: RunLoop.main)) { notification in
+            DDLogVerbose("chat view got file transfer update notice \(String(describing:notification.userInfo))")
+
+            guard let message = notification.userInfo?["message"] as? MLMessage else {
+                DDLogError("Notification without message");
```
tmolitor-stud-tu: I think this should be an assertion to catch implementation errors in notification senders, what do you think?

Monal/Classes/ChatView.swift
```diff
@@ -415,6 +416,20 @@ struct ChatView: View {
...
+                return
+            }
+            for index in messages.indices {
+                if messages[index].message.messageDBId == message.messageDBId {
+                    messages[index] = ChatViewMessage(message)
```
tmolitor-stud-tu: Shouldn't this be done automatically if the underlying `MLMessage` is changed to a auto-updating singleton like done with `MLContact`?

Monal/Classes/ChatView.swift
```diff
@@ -424,7 +439,47 @@ class ChatViewMessage: ExyteChat.Message {
...
+                    Link: \(message.messageText)
+                    """
+                super.init(id: message.id, user: user, createdAt: message.timestamp, text: messageText)
+            } else {
+                // The file was already downloaded
+                let cacheFile = info?["cacheFile"] as! String
+
+                switch message.filetransferMimeType {
+                    case let mimeType where mimeType.starts(with: "image/"):
+                        let attachment = Attachment(id: UUID().uuidString, url: URL(fileURLWithPath: cacheFile), type: .image)
```
tmolitor-stud-tu: Please use `HelperTools` method `stringToUUID:`  to generate attachment uuids using the file path of the cacheFile. That way, the attachment ids will be stable :)

Monal/Classes/ChatView.swift
```diff
@@ -424,7 +439,47 @@ class ChatViewMessage: ExyteChat.Message {
...
+                        let attachment = Attachment(id: UUID().uuidString, url: URL(fileURLWithPath: cacheFile), type: .video, mimeType: mimeType)
+                        super.init(id: message.id, user: user, createdAt: message.timestamp, attachments: [attachment])
+                    case let mimeType where mimeType.starts(with: "audio/"):
+                        let recording = Recording(url: URL(fileURLWithPath: cacheFile), mimeType: mimeType)
+                        super.init(id: message.id, user: user, createdAt: message.timestamp, recording: recording)
+                    default:
+                        // Display the file:// URL instead of the upload URL for other attachment types.
+                        // This allows them to be displayed in the browser, even if they're encrypted
+                        // or if they expired on the server.
+                        super.init(id: message.id, user: user, createdAt: message.timestamp, text: "file://" + cacheFile)
```
tmolitor-stud-tu:
```suggestion
                        //TODO: show proper "open" button instead of file links
                        super.init(id: message.id, user: user, createdAt: message.timestamp, text: "file://" + cacheFile)
```

Monal/Classes/ChatView.swift
```diff
@@ -424,7 +439,47 @@ class ChatViewMessage: ExyteChat.Message {
...
+            super.init(id: message.id, user: user, createdAt: message.timestamp, text: message.messageText)
+        } else {
+            let info = MLFiletransfer.getFileInfo(for: message)
+            let needsDownloading = info?["needsDownloading"] as? Bool ?? true
+            if needsDownloading && (info?["mimeType"] == nil || info?["size"] == nil) {
+                // The file size and mimeType were not checked yet
+                MLFiletransfer.checkMimeTypeAndSize(forHistoryID: message.messageDBId)
+                super.init(id: message.id, user: user, createdAt: message.timestamp, text: "[File transfer; checking the size...]")
+            } else if needsDownloading {
+                // We already know the file's size / mimeType and it wasn't automatically downloaded (yet)
```
tmolitor-stud-tu:
```suggestion
                // We already know the file's size / mimeType and it wasn't automatically downloaded (yet)
                // TODO: show proper "download file" button instead of automatically triggering this
```

Monal/Classes/ChatView.swift
```diff
@@ -424,7 +439,47 @@ class ChatViewMessage: ExyteChat.Message {
     init(_ message: MLMessage) {
         self.message = message
         let user = ExyteChat.User(id: message.senderID, name: message.contactDisplayName, avatarURL: nil, isCurrentUser: !message.inbound)
-        super.init(id: message.id, user: user, createdAt: message.timestamp, text: message.messageText)
+        if message.messageType != kMessageTypeFiletransfer {
+            super.init(id: message.id, user: user, createdAt: message.timestamp, text: message.messageText)
```
tmolitor-stud-tu: Not all messages not being a filetransfer are text messages, we have:
```
kMessageTypeStatus
(kMessageTypeMessageDraft)
kMessageTypeText
kMessageTypeGeo
kMessageTypeUrl
kMessageTypeFiletransfer
```

Showing those as text should at least be annotated by a "TODO" comment.

Monal/Classes/ChatView.swift
```diff
@@ -424,7 +439,47 @@ class ChatViewMessage: ExyteChat.Message {
...
+        } else {
+            let info = MLFiletransfer.getFileInfo(for: message)
+            let needsDownloading = info?["needsDownloading"] as? Bool ?? true
+            if needsDownloading && (info?["mimeType"] == nil || info?["size"] == nil) {
```
tmolitor-stud-tu:
```suggestion
            // TODO: show a proper button for mime type checks instead of doing this automatically
            if needsDownloading && (info?["mimeType"] == nil || info?["size"] == nil) {
```

Monal/Classes/HelperTools.m
tmolitor-stud-tu: yeah, some notifications can crash that logging. I think the proper fix would be to check if the thing at `*object` is a `NSObject` (= an objc class for which we can use `%@`) or not (and use `%p` in that case).

tmolitor-stud-tu (changes requested):

Monal/Classes/HelperTools.m
lissine0: It looks like we'll need to use the code from https://www.cocoawithlove.com/2010/10/testing-if-arbitrary-pointer-is-valid.html to do that.

tmolitor-stud-tu: I think this one is more complete and a bit less hacky than the solution using the signal handlers: https://blog.timac.org/2016/1124-testing-if-an-arbitrary-pointer-is-a-valid-objective-c-object/

Monal/Classes/ChatView.swift
lissine0: Why do we need to show a button for mime type checks?
We want to do it automatically, because `MLFiletransfer.checkMimeTypeAndSize(forHistoryID:)` does the auto-downloading if the settings permit it.

Monal/Classes/ChatView.swift
lissine0: Is it okay to use both `guard` and `MLAssert` at the same time? (see the version I pushed)
I like `guard` because after using it I don't need to use force-unwrapping

Monal/Classes/HelperTools.m
lissine0: The link you posted relies on internal symbols, which changed since the article was written. (In fact they also changed when the author was drafting the article)
That's why I prefer the link I posted, unless you think it has a bigger performance overhead.

tmolitor-stud-tu: Its not the performance overhead, but those signals will likely conflict with the `KSCrash` crash recording/reporting.

What internal symbols did change? If we only want to check if a pointer is valid, we only need to know if its a tagged one, not what class the exact tag points to. And that "ABI" is pretty stable (`(pointer % 2) == 1`). In fact we just need to implement the python functions `is_valid_pointer` and `is_allowed_pointer` in objc and some other stuff from this post, but we **don't need the classname nor do we need to check if the class exists**. 

Why? Because the objc runtime will do that for us when using `%@`. We need to only make sure, that the pointer is valid and points to readable memory to prevent a `SIGSEGV` crash when accessing it through `%@`.

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Sure, just do a `MLAssert(false, "error message")` in the else branch of your guard clause then :)

Monal/Classes/ChatView.swift
tmolitor-stud-tu: See my other answer to matthew in the other PR: https://github.com/monal-im/Monal/pull/1426#discussion_r2087780948

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Because we don't want to auto-download things in the following cases. Remember: fetching the mime-type from the server already leaks our IP to some arbitrary webserver not controlled by the admin of *our* xmpp server, but by the admin of the *sender's* xmpp server.

1. The user turned off auto-download in the setting, but now they should still be able to manually trigger a download (e.g. first fetch the mime-type and size which should be displayed in a new button which triggers the actual download; this makes sure the user can decide if they really want to download that 6GB file or not).
2. We don't want to auto-download media in public channels for security and privacy reasons (yes afaik we currently do this, but that's an improvement I postponed for the chatview rewrite, e.g. now).

Monal/Classes/HelperTools.m
lissine0: I tried the approach of https://blog.timac.org/2016/1124-testing-if-an-arbitrary-pointer-is-a-valid-objective-c-object/ but all crash-causing notifications are false positives.
(In these cases the pointers are not tagged. But they are aligned and point to valid and readable memory. So, the method returns true) 

<details>
<summary>
For reference, here's the diff:
</summary>

```diff
diff --git a/Monal/Classes/HelperTools.m b/Monal/Classes/HelperTools.m
index dcec25b8b..e9626aefb 100644
--- a/Monal/Classes/HelperTools.m
+++ b/Monal/Classes/HelperTools.m
@@ -322,9 +322,108 @@ void swizzle(Class c, SEL orig, SEL new)
         method_exchangeImplementations(origMethod, newMethod);
 }
 
+/**
+ Test if the pointer points to readable and valid memory.
+ 
+ @param inPtr is the pointer
+ @return true if the pointer points to readable and valid memory.
+ */
+static bool IsValidReadableMemory(const void *inPtr)
+{
+    kern_return_t error = KERN_SUCCESS;
+     
+    // Check for read permissions
+    bool hasReadPermissions = false;
+     
+    vm_size_t vmsize;
+    vm_address_t address = (vm_address_t)inPtr;
+    vm_region_basic_info_data_t info;
+    mach_msg_type_number_t info_count = VM_REGION_BASIC_INFO_COUNT_64;
+ 
+    memory_object_name_t object;
+ 
+    error = vm_region_64(mach_task_self(), &address, &vmsize, VM_REGION_BASIC_INFO, (vm_region_info_t)&info, &info_count, &object);
+    if(error != KERN_SUCCESS)
+    {
+        // vm_region/vm_region_64 returned an error
+        hasReadPermissions = false;
+    }
+    else
+    {
+        hasReadPermissions = (info.protection & VM_PROT_READ);
+    }
+     
+    if(!hasReadPermissions)
+    {
+        return false;
+    }
+     
+    // Read the memory
+    vm_offset_t readMem = 0;
+    mach_msg_type_number_t size = 0;
+    error = vm_read(mach_task_self(), (vm_address_t)inPtr, sizeof(uintptr_t), &readMem, &size);
+    if(error != KERN_SUCCESS)
+    {
+        // vm_read returned an error
+        return false;
+    }
+     
+    return true;
+}
+
+/**
+ Test if a pointer is an Objective-C object
+ 
+ @param inPtr is the pointer to check
+ @return true if the pointer is an Objective-C object
+ */
+static bool IsObjcObject(const void *inPtr)
+{
+    // NULL pointer is not an Objective-C object
+    if(inPtr == NULL)
+    {
+        return false;
+    }
+     
+    // Check for tagged pointers
+    const unsigned long long TAG_MASK = (1ULL<<63); //the tag bit is the MSB
+    if(((uintptr_t)inPtr & TAG_MASK) == TAG_MASK)
+    {
+        return true;
+    }
+     
+    // Check if the pointer is aligned
+    if (((uintptr_t)inPtr % sizeof(uintptr_t)) != 0)
+    {
+        return false;
+    }
+     
+    // From LLDB:
+    // Objective-C runtime has a rule that pointers in a class_t will only have bits 0 thru 46 set
+    // so if any pointer has bits 47 thru 63 high we know that this is not a valid isa
+    // See https://github.com/llvm/llvm-project/blob/main/lldb/examples/summaries/cocoa/objc_runtime.py
+    if(((uintptr_t)inPtr & 0xFFFF800000000000) != 0)
+    {
+        return false;
+    }
+     
+    // Check if the memory is valid and readable
+    if(!IsValidReadableMemory(inPtr))
+    {
+        return false;
+    }
+
+    return true;
+}
+
 static void notification_center_logging(CFNotificationCenterRef center, void* observer, CFStringRef name, const void* object, CFDictionaryRef userInfo)
 {
-    DDLogDebug(@"NSNotification %@ with %@: %@", name, object, userInfo);
+    if(IsObjcObject(object))
+        DDLogDebug(@"NSNotification %@ with %@: %@", name, object, userInfo);
+    else
+        DDLogDebug(@"NSNotification %@ with %@", name, userInfo);
 }
 
 @implementation WeakContainer
```
</details>

I suggest that we don't log the notification `object` because in most notifications we send, the `object` is either an xmpp instance,  the class instance that's sending the notification (i.e. `self`), or `nil`. And the log viewer already gives us this information.

The only exceptions are https://github.com/monal-im/Monal/blob/f92692ce50cc00e2c7bdc89c2fdc941be97c93fb/Monal/Classes/MLImageManager.m#L374  https://github.com/monal-im/Monal/blob/f92692ce50cc00e2c7bdc89c2fdc941be97c93fb/Monal/Classes/MLVoIPProcessor.m#L116  https://github.com/monal-im/Monal/blob/f92692ce50cc00e2c7bdc89c2fdc941be97c93fb/Monal/Classes/MLVoIPProcessor.m#L125  https://github.com/monal-im/Monal/blob/f92692ce50cc00e2c7bdc89c2fdc941be97c93fb/Monal/Classes/MLMessageProcessor.m#L886 https://github.com/monal-im/Monal/blob/f92692ce50cc00e2c7bdc89c2fdc941be97c93fb/Monal/Classes/IPC.m#L228
But the first three already have other log statements printing the `object`. And the last one has the object inside the `userInfo` dictionary.
So, I'd say there's no harm in removing the `object` from the log statement.

tmolitor-stud-tu: Thing is: logging our own notifications is kind of moot, because we already log them when sending them (most of the time). The real reason why I added this notification observer was to log notifications coming from apple frameworks or other libs. I'm using this in the alpha (the observer is only active in alpha builds anyways) to learn more about the internal workings of SwiftUI, UIKit etc. and I find this really useful.

I think the solution (in addition to the tagged pointer checking etc.) would be to blacklist the notifications you listed above and don't log the object only for these (log something like `<unprintable>` or somesuch to distinguish them from real `nil`.

Monal/Classes/DataLayerMigrations.m
```diff
@@ -988,6 +990,30 @@ FOREIGN KEY('message_history_id') REFERENCES 'message_history'('message_history_
             [db executeNonQuery:@"INSERT INTO flags (name, value) VALUES('autodecrement~message_history', ?);" andArguments:@[initialValue]];
         }];
 
+        [self updateDB:db withDataLayer:dataLayer toVersion:7.004 withBlock:^{
+            [db executeNonQuery:@"ALTER TABLE message_history ADD COLUMN media_duration REAL NULL DEFAULT NULL;"];
+
+            // Compute the media_duration of old audio attachments and update the db
```
tmolitor-stud-tu: This should be moved to HelperTools

Monal/Classes/MLFiletransfer.m
```diff
@@ -283,15 +289,47 @@ +(void) downloadFileForHistoryID:(NSNumber*) historyId andForceDownload:(BOOL) f
             DDLogDebug(@"Updating db and sending out kMonalMessageFiletransferUpdateNotice");
             //update db with content type and size
             [[DataLayer sharedInstance] setMessageHistoryId:historyId filetransferMimeType:mimeType filetransferSize:filetransferSize];
+
+            NSDictionary* fileInfo = [self getFileInfoForMessage:msg];
```
tmolitor-stud-tu: the fileinfo dictionary should become a real objc model class...that has been low priority item I wanted to do for quite some time, would be a good opportunity to do that now

Monal/Classes/MLFiletransfer.m
```diff
@@ -570,6 +608,11 @@ +(void) deleteFileForMessage:(MLMessage*) msg
     {
         DDLogDebug(@"Deleting file in cache: %@", info[@"cacheFile"]);
         [_fileManager removeItemAtPath:info[@"cacheFile"] error:nil];
+        if([info[@"mimeType"] hasPrefix:@"video/"])
+        {
+            DDLogVerbose(@"Deleting video thumbnail stored at %@", msg.thumbnailURL.path);
```
tmolitor-stud-tu: use the new fileinfo model class rather than mixing in the media stuff into MLMessage

Monal/Classes/MLImageManager.m
```diff
@@ -224,6 +224,48 @@ -(NSString*) fileNameforContact:(MLContact*) contact
...
 
+-(NSString*) fileNameforThumbnailOfMessage:(MLMessage*) message
+{
+    return [NSString stringWithFormat:@"%@.png", message.messageDBId.stringValue];
+}
+
+-(void) setThumbnailOfMessage:(MLMessage*) message withData:(NSData* _Nullable) data
+{
+    // Notify message.thumbnailURL's observers that it's about to change (needed for KVO compliance)
+    [message willChangeValueForKey:@"thumbnailURL"];
```
tmolitor-stud-tu: with a proper design, you should almost never have to do this. so this is a strong hint that there probably is a better design/structure for these things: the fileinfo model class.

tmolitor-stud-tu: also: when using proper properties, setting a property value (through self.popertyName = value) automatically generates KVO notifications (using the underlying _propertyName instance variable won't).

Monal/Classes/MLMessage.h
```diff
@@ -15,6 +15,12 @@ NS_ASSUME_NONNULL_BEGIN
 @class MLContact;
 @class MLReactionsEntry;
 
+typedef NS_ENUM(NSUInteger, MLMessageDownloadState) {
```
tmolitor-stud-tu: this should be part of a new MLFiletransferInfo model class.

Monal/Classes/ChatView.swift
```diff
@@ -747,6 +751,32 @@ class ChatViewMessage: ExyteChat.Message {
     private var subscriptions: Set<AnyCancellable> = Set()
     override var text: String {
         get {
+            if innerMessage.messageType == kMessageTypeFiletransfer {
+                let downloadState = innerMessage.downloadState as MLMessageDownloadState.RawValue
+                switch(downloadState) {
+                    case MLMessageDownloadState.complete.rawValue:
+                        let mimeType = innerMessage.filetransferMimeType as String
+                        if mimeType.starts(with: "audio") || mimeType.starts(with: "image") || mimeType.starts(with: "video") {
```
tmolitor-stud-tu: should be "audio/", "image/" and "video/" rather than "audio", "image" and "video"

Monal/Classes/ChatView.swift
```diff
@@ -799,6 +829,55 @@ class ChatViewMessage: ExyteChat.Message {
...
+                return nil
+            }
+            let mimeType = innerMessage.filetransferMimeType as String
+            guard mimeType.starts(with: "audio/") else {
+                return nil
+            }
+            let info = MLFiletransfer.getFileInfo(for: innerMessage.obj)
+            let cacheFile = info["cacheFile"] as! String
+            let url = URL(fileURLWithPath: cacheFile)
+            let duration = innerMessage.mediaDuration as Double
```
tmolitor-stud-tu: I don't think caching the duration in the database is necessary. is calculating it on-the-fly that slow?

tmolitor-stud-tu (changes requested):
I've added some comments

Monal/Classes/ChatView.swift
lissine0: There are two ways to compute the duration:
- Using AVPlayer, like I did in the migration (https://github.com/lissine0/Monal/blob/c67dab2617cb/Monal/Classes/DataLayerMigrations.m#L1005)
- Using AVURLAsset.duration, like I do after downloading an attachment (https://github.com/lissine0/Monal/blob/c67dab2617c/Monal/Classes/HelperTools.m#L1155)

The AVURLAsset way can be used only asynchronously, otherwise it returns zero.

About the AVPlayer way, in previous draft of this PR I had the following code in ChatView.swift:
```swift
override var recording: Recording? {
    get {
        /* ... */
        let url = URL(fileURLWithPath: cacheFile)
        do {
            let player = try AVAudioPlayer(contentsOf: url)
            return Recording(duration: player.duration, url: url, mimeType: mimeType)
        } catch {
            DDLogWarn("Couldn't compute the duration of the audio file \(cacheFile)")
        }
```
This works but has one caveat: the very first time the `recording` property is computed (all audio files considered), it takes ~70ms to execute this line:
```swift
let player = try AVAudioPlayer(contentsOf: url)
```
In subsequent computations of the property (even for other files), it takes less than 1ms
(The AVAudioPlayer is likely cached somehow)
This means that if an audio file is in the ChatView, opening that chat after launching the app will have a ~70ms lag (because the line is executed synchronously on the main thread).

If the `MLFiletransferInfo` model class you suggested will be persisted in the db, the duration should be included there.

lissine0: The latest version I pushed does it on the fly, and it's not slow at all :)
It uses `AVURLAsset.duration` in a background thread when the getter is called
The getter launches the work in the background, and returns 0.0 / nil
Later when the work finishes, it calls the setter to trigger KVO notification and a ChatView update

Monal/Classes/DataLayer.m
```diff
@@ -1478,6 +1467,39 @@ -(void) setMessageId:(NSString*) messageid stanzaId:(NSString*) stanzaId
     }];
 }
 
+-(NSDictionary*) getFiletransferInfoForHistoryId:(NSNumber*) historyId
+{
+    if(historyId == nil)
+        return (NSDictionary*) nil;
+
+    return [self.db idReadTransaction:^{
+        NSArray* results = [self.db executeReader:@"SELECT * FROM filetransfer_info WHERE message_history_id=?;" andArguments:@[historyId]];
```
tmolitor-stud-tu, line 1476: This could be improved to use the scalar reader rather thanthe array reader and then returning only the first entry (the scalar reader returns nil if no db row matches)

Monal/Classes/MLContactCell.m
```diff
@@ -69,13 +70,13 @@ -(void) displayLastMessage:(MLMessage* _Nullable) lastMessage forContact:(MLCont
             [self showStatusText:NSLocalizedString(@"🔗 A Link", @"") inboundDir:lastMessage.inbound fromUser:senderOfLastGroupMsg];
         else if([lastMessage.messageType isEqualToString:kMessageTypeFiletransfer])
         {
-            if([lastMessage.filetransferMimeType hasPrefix:@"image/"])
+            if([lastMessage.fileInfo.mimeType hasPrefix:@"image/"])
```
tmolitor-stud-tu, line 73: please add new computed properties named `isVideo` and `isImage` etc. to the new fileInfo class and then replace all mimeType tests for `image/` etc. with these new properties everywhere in our codebase (internally still doing the string comparisons of course).

Monal/Classes/MLFiletransfer.m
```diff
@@ -282,14 +284,16 @@ +(void) downloadFileForHistoryID:(NSNumber*) historyId andForceDownload:(BOOL) f
             NSNumber* filetransferSize = @([[_fileManager attributesOfItemAtPath:cacheFile error:nil] fileSize]);
             DDLogDebug(@"Updating db and sending out kMonalMessageFiletransferUpdateNotice");
             //update db with content type and size
-            [[DataLayer sharedInstance] setMessageHistoryId:historyId filetransferMimeType:mimeType filetransferSize:filetransferSize];
+            [[DataLayer sharedInstance] setFiletransferInfoForHistoryId:historyId withMimeType:mimeType andSize:filetransferSize];
+
             //send out update notification
             xmpp* account = [[MLXMPPManager sharedInstance] getEnabledAccountForID:msg.accountID];
             if(account != nil)      //don't send out update notices for already deleted accounts
                 [[MLNotificationQueue currentQueue] postNotificationName:kMonalMessageFiletransferUpdateNotice object:account userInfo:@{
```
tmolitor-stud-tu, line 292: Not sure if we still need all these values in the userInfo dictionary.

Monal/Classes/MLFiletransferInfo.h
```diff
@@ -0,0 +1,39 @@
...
+//  Created by lissine on 16/12/2025.
+//  Copyright © 2025 monal-im.org. All rights reserved.
+//
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class MLMessage;
+
+typedef NS_ENUM(NSUInteger, DownloadState) {
+    DownloadStateUndefined, // equivalent of nil
```
tmolitor-stud-tu, line 14: Why the distinction between `Undefined` and `None`?

Monal/Classes/MLFiletransferInfo.m
```diff
@@ -0,0 +1,176 @@
...
+    else
+        state = DownloadStateNone;
+
+    self.downloadState = state;
+    return _downloadState;
+}
+
+-(NSString*) url
+{
+    return self.message.messageText;
```
tmolitor-stud-tu, line 193: You'll need a `keyPathsForValuesAffectingUrl` method...and similar methods for all other computed properties.

Monal/Classes/MLFiletransferInfo.m
```diff
@@ -0,0 +1,176 @@
...
+}
+
+-(MLMessage*) message
+{
+    return [MLMessage createMessageFromHistoryID:self.historyId];
+}
+
+-(NSString*) id
+{
+    return [NSString stringWithFormat:@"%@", self.historyId];
```
tmolitor-stud-tu, line 247: Please add a comment: `//Needed to not create a retain cycle between MLFiletransferInfo and MLMessage`

Monal/Classes/MLMessage.m
```diff
@@ -293,6 +273,11 @@ -(NSString*) contactDisplayName
         return [MLContact createContactFromJid:self.buddyName andAccountID:self.accountID].contactDisplayName;
 }
 
+-(MLFiletransferInfo*) fileInfo
+{
+    return [MLFiletransferInfo createFiletransferInfoFromMessage:self];
```
tmolitor-stud-tu: I'd call it "...forMessage" rather than "...fromMessage"

Monal/Classes/MLNotificationManager.m
```diff
@@ -504,64 +505,55 @@ -(void) showModernNotificationForMessage:(MLMessage*) message withSound:(BOOL) s
...
-                        UNNotificationAttachment* attachment;
-                        UTType* typeHint = [UTType typeWithMIMEType:mimeType];
-                        if(typeHint == nil)
-                            typeHint = UTTypeMovie;
-                        attachment = [self createNotificationAttachmentForFileInfo:info havingTypeHint:typeHint];
-                        if(attachment)
-                            content.attachments = @[attachment];
-                    }
+                    UNNotificationAttachment* attachment;
+                    UTType* typeHint = [UTType typeWithMIMEType:mimeType];
```
tmolitor-stud-tu: Please add a new `UTType*` computed property to the new file info class. That might become handy (not just in this case).

tmolitor-stud-tu (changes requested):
That's only the review for this commit, reviewing the remaining ones now. Most of it is only small changes, and a few questions and suggestions you don't need to implement.

Monal/Classes/MLFiletransferInfo.m
```diff
@@ -122,6 +125,29 @@ -(DownloadState) downloadState
...
 }
 
+-(double) mediaDuration {
+    // return already cached value.
+    if(_mediaDuration != 0.0)
+        return _mediaDuration;
+
+    // Compute the duration, but ensure that only one thread is doing so
+    // Also prevent further retries in the case of an error
+    if(!self.isComputingMediaDuration) {
```
tmolitor-stud-tu, line 141: Please use proper locking through `NSLock`. Even though it isn't harmful if we accidentally compute this more than once, the BOOL handling makes the impression of preventing this and this might be unexpected in the future.

tmolitor-stud-tu (changes requested):
one more

Monal/Classes/MLFiletransferInfo.m
```diff
@@ -148,6 +151,39 @@ -(double) mediaDuration {
...
+    NSURL* url = [[MLImageManager sharedInstance] getThumbnailURLOfMessage:self.message];
+    if(url != nil)
+    {
+        self.thumbnailURL = url;
+        return url;
+    }
+
+    // Generate the thumbnail now, but ensure that only one thread is doing the generation
+    // Also, prevent retries in the case of an error (to avoid repeated failures)
+    if(!self.isGeneratingThumbnail)
```
tmolitor-stud-tu, line 173: same here

tmolitor-stud-tu (changes requested):

Monal/Classes/MLFiletransferInfo.h
lissine0: `Undefined` is a value that represents `nil` (because an `NSUInteger` can't be nil)
This way, I could distinguish between the case of the variable being uninitialized (.undefined) and between the case of the variable being initialized but not having done the size / mimeType check yet (.none)

Note that the getter never returns .undefined

Monal/Classes/MLFiletransfer.m
lissine0: Yes we need them.
- The `message` allows us to distinguish which message is concerned with this notification.
- The `mimeType` and `size` update the properties of `MLFiletransferInfo`.
In fact, that is the only way these properties get updated, aside from the message retraction notification (kMonalDeletedMessageNotice)
- The `downloadState` modifies the `MLFiletransferInfo` property of the same name, triggering a ChatView update.

tmolitor-stud-tu: We could just pass the historyID in the notification and reload everything else from the db. feels a bit cleaner (only one source of truth: the db).

lissine0: Or, we could write to the db when handling the notification.
To explain more: the `kMonalMessageFiletransferUpdateNotice` is sent only from two places, both of them in `MLFiletransfer.m`.
And in each of these cases, it writes to the db before sending the notification.
We could move the writing of the db from `MLFiletransfer.m` (before sending the notification) to `MLFiletransferInfo.m` (while processing the notification)

tmolitor-stud-tu: Well, it feels better to leave the db writing in MLFiletransfer and just read it from db in the info class. notifications can be delayed by the notification queue (which is okay for the ui, but not for writing to the db)

Monal/Classes/DataLayer.m
lissine0: We have to use `executeReader` here because we're reading from multiple columns (`SELECT *`)
If we use `executeScalarReader`, we'd only get results from the first column.

Monal/Classes/MLFiletransferInfo.m
lissine0: I did not add a similar method for `downloadState`, `mediaDuration` and `thumbnailURL` because these computed properties have an ivar cache (which is needed for KVO in order to trigger ChatView re-renders, since these properties can change even if the properties they depend don't change)

Also, I did not add a similar method for `url` and for the properties that depend solely on it, because `url` isn't supposed to ever change.
Even if LMC is used, it only changes the message body, not the `oob`. In fact Monal doesn't let you do LMC on filetransfer messages.

Monal/Classes/DataLayer.m
tmolitor-stud-tu: Oh, good catch!

PR #1426, opened 2025-05-07
New ChatView: Support incoming updates of messages

lissine0:
Use `kMonalUpdatedMessageNotice` for message corrections and MUC message reflections, instead of using `kMonalNewMessageNotice`.

I added observing of this new notification in all the places that are observing `kMonalNewMessageNotice`, except for `MonalAppDelegate`.
I think that's not needed because message updates don't cause the unread counter to be changed.

Also, I'm not sure observing the new notification in `MLContact` is strictly necessary, but it's not harmful.

On a semi-related note, you once said that when posting notifications, `MLNotificationQueue` should always be used (as opposed to `NSNotificationCenter`)
It turns out there are some places in the code base that use `NSNotificationCenter` for posting notifications.
See them with `grep --color=auto -in 'NotificationCenter.*post' Monal/Classes/*.{m,swift}`
Should I use `MLNotificationQueue` in these places?
(The `CFNotificationCenter` ones should remain)

### DIFF
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -200,7 +200,9 @@ struct ChatView: View {
             guard let newMLMessage = MLXMPPManager.sharedInstance().sendMessageAndAddToHistory(message: draft.text, havingType: kMessageTypeText, toContact: self.contact.obj, isEncrypted: self.contact.isEncrypted, uploadInfo: nil) else {
                 return
             }
-            messages.append(ChatViewMessage(newMLMessage))
+            messages.append(ChatViewMessage(ObservableKVOWrapper(newMLMessage)))
+        } messageBuilder: { message, viewModel, positionInUserGroup, positionInMessagesSection, positionInCommentsGroup, showContextMenuClosure, messageActionClosure, showAttachmentClosure in
+            MessageView(message: (message as! ChatViewMessage).message, viewModel: viewModel, positionInUserGroup: positionInUserGroup, positionInMessagesSection: positionInMessagesSection)
         }
         .showNetworkConnectionProblem(false)
 //         .enableLoadMore(pageSize: 3) { message in
@@ -350,7 +352,7 @@ struct ChatView: View {
             if messages.isEmpty {
                 let dbMessages = DataLayer.sharedInstance().messages(forContact:contact.contactJid, forAccount:contact.accountID) as! [MLMessage]
                 for msg in dbMessages {
-                    messages.append(ChatViewMessage(msg))
+                    messages.append(ChatViewMessage(ObservableKVOWrapper(msg)))
                 }
             }
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
@@ -408,7 +410,7 @@ struct ChatView: View {
             if message.isEqual(to: self.contact.obj) {
                 // Do not insert based on delay timestamp because that
                 // would make it possible to fake history entries.
-                messages.append(ChatViewMessage(message))
+                messages.append(ChatViewMessage(ObservableKVOWrapper(message)))
             }
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
         }
@@ -419,9 +421,9 @@ struct ChatView: View {
 }
 
 class ChatViewMessage: ExyteChat.Message {
-    @Published public var message: MLMessage
-    
-    init(_ message: MLMessage) {
+    let message: ObservableKVOWrapper<MLMessage>
+
+    init(_ message: ObservableKVOWrapper<MLMessage>) {
         self.message = message
         let user = ExyteChat.User(id: message.senderID, name: message.contactDisplayName, avatarURL: nil, isCurrentUser: !message.inbound)
         super.init(id: message.id, user: user, createdAt: message.timestamp, text: message.messageText)
@@ -481,3 +483,32 @@ public extension ExyteChat.MessageView {
 //         }
     }
 }*/
+struct MessageView: View {
+    @StateObject var message: ObservableKVOWrapper<MLMessage>
+    @ObservedObject var viewModel: ExyteChat.ChatViewModel
+    let positionInUserGroup: PositionInUserGroup
+    let positionInMessagesSection: PositionInMessagesSection
+    init(message: ObservableKVOWrapper<MLMessage>, viewModel: ChatViewModel, positionInUserGroup: PositionInUserGroup, positionInMessagesSection: PositionInMessagesSection) {
+        _message = StateObject(wrappedValue: message)
+        self.viewModel = viewModel
+        self.positionInUserGroup = positionInUserGroup
+        self.positionInMessagesSection = positionInMessagesSection
+    }
+    var body: some View {
+        ExyteChat.MessageView(
+            viewModel: viewModel,
+            message: ChatViewMessage(message),
+            positionInUserGroup: positionInUserGroup,
+            positionInMessagesSection: positionInMessagesSection,
+            chatType: .conversation,
+            avatarSize: 32,
+            tapAvatarClosure: nil,
+            messageStyler: AttributedString.init,
+            shouldShowLinkPreview: { _ in true },
+            isDisplayingMessageMenu: false,
+            showMessageTimeView: true,
+            messageLinkPreviewLimit: 8,
+            font: UIFontMetrics.default.scaledFont(for: UIFont.systemFont(ofSize: 15))
+        )
+    }
+}
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -386,7 +386,7 @@ struct ChatView: View {
                 MLNotificationManager.sharedInstance().currentContact = nil
             }
         }
-        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("kMonalOmemoFetchingStateUpdate")).receive(on: RunLoop.main)) { notification in
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kMonalOmemoFetchingStateUpdate)).receive(on: RunLoop.main)) { notification in
             if let xmppAccount = notification.object as? xmpp, let notificationJid = notification.userInfo?["jid"] as? String {
                 if xmppAccount.accountID == contact.accountID && notificationJid == contact.contactJid {
                     DDLogDebug("Got omemo fetching update: \(contact) --> \(String(describing:notification.userInfo))")
@@ -398,7 +398,7 @@ struct ChatView: View {
                 }
             }
         }
-        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("kMonalNewMessageNotice")).receive(on: RunLoop.main)) { notification in
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kMonalNewMessageNotice)).receive(on: RunLoop.main)) { notification in
             DDLogVerbose("chat view got new message notice \(String(describing:notification.userInfo))")
 
             guard let message = notification.userInfo?["message"] as? MLMessage else {
@@ -412,7 +412,7 @@ struct ChatView: View {
             }
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
         }
-        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("kMonalRefresh")).receive(on: RunLoop.main)) { notification in
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kMonalRefresh)).receive(on: RunLoop.main)) { notification in
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
         }
     }
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -425,8 +425,9 @@ class ChatViewMessage: ExyteChat.Message {
 
     init(_ message: ObservableKVOWrapper<MLMessage>) {
         self.message = message
+        let messageText = message.retracted ? NSLocalizedString("This message got retracted", comment: "") : message.messageText
         let user = ExyteChat.User(id: message.senderID, name: message.contactDisplayName, avatarURL: nil, isCurrentUser: !message.inbound)
-        super.init(id: message.id, user: user, createdAt: message.timestamp, text: message.messageText)
+        super.init(id: message.id, user: user, createdAt: message.timestamp, text: messageText)
     }
 }
 
--- a/Monal/Classes/MLMessage.m
+++ b/Monal/Classes/MLMessage.m
@@ -9,52 +9,96 @@
 #import <monalxmpp/MLMessage.h>
 #import <monalxmpp/MLContact.h>
 #import <monalxmpp/MLConstants.h>
+#import <monalxmpp/DataLayer.h>
 #import <monalxmpp/xmpp.h>
+
+static NSMutableDictionary* _singletonCache;
 
 @implementation MLMessage
 {
     MLContact* _contact;
 }
 
++(void) initialize
+{
+    _singletonCache = [NSMutableDictionary new];
+}
+
++(MLMessage*) createMessageFromHistoryID:(NSNumber*) historyID
+{
+    if(historyID == nil)
+        return nil;
+    NSArray<MLMessage*>* result = [[DataLayer sharedInstance] messagesForHistoryIDs:@[historyID]];
+    if(![result count])
+        return nil;
+    return result[0];
+}
+
 +(MLMessage*) messageFromDictionary:(NSDictionary*) dic
 {
-    MLMessage* message = [MLMessage new];
-    message.accountID = [dic objectForKey:@"account_id"];
-    
-    message.buddyName = [dic objectForKey:@"buddy_name"];
-    message.inbound = [(NSNumber*)[dic objectForKey:@"inbound"] boolValue];
-    message.actualFrom = [dic objectForKey:@"af"];
-    message.messageText = [dic objectForKey:@"message"];
-    message.isMuc = [(NSNumber*)[dic objectForKey:@"Muc"] boolValue];
-    
-    message.messageId = [dic objectForKey:@"messageid"];
-    message.stanzaId = [dic objectForKey:@"stanzaid"];
-    message.messageDBId = [dic objectForKey:@"message_history_id"];
-    message.timestamp = [dic objectForKey:@"thetime"];
-    message.messageType = [dic objectForKey:@"messageType"];
-    message.mucType = [dic objectForKey:@"muc_type"];
-    message.participantJid = [dic objectForKey:@"participant_jid"];
-    
-    message.hasBeenDisplayed = [(NSNumber*)[dic objectForKey:@"displayed"] boolValue];
-    message.hasBeenReceived = [(NSNumber*)[dic objectForKey:@"received"] boolValue];
-    message.hasBeenSent = [(NSNumber*)[dic objectForKey:@"sent"] boolValue];
-    message.encrypted = [(NSNumber*)[dic objectForKey:@"encrypted"] boolValue];
-    
-    message.unread = [(NSNumber*)[dic objectForKey:@"unread"] boolValue];
-    message.displayMarkerWanted = [(NSNumber*)[dic objectForKey:@"displayMarkerWanted"] boolValue];
-    
-    message.previewText = [dic objectForKey:@"previewText"];
-    message.previewImage = [NSURL URLWithString:[dic objectForKey:@"previewImage"]];
-    
-    message.errorType = [dic objectForKey:@"errorType"];
-    message.errorReason = [dic objectForKey:@"errorReason"];
-    
-    message.filetransferMimeType = [dic objectForKey:@"filetransferMimeType"];
-    message.filetransferSize = [dic objectForKey:@"filetransferSize"];
-    
-    message.retracted = [(NSNumber*)[dic objectForKey:@"retracted"] boolValue];
-    
-    return message;
+    // Draft messages don't have a historyID and shouldn't be cached
+    if ([[dic objectForKey:@"messageType"] isEqualToString:kMessageTypeMessageDraft])
+    {
+        MLMessage* message = [MLMessage new];
+        // Fill only the properties useful for a draft message
+        // The timestamp is used when rendering MLContactCell, among other things
+        message.messageText = [dic objectForKey:@"message"];
+        message.messageType = [dic objectForKey:@"messageType"];
+        message.timestamp = [dic objectForKey:@"thetime"];
+        return message;
+    }
+
+    NSNumber* cacheKey = [dic objectForKey:@"message_history_id"];
+    MLAssert(cacheKey != nil, @"A non-draft message can't have a nil historyID!");
+    @synchronized(_singletonCache) {
+        if(_singletonCache[cacheKey] != nil)
+        {
+            MLMessage* obj = ((WeakContainer*)_singletonCache[cacheKey]).obj;
+            if(obj != nil)
+                return obj;
+            else
+                [_singletonCache removeObjectForKey:cacheKey];
+        }
+
+        MLMessage* message = [MLMessage new];
+        message.accountID = [dic objectForKey:@"account_id"];
+
+        message.buddyName = [dic objectForKey:@"buddy_name"];
+        message.inbound = [(NSNumber*)[dic objectForKey:@"inbound"] boolValue];
+        message.actualFrom = [dic objectForKey:@"af"];
+        message.messageText = [dic objectForKey:@"message"];
+        message.isMuc = [(NSNumber*)[dic objectForKey:@"Muc"] boolValue];
+
+        message.messageId = [dic objectForKey:@"messageid"];
+        message.stanzaId = [dic objectForKey:@"stanzaid"];
+        message.messageDBId = [dic objectForKey:@"message_history_id"];
+        message.timestamp = [dic objectForKey:@"thetime"];
+        message.messageType = [dic objectForKey:@"messageType"];
+        message.mucType = [dic objectForKey:@"muc_type"];
+        message.participantJid = [dic objectForKey:@"participant_jid"];
+
+        message.hasBeenDisplayed = [(NSNumber*)[dic objectForKey:@"displayed"] boolValue];
+        message.hasBeenReceived = [(NSNumber*)[dic objectForKey:@"received"] boolValue];
+        message.hasBeenSent = [(NSNumber*)[dic objectForKey:@"sent"] boolValue];
+        message.encrypted = [(NSNumber*)[dic objectForKey:@"encrypted"] boolValue];
+
+        message.unread = [(NSNumber*)[dic objectForKey:@"unread"] boolValue];
+        message.displayMarkerWanted = [(NSNumber*)[dic objectForKey:@"displayMarkerWanted"] boolValue];
+
+        message.previewText = [dic objectForKey:@"previewText"];
+        message.previewImage = [NSURL URLWithString:[dic objectForKey:@"previewImage"]];
+
+        message.errorType = [dic objectForKey:@"errorType"];
+        message.errorReason = [dic objectForKey:@"errorReason"];
+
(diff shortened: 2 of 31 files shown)


tmolitor-stud-tu:
Most of these notifications have a comment, explaining why `NSNotificationCenter` is used directly, someting linke this one: `//don't queue this notification because it should be handled immediately`.

`kMonalFrozen` and `kMonalUnfrozen` as well as `kMonalIncomingIPC` should be handled inline, too, but don't have such a comment.

Only `Classes/PasswordMigration.swift` should use `MLNotificationQueue` instead of `NSNotificationCenter`.

Background: We don't want incoming stanzas to have side effects through notifications while we are still inside the db transaction wrapping this stanza processing. So We use `MLContificationQueue` instead of `NSNotificationCenter` everywhere in the code. That class will forward the call directly to `NSNotificationCenter` unless a queue was created by its "context manager" (context manager in the python sense, using objc blocks). Then the notification is queued until the context manager exits. Context managers can be nested, which lets the queued notifications bubble up to the next context manager, if the inner one exits. Just to mention the obvious: the context manager only captures notifications created inside the thread the context manager runs in.

Incoming stanza processing is one place, we use that context manager to queue notifications. Another one is processing history fetches from MAM. We use it to suppress all notifications for "new" messages in this case and clean the queue instead of letting the notifications "bubble up".

There are only a few places (those you found) where we don't want to queue notifications and always handle them immediately.

tmolitor-stud-tu:
> I added observing of this new notification in all the places that are observing kMonalNewMessageNotice, except for MonalAppDelegate.
> I think that's not needed because message updates don't cause the unread counter to be changed.

Yes, that makes sense :)

> Also, I'm not sure observing the new notification in MLContact is strictly necessary, but it's not harmful.

No, that's not needed, you can remove that one :)

Monal/Classes/ChatView.swift
```diff
@@ -194,6 +194,16 @@ struct ChatView: View {
     }
 #endif
 }
+    func findAndUpdateMessage(with message: MLMessage) {
+        if message.isEqual(to: self.contact.obj) {
+            for index in messages.indices {
+                if messages[index].message.messageDBId == message.messageDBId {
```
tmolitor-stud-tu: I think, the proper implementation is to extend `MLMessage` to be a full fledged model class handling updates itself (and being a per-db-id singleton), just like `MLContact`.

Then use the KVO wrapper to automatically update the UI.

tmolitor-stud-tu (changes requested):

Monal/Classes/ChatView.swift
lissine0: The problem is, if I just update the MLMessage object, for example:

```swift
if messages[index].message.messageDBId == message.messageDBId {
    messages[index].message.update(with: message)
```
Then SwiftUI doesn't detect that the `messages` `@State` variable changed, and doesn't re-render the view.
By inserting a fresh new entry in the `messages` array, SwiftUI figures this out correctly and updates the view. (the live update requires a SwiftPM dependencies bump to work correctly. But I did that in my other PR)

lissine0:
> > Also, I'm not sure observing the new notification in MLContact is strictly necessary, but it's not harmful.
> 
> No, that's not needed, you can remove that one :)

That was my intuition, but for some reason `MLContact` observes for `kMonalDeletedMessageNotice` even though message retraction don't cause the unread counter to be updated.
Edit: See https://github.com/monal-im/Monal/commit/d4c21167a8

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Do I understand correctly, that you can update this PR to use the `ObservableKVOWrapper` and let `MLMessage` be a proper  model class, once I merge your other PR?

tmolitor-stud-tu:
> > > Also, I'm not sure observing the new notification in MLContact is strictly necessary, but it's not harmful.
> > 
> > 
> > No, that's not needed, you can remove that one :)
> 
> That was my intuition, but for some reason `MLContact` observes for `kMonalDeletedMessageNotice` even though message retraction don't cause the unread counter to be updated. Edit: See [d4c21167a8](https://github.com/monal-im/Monal/commit/d4c21167a8)

Well, a retraction should update the unread counter. I think we should set the `unread` flag to `1` in `DataLayer.m` method `retractMessageHistory:`. Could you add that to this PR and remove the `kMonalUpdatedMessageNotice` handling from `MLContact.m`?

Monal/Classes/ChatView.swift
lissine0: I think the following can work:
In the [`ChatViewMessage`](https://github.com/monal-im/Monal/blob/ce68e69686e463b835f05ea45f1a6d02d43bc08f/Monal/Classes/ChatView.swift#L421) class, we make the `MLMessage` property an `ObservedObject` (using `KVOObserver` for example).
Then in that same class, we listen for changes to the MLMessage property, and when it does change, we update the properties of the class (the ones that are currently only set in the initializer e.g. the `text` property)

tmolitor-stud-tu:
> > > > Also, I'm not sure observing the new notification in MLContact is strictly necessary, but it's not harmful.
> > > 
> > > 
> > > No, that's not needed, you can remove that one :)
> > 
> > 
> > That was my intuition, but for some reason `MLContact` observes for `kMonalDeletedMessageNotice` even though message retraction don't cause the unread counter to be updated. Edit: See [d4c21167a8](https://github.com/monal-im/Monal/commit/d4c21167a8)
> 
> Well, a retraction should update the unread counter. I think we should set the `unread` flag to `1` in `DataLayer.m` method `retractMessageHistory:`. Could you add that to this PR and remove the `kMonalUpdatedMessageNotice` handling from `MLContact.m`?

Sorry, my fault, of course I meant `unread=0`, not `unread=1` --> mark the retracted message as read because we also remove the push notification for a retracted message

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Why not just override the properties of the base class in our `ChatViewMessage` class using computed properties? The could compute the right values using the `MLMessage` and even properly propagate property changes of the `MLMessage` properties up the chain (using the `ObservableKVOWrapper` or something similar).

@matthewrfennell what do you think? imho this should be possible, so that changing the `messageText` property inside an `MLMessage` should automatically bubble up the chain all the way to swiftui consuming that message text.

matthewrfennell: This makes sense to me, I haven't tried it, but I think it should work

tmolitor-stud-tu: Yes, that should work, but it's not the question/solution I proposed. I want the `MLMessage` be a singleton over the db id, just like `MLContact` is a singleton over the `(jid, accountid)` tuple. Then The MLMessage should listen for all update notifications and change the values of its properties according to these updates.

The rest of our code in `ChatView.swift` (etc.?) should make sure that these property updates of the underlying `MLMessage` are automatically picked up by `SwiftUI` using our `ObservableKVOWrapper` etc. (just like in the contact details + `MLContact` case).

@matthewrfennell My question is now: could you (together with lissine) figure out, what changes/additions etc. need to be done to the `ChatViewMessage` class (and of course `MLMessage`) to realize my vision depicted above.

tmolitor-stud-tu: Additional note: to test, you could use a timer (`createTimer(1, (^{ ... }));`) in `MLMessage` to automatically add a dot to the end of every message every second. That should automatically trigger an UI update, if everything works out as intended :)

lissine0:
> Well, a retraction should update the unread counter. I think we should set the `unread` flag to `1` in `DataLayer.m` method `retractMessageHistory:`. Could you add that to this PR and remove the `kMonalUpdatedMessageNotice` handling from `MLContact.m`?

Note that doing this will change the position of the "Unread Messages Below" indicator to under the most recent retracted message.
i.e. if someone sends multiple messages and then retracts one of them, the unread indicator will move down even if I didn't open the chat.

tmolitor-stud-tu:
> Note that doing this will change the position of the "Unread Messages Below" indicator to under the most recent retracted message. i.e. if someone sends multiple messages and then retracts one of them, the unread indicator will move down even if I didn't open the chat.

Oh, that's an unintended side effect I didn't think of. That's unfortunate.
I think we should introduce a new "retracted" boolean db field and set this to 1 if a message gets retracted. Then, when counting the unread messages, we should not count retracted ones.

Other than that: is this PR ready to be merged?

lissine0:
> I think we should introduce a new "retracted" boolean db field and set this to 1 if a message gets retracted.

We already have a `retracted` field in the `message_history` table, as well as a `retracted` field in `MLMessage`

> Other than that: is this PR ready to be merged?

It's only missing the handling of Status message update notifications (message received, displayed, error etc.) in MLMessage
I didn't include it yet because I want to try to refactor all of the Status update notifications into one notification.
Other than that, the PR is ready

lissine0:
Note that the changes in ChatView.swift depend on changes in the ExyteChat fork: the `MessageView` and `ChatViewModel` had to become public

tmolitor-stud-tu:
> Note that the changes in ChatView.swift depend on changes in the ExyteChat fork: the `MessageView` and `ChatViewModel` had to become public

@matthewrfennell will we be able to get this change upstream?

lissine0:
> > Note that the changes in ChatView.swift depend on changes in the ExyteChat fork: the `MessageView` and `ChatViewModel` had to become public
> 
> @matthewrfennell will we be able to get this change upstream?

I don't think so, as I made internal structures public.
But it's a small change, so it won't be hard to keep it when rebasing: https://github.com/monal-im/ExyteChat/commit/a4fa92f49397c0a5b0c798fddf123afe7f47bc4e
https://github.com/monal-im/ExyteChat/commit/864c1f6181641aa34a0f9ffd820e9e95a94fd496

tmolitor-stud-tu:
> We already have a `retracted` field in the `message_history` table, as well as a `retracted` field in `MLMessage`

Oh great! Then only the calculation of the unread counter has to be updated to take that field into account (or does it already?).
Setting a retracted message to unread=0 should not be necessary anymore, you can remove that :)
 
> It's only missing the handling of Status message update notifications (message received, displayed, error etc.) in MLMessage I didn't include it yet because I want to try to refactor all of the Status update notifications into one notification. Other than that, the PR is ready

Okay, would you like to create a new PR for these status notifications and me to do a full review and merge this PR now? Or do you want me to review and merge this only after the status handling is complete?

lissine0:
> > We already have a `retracted` field in the `message_history` table, as well as a `retracted` field in `MLMessage`
> 
> Oh great! Then only the calculation of the unread counter has to be updated to take that field into account (or does it already?). Setting a retracted message to unread=0 should not be necessary anymore, you can remove that :)

Unfortunately, the situation is a bit complicated.
The algorithm for the unread counter is as follows: we go through messages from the most recent to older, until we find a message that is read. And we increment a counter each time.
Then the "Unread messages below" is inserted based on the counter.
Now, if we ignore retracted messages (i.e. increment the counter regardless), then it causes a problem when the most recent message is retracted. In that situation, the unread counter will always be above it.
> > It's only missing the handling of Status message update notifications (message received, displayed, error etc.) in MLMessage I didn't include it yet because I want to try to refactor all of the Status update notifications into one notification. Other than that, the PR is ready
> 
> Okay, would you like to create a new PR for these status notifications and me to do a full review and merge this PR now? Or do you want me to review and merge this only after the status handling is complete?

I decided against the refactor. So, implementing the missing handling of those notifications shouldn't take a lot time. Thus I'd like to include it in this PR

tmolitor-stud-tu:
Hmm we'll have to split the counting for the unread badge from the counting for the unread below marker then. Two datalayer methods instead of one.

Monal/Classes/ChatView.swift
```diff
@@ -350,7 +353,7 @@ struct ChatView: View {
             if messages.isEmpty {
                 let dbMessages = DataLayer.sharedInstance().messages(forContact:contact.contactJid, forAccount:contact.accountID) as! [MLMessage]
                 for msg in dbMessages {
-                    messages.append(ChatViewMessage(msg))
+                    messages.append(ChatViewMessage(ObservableKVOWrapper(msg)))
```
tmolitor-stud-tu, line 355: Maybe don't wrap it into `ObservableKVOWrapper` outside, but wrap it inside the constructor of `ChatViewMessage`. That seems a bit cleaner because it doesn't expose these implementation details to the caller.

Monal/Classes/ChatView.swift
```diff
@@ -481,3 +485,32 @@ public extension ExyteChat.MessageView {
 //         }
     }
 }*/
+struct MessageView: View {
+    @StateObject var message: ObservableKVOWrapper<MLMessage>
+    @ObservedObject var viewModel: ExyteChat.ChatViewModel
+    let positionInUserGroup: PositionInUserGroup
+    let positionInMessagesSection: PositionInMessagesSection
+    init(message: ObservableKVOWrapper<MLMessage>, viewModel: ChatViewModel, positionInUserGroup: PositionInUserGroup, positionInMessagesSection: PositionInMessagesSection) {
```
tmolitor-stud-tu, line 492: Why don't you pass in the `ChatViewMessage` here? You can get the underlying `MLMessage` from it just fine and you won't have to instantiate a new `ChatViewMessage` in the view body every time this gets rendered.

Monal/Classes/MLMessage.m
```diff
@@ -9,13 +9,43 @@
...
 {
     MLContact* _contact;
 }
 
++(void) initialize
+{
+    _singletonCache = [NSMutableDictionary new];
+}
+
++(MLMessage*) createMessageFromHistoryID:(NSNumber*) historyID
```
tmolitor-stud-tu, line 28: So we have `MLMessage messageFromDictionary:` which is used in various places by the `DataLayer` and this method here. That's a duplication I don't like. Especially because one of them has a cache and the other one doesn't. In `MLContact`for example, the `MLContact contactFromDictionary:` method is only called internally in `MLContact` and isn't exposed publicly.

I know you want to preserve all these high level `DataLayer` methods returning ready made `MLMessage` objects and that's a good decision. But then the `MLMessage createMessageFromHistoryID:` method should be removed and the cache added to `MLMessage messageFromDictionary:` . Every code currently calling into `createMessageFromHistoryID` should use the appropriate `DataLayer` method which is `messageForHistoryID:`.

**Even better:** move that `DataLayer messageForHistoryID:` method to `MLMessage` and name it `createMessageFromHistoryID:` (that's possible because this `DataLayer` method doesn't use any database calls but merely forwards the call to `DataLayer messagesForHistoryIDs:`), then change every call to that `DataLayer`  method to use the new `MLMessage` method instead.
**Note:** caching should be still moved to `MLMessage messageFromDictionary:`.

**So to sum it up:**

1. Move the cache from  to `MLMessage messageFromDictionary:`
2. Replace the code of `MLMessage createMessageFromHistoryID:` with `DataLayer messageForHistoryID:`
3. Remove `DataLayer messageForHistoryID:` and change all calls to use `MLMessage createMessageFromHistoryID:` now

Monal/Classes/MLNotificationManager.m
```diff
@@ -278,6 +279,14 @@ -(void) handleFiletransferUpdate:(NSNotification*) notification
 }
 
 -(void) handleNewMessage:(NSNotification*) notification
+{
```
tmolitor-stud-tu: Can' t  `handleNewMessage:` and `handleUpdatedMessage:`be merged? They share the exact same code. Use the same method name when adding the observer.

Monal/Classes/chatViewController.m
```diff
@@ -1981,13 +1945,12 @@ -(NSString*) formattedTimeStampWithSource:(NSDate *) sourceDate
 -(void) retry:(id) sender
 {
     NSInteger msgHistoryID = ((UIButton*) sender).tag;
-    NSArray* msgArray = [[DataLayer sharedInstance] messagesForHistoryIDs:@[[NSNumber numberWithInteger:msgHistoryID]]];
-    if(![msgArray count])
+    MLMessage* msg = [MLMessage createMessageFromHistoryID:[NSNumber numberWithInteger:msgHistoryID]];
```
tmolitor-stud-tu:
```suggestion
    MLMessage* msg = [MLMessage createMessageFromHistoryID:@(msgHistoryID)];
```

Monal/Classes/chatViewController.m
```diff
@@ -2512,15 +2475,10 @@ -(UISwipeActionsConfiguration*) tableView:(UITableView*) tableView trailingSwipe
         {
             [self.xmppAccount retractMessage:message];
             [[DataLayer sharedInstance] retractMessageHistory:message.messageDBId];
-            [message updateWithMessage:[[[DataLayer sharedInstance] messagesForHistoryIDs:@[message.messageDBId]] firstObject]];
-
-            //update table entry
```
tmolitor-stud-tu, line 2517: Shouldn't the deleted/retracted message still be removed from the messages list? Currently the message is still in there, but its text gets reset to an empty string.

Monal/Monal.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,53 +6,62 @@
...
       },
       {
         "package": "CocoaLumberjack",
         "repositoryURL": "https://github.com/cocoalumberjack/cocoalumberjack",
         "state": {
           "branch": null,
-          "revision": "4b8714a7fb84d42393314ce897127b3939885ec3",
-          "version": "3.8.5"
+          "revision": "a9ed4b6f9bdedce7d77046f43adfb8ce1fd54114",
+          "version": "3.9.0"
```
tmolitor-stud-tu, line 19: Not a good idea without adapting to the new release, from the release notes:

> DDFileLogger: File Locking is now optional

We need to explicitly turn locking on. We'd get corrupted rawlog files without locking. That was the reason why I added locking upstream. But since some folks complained [over here](https://github.com/CocoaLumberjack/CocoaLumberjack/issues/1424), the locking was made optional.

Please implement the `shouldLockLogFile:` method in `MLLogFileManager.m` always returning `YES`, see https://github.com/CocoaLumberjack/CocoaLumberjack/pull/1425/files

Monal/Monal.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,53 +6,62 @@
...
-          "revision": "cf77c2f124df1423d90a9a1985e9b9ccfa4b9b3e",
-          "version": "1.3.0"
+          "revision": "0fccfdc93a01c8555cec86bf2c64caee30c23704",
+          "version": "0.9.11"
         }
       },
       {
-        "package": "FrameUp",
-        "repositoryURL": "https://github.com/ryanlintott/FrameUp",
+        "package": "GiphyUISDK",
```
tmolitor-stud-tu: Can we somehow get rid of this entirely? I don't want code of some third party framework with unknown security issue handling to run inside our app (even if the input method was disabled, the `+load` and `+initialize` methods might still run).

Monal/Classes/Quicksy_RegisterAccount.swift
```diff
@@ -447,7 +447,7 @@ struct Quicksy_RegisterAccount: View {
         })
         .padding()
         .addLoadingOverlay(overlay)
-        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name("kXMPPError")).receive(on: RunLoop.main)) { notification in
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kXMPPError)).receive(on: RunLoop.main)) { notification in
```
tmolitor-stud-tu, line 450: Where are all these new constants defined in? I know they are `#define` constants in ObjC, but these unfortunately don't automatically leak to Swift afaik, see the top of `SwiftHelpers.swift` and `HelperTools getObjcDefinedValue:`.

tmolitor-stud-tu (changes requested):
That's a full review of all changes now :)

Monal/Classes/Quicksy_RegisterAccount.swift
lissine0: The constants are defined in MLConstants.h, which is part of monalxmpp.
monalxmpp is globally imported in our Swift source files.
And since commit https://github.com/monal-im/Monal/commit/12c4d23ed52b93, MLConstants.h has public visibilty.
Thus we don't need to manually import constants from objective-C to Swift anymore.

Monal/Monal.xcworkspace/xcshareddata/swiftpm/Package.resolved
lissine0: You already did so in https://github.com/monal-im/Monal/commit/76ff3612a3279 :)

Monal/Classes/chatViewController.m
lissine0: Retracted and moderated messages don't get deleted, but rather have their text changed to "This message got retracted".
See https://github.com/monal-im/Monal/blob/198778f9889469a/Monal/Classes/chatViewController.m#L2245
That string is not stored in the DB / MLMessage object, in order for it to be localizeable.

The notification name might be a bit confusing. `kMonalDeletedMessageNotice` is only related to message retraction and moderation, not local deletion.

Monal/Classes/ChatView.swift
lissine0: Doing that causes the following error on L502:
```
message: ChatViewMessage(message),
                         ^^^^^^^ 
Cannot convert value of type 'ObservableKVOWrapper<MLMessage>' to expected argument type 'MLMessage'
```
If I replace `message` with `message.obj`, the automatic view updates won't work anymore

Monal/Classes/ChatView.swift
lissine0: I tested this suggestion and the view auto updates stopped working.
If the initializer of ChatViewMessage doesn't run on every view render, how are changes from MLMessage going to be picked up?

tmolitor-stud-tu: They are going to be picked up by the `ObservableKVOWrapper` which is emitting a `objectWillChange` event if a property of `MLMessage` (being used by the swiftui view) changes. That event is used by swiftui as a rerender signal.

Monal/Classes/chatViewController.m
tmolitor-stud-tu: Ah okay, on Conversations the moderated/retracted messages are removed from the ui...maybe we should use the opportunity to change Monal's behavior to be the same. What do you think?

Monal/Monal.xcworkspace/xcshareddata/swiftpm/Package.resolved
tmolitor-stud-tu: Ah, all right then :)

Monal/Classes/Quicksy_RegisterAccount.swift
tmolitor-stud-tu: Wow, that's great! :)
You can remove all these manual imports and the HelperTools method used for this, then :)

tmolitor-stud-tu: Ah I see, already done in https://github.com/monal-im/Monal/pull/1426/commits/901d54aff2a3baf3283151b4d6cac6c03f29a96b :)

Monal/Classes/ChatView.swift
lissine0: > if a property of MLMessage **(being used by the swiftui view)** changes

The `MessageView` isn't using any `MLMessage` properties directly.
The `MessageView` uses `ChatViewMessage`; and `ChatViewMessage` gets initialized with `MLMessage` properties.
Thus the need for the initializer of `ChatViewMessage` to run in order to get updates from `MLMessage`

lissine0: We need to write our custom MessageView, and use MLMessage properties in it directly, in order for it to work how you said

Monal/Classes/MLMessage.m
tmolitor-stud-tu: Did you solve this one yet?

lissine0: Yes

tmolitor-stud-tu: Great! :)

Monal/Classes/ChatView.swift
tmolitor-stud-tu: No, you should change the code to "bubble that change up", aka: only wrap the MLMessage object into the kvo observer once and pass around the kvo observer or some object containing it rather then recreating things over and over again on every render.

Monal/Classes/chatViewController.m
lissine0: Conversations doesn't support retraction yet.
I think the placeholder for retracted messages should remain. Because retraction is not limited by time, and your contact can retract a message that you already read. And later you can re-read the history and wonder if there a was a message there or you imagined it.

As for moderation, it usually is spam cleaning in public channels. Deletion is fine in this case, since the server removes the messages from the MAM archive anyway (on the other hand, no server currently removes retracted messages from the MAM archive)
Also, spam can come in very large number of messages, which can be annoying if we keep the "this message got retracted" placeholder.
I _think_ Gajim and Cheogram compress all (subsequent?) moderated messages into one, containing the reason of the moderation

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Yes, I alreay said in some review that you should overwrite the properties of `Message` in `ChatViewMessage` using computed properties returning the values of `MLMessage` and use the `MLMessage` as `@StateObject`. That should work.

See https://github.com/monal-im/Monal/pull/1426#discussion_r2083620519

That means something like this (maybe you could try a proof of concept first, before updating this PR to save some work in case it doesn't work like depicted below):
1. A view using an `@StateObject` to a class named `ChatViewMessage` and using its `text` property in the view body
2. The `ChatViewMessage` class being a child of the ExyteChat `open class Message: ObservableObject, Identifiable`
3. The `ChatViewMessage` class having a property `innerMessage` of type `ObservableKVOWrapper<MLMessage>`
4.  The `ChatViewMessage` class having an `@Published` computed property of type `String` named `text`, that always returns the `innerMessage.messageText` value. This property overwrites the `@Published public var text: String` property of the base class

Okay, step 4 is a problem here, apparently you cannot overwrite a stored property with a computed one :(
I hate Swift, in ObjC that could easily be done :(

**Solution:**
1. Check if the proof of concept above works while not deriving the `ChatViewMessage` class from anything. (Only check if the computed property properly forwards the change event of the `ObservableKVOWrapper` wrapped `MLMessage` object (use some timer to periodically change the text).
2. If that works, then either:
    1. Change the stored properties in the base class to be computed ones (computed ones can be overwritten).
    2. Update the ExyteChat to define a `protocol` named `MessageProtocol` that defines all the properties of the ExyteChat `Message` class/struct; let the ExyteChat `Message` derive from that protocol and change all mentions of `Message` in the ExyteChat codebase to be `MessageProtocol`  ones; then derive our implementation from that protocol, too, instead of deriving it from the `Message` class.

_I think introducing a new protocol is the way to go here, because it makes the original ExyteChat stuff more extensible but doesn't change any behavior. I think we can even upstream that change (together with making `Message` a class rather than a struct)._

btw: the `public enum Status` doesn't contain a state for `received`, too, so we'll have to update the base class anyways.

Monal/Classes/chatViewController.m
tmolitor-stud-tu: Hmmm, yes, good point.

tmolitor-stud-tu (changes requested):
Do you want to do the ExyteChat Protocol change in a new PR or in this one?

lissine0:
> Do you want to do the ExyteChat Protocol change in a new PR or in this one?

Let's keep that in a separate PR

Monal/Classes/ChatView.swift
tmolitor-stud-tu: @matthewrfennell what do you think, would that protocol be upstreamable?

tmolitor-stud-tu:
> > Do you want to do the ExyteChat Protocol change in a new PR or in this one?
> 
> Let's keep that in a separate PR

Okay, I created an issue for it over here: https://github.com/monal-im/Monal/issues/1462

PR #1427, opened 2025-05-11
6.4.11-b4

tmolitor-stud-tu:
- Properly open links in external browser, if configured to do so
- Fix accidental reordering of chats when opening a chat


lissine0:
You can include #1420 and #1422 in the next beta / stable.
They fix the wrong chat-list order issue, and they're easy to backport

PR #1452, opened 2025-08-07
Update opensource.html

lissine0:
Use the `<details>` HTML element to hide license texts by default.

Remove EARestrictedScrollView and PSTCollectionView, as they're no longer used.

Add the licenses and copyright notices of the following dependencies:
- ExyteChat (chat UI framework)
- MediaPicker (dependency of ExyteChat)
- ActivityIndicatorView (dependency of ExyteChat)
- KSCrash (crash reporter)
- ASN1Decoder (ASN1 DER Decoder for X.509 Certificate)
- FLAnimatedImagea (animated GIF engine)
- swift-log (dependency of CocoaLumberjack)
- ViewExtractor

### DIFF
--- a/Monal/opensource.html
+++ b/Monal/opensource.html
@@ -36,7 +36,14 @@
         Parts of the WebRTC Demo by Stasel<br><br>
         <a href="https://github.com/stasel/WebRTC-iOS">https://github.com/stasel/WebRTC-iOS</a><br>
         <a href="https://github.com/stasel/WebRTC-iOS/blob/main/WebRTC-Demo-App/Sources/Services/WebRTCClient.swift">https://github.com/stasel/WebRTC-iOS/blob/main/WebRTC-Demo-App/Sources/Services/WebRTCClient.swift</a><br>
-        TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION<br>
+        <br>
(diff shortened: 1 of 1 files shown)


Monal/opensource.html
```diff
@@ -129,14 +146,145 @@
         OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
         SOFTWARE.<br>
         <br>
-        
+        </details>
+
+        <hr>
+        ExyteChat: A SwiftUI Chat UI framework with fully customizable message cells and a built-in media picker<br>
+        <a href="https://github.com/exyte/Chat">https://github.com/exyte/Chat</a><br>
```
lissine0: Should I link to our fork instead?

tmolitor-stud-tu: good question, for the time being: yes

tmolitor-stud-tu (changes requested):
I think we can remove the verbatim copy of gplv3 and just link to the license tet instead, other than that: thank you for this quik pr! :)

lissine0:
> I think we can remove the verbatim copy of gplv3 and just link to the license text instead

With the HTML `<details>`, the license text is not visible unless you click on it. But as you wish

PR #1457, opened 2025-08-11
Copy contact addresses as xmpp URIs

lissine0:
Keep displaying the address as is.
But when copying to the clipboard, make it an `xmpp:` link so it can be clickable when shared.

### DIFF
--- a/Monal/Classes/ContactDetails.swift
+++ b/Monal/Classes/ContactDetails.swift
@@ -159,7 +159,8 @@ struct ContactDetails: View {
                     }
                     
                     Button {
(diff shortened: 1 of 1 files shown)


tmolitor-stud-tu:
I think adding a share button would be much better. That could even show a menu to let the user choose if they want to share using an xmpp: uri or a qr code, like in conversations. Sharing the link should open ios share menu, too (which includes a copy, but allows the user to share using sms, email etc., too)

lissine0:
I'm closing this PR, as it's not a high priority for now.

PR #1461, opened 2025-08-17
Fix bookmark pinning

lissine0:
Tested with Movim.

### DIFF
--- a/Monal/Classes/MLPubSubProcessor.m
+++ b/Monal/Classes/MLPubSubProcessor.m
@@ -296,7 +296,7 @@ $$class_handler(bookmarks2Handler, $$ID(xmpp*, account), $$ID(NSString*, jid), $
                 }
             }
             //check if pinned status changed (the check is done inside of [MLContact togglePinnedChat:]
-            else if([ownFavorites containsObject:room])
+            if([ownFavorites containsObject:room])
             {
                 MLContact* contact = [MLContact createContactFromJid:room andAccountID:account.accountID];
                 [contact togglePinnedChat:pinned];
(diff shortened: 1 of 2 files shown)


tmolitor-stud-tu:
What bug exactly is this fixing?

lissine0:
> What bug exactly is this fixing?

Currently on the latest alpha, if the user presses the pinned toggle for a MUC, it immediately reverts back to its previous state.
Moreover, the bookmark doesn't get updated with the `<pinned/>` element

lissine0:
One of the bugs is because of the else here https://github.com/monal-im/Monal/blob/198778f98894/Monal/Classes/MLPubSubProcessor.m#L299 , since one of the previous `else if` blocks always matches. (especially the one where the room already exists in the DB, and the bookmark has a nickname)

lissine0:
The other bug is that we simply never publish XEP-0469 updates to the server, unless when creating a bookmark for the first time.

Also note: the CI is failing because Github is changing the `macos-latest` runner from `macos-14` to `macos-15`
(https://github.com/actions/runner-images/issues/12520)
It seems to be a bug on their end.

tmolitor-stud-tu:
Ah, yes, now I see it, thanks for fixing :)

Monal/Classes/MLPubSubProcessor.m
```diff
@@ -403,11 +403,34 @@ @implementation MLPubSubProcessor
...
+        //sync the bookmark with the local pinned status, if they differ
         else if([ownFavorites containsObject:room])
         {
             MLContact* contact = [MLContact createContactFromJid:room andAccountID:account.accountID];
-            [contact togglePinnedChat:pinned];
+            if(pinned == contact.isPinned)
+                continue;
+            else if(contact.isPinned)
+            {
+                DDLogVerbose(@"Adding <pinned> element to the bookmark of the room %@", room);
```
tmolitor-stud-tu: Could you merge all of this into the updating `if` in line 383?

tmolitor-stud-tu, line 414: Could you merge all of this into the updating if in line 383?

tmolitor-stud-tu (changes requested):

Monal/Classes/MLPubSubProcessor.m
lissine0: But we want to execute this code when the condition in that `if` is false.
Did you mean copying the code there, so it always executes?

tmolitor-stud-tu: No, I meant to merge that into the body of that if and change the condition of the if to match "we know the bookmark, but either the nick changed OR the pinning changed".

tmolitor-stud-tu: If I don't miss something, you're missing the publishing now. This is missing:
```
            //publish this bookmark item again
            [account.pubsub publishItem:item onNode:@"urn:xmpp:bookmarks:1" withConfigOptions:@{
                @"pubsub#persist_items": @"true",
                @"pubsub#access_model": @"whitelist",
                @"pubsub#max_items": max_items,
            } andHandler:$newHandler(self, bookmarks2Published, $ID(room))];
```

Also one `else if` can really be an `else` since we are checking a boolean ;)

Monal/Classes/MLPubSubProcessor.m
```diff
@@ -378,37 +378,55 @@ @implementation MLPubSubProcessor
...
+                {
+                    DDLogVerbose(@"Adding <pinned> element to the bookmark of the room %@", room);
+                    if([item check:@"{urn:xmpp:bookmarks:1}conference/extensions"])
+                        [[item findFirst:@"{urn:xmpp:bookmarks:1}conference/extensions"] addChildNode:[[MLXMLNode alloc] initWithElement:@"pinned" andNamespace:@"urn:xmpp:bookmarks-pinning:0"]];
+                    else
+                        [[item findFirst:@"{urn:xmpp:bookmarks:1}conference"] addChildNode: [[MLXMLNode alloc] initWithElement:@"extensions" withAttributes:@{} andChildren:@[
+                            [[MLXMLNode alloc] initWithElement:@"pinned" andNamespace:@"urn:xmpp:bookmarks-pinning:0" withAttributes:@{} andChildren:@[] andData:nil],
+                        ] andData:nil]];
+                }
+                else if(!contact.isPinned)
```
tmolitor-stud-tu:
```suggestion
                else
```

tmolitor-stud-tu (approved):
Just this else, everything else is fine now :)

Monal/Classes/MLPubSubProcessor.m
tmolitor-stud-tu: Ah, my failure :D

tmolitor-stud-tu:
Is this fix something that should be backported to stable, too?

lissine0:
No.
The bookmarks pinning XEP was implemented in alpha only.
Thus the regressions only affected the alpha

lissine0:
(These two commits are not in stable: https://github.com/monal-im/Monal/commit/152f3173ca4df7160e24b27619bf591531412630 and https://github.com/monal-im/Monal/commit/ea008938f4bfd39882bf4619afd5e38a712831f5)

PR #1465, opened 2025-08-19
Dynamically load more history when scrolling up

lissine0:
Notes:
The `pages` parameter in ExyteChat's `enableLoadMore` is not taken into consideration. `pages` is hardcoded to `1` in ExyteChat's source code. See Matthew's open PR for more details.

When loading more history from the _db_, I occasionally triggered a bug, which I suspect is a race condition, where the `messages` array gets update but that doesn't result in a SwiftUI re-render. It turns out that ExyteChat has an optimization that says: Only render insertion updates if the view is scrolled to the very top or the very bottom.[[0]](https://github.com/exyte/Chat/blob/18c59301822/README.md?plain=1#L256)   [[1]](https://github.com/exyte/Chat/blob/18c59301822/Sources/ExyteChat/Views/UIList.swift#L192)
So, I disabled that optimization in our fork. The bug doesn't happen anymore, and the experience is smoother.
(Note: when triggering this bug, tapping the top message / date separator makes the new messages show up)

##### Changes from the UIKit backscrolling:
Use a ProgressView() spinner when fetching history from the server, but not from the DB.

History fetching happens automatically when the top message is rendered, rather than on pull-to-refresh. This change is baked into ExyteChat.

Since History fetches will happen when a single message is loaded in the ChatView, I went ahead and enabled automatic fetches when the ChatView is empty as well.

It can be annoying when the alert "finished fetching messages successfully" shows up every time you open a chat with a small number of messages. Therefore:
1- If we reach the beginning of the available MAM history, we disable further MAM queries =>  the ProgressView() doesn't spin anymore when scrolling up => this is not a confusing UX, as this will implicitly tell users that there's no more history to fetch
2- We don't show the "finished fetching messages successfully" alert. Because it is weird when you get that alert immediately after creating a group chat / adding a contact. (remember, we fetch history automatically even when the chatView is empty)

I want your opinion on also disabling the alert / MAM queries when we get a `element-not-found` error. This error happens when the oldest stanzaId in the chat is not known to the server (i.e. it was deleted because of the retention policy). This case should be indistinguishable to users from the "finished fetching messages successfully" case.

### DIFF
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -91,6 +91,7 @@ struct ChatView: View {
     @StateObject private var overlay = LoadingOverlayState()
     @State var messages: [ChatViewMessage] = []
     private var account: xmpp
+    @State private var isLoadingMamHistory = false
     
     init(contact: ObservableKVOWrapper<MLContact>) {
         _contact = StateObject(wrappedValue: contact)
@@ -194,7 +195,85 @@ struct ChatView: View {
         }
 #endif
     }
-    
+
+    private func findOldestStanzaId() -> String? {
+        //not all messages in history db have a stanzaId (messages sent by this monal instance won't have one for example)
+        //--> search for the oldest message having a stanzaId and use that one
+        for msg in self.messages {
+            if !msg.innerMessage.obj.stanzaId.isEmpty {
+                DDLogVerbose("Found oldest stanzaId in currently displayed messages: \(msg.innerMessage.obj.stanzaId)")
+                return msg.innerMessage.stanzaId
+            }
+        }
+
+        //history database for this contact is completely empty, use global last stanza id for this mam archive
+        if self.contact.isMuc {
+            return DataLayer.sharedInstance().lastStanzaId(forMuc:self.contact.contactJid, andAccount:self.account.accountID)
+        }
+        else {
+            return DataLayer.sharedInstance().lastStanzaId(forAccount:self.account.accountID);
+        }
+    }
+
+    private func loadHistory() {
+        guard !self.isLoadingMamHistory else {
+            DDLogDebug("Not going to load more history because a backscrolling mam query is ongoing for this chat")
+            // Don't allow concurrent backscrolling mam fetches.
+            // And don't allow loading history from the DB while an mam query is ongoing;
+            // this prevents potential message duplication: if a DB history fetch is triggered after the mam query
+            // has written some messages to the DB but before it has returned the results, those messages will be
+            // inserted twice in the ChatView.
+            return
+        }
+        var beforeId: NSNumber? = nil
+        if !messages.isEmpty {
+            beforeId = messages[0].innerMessage.messageDBId
+        }
+        let oldMessages: [ChatViewMessage] = (DataLayer.sharedInstance().messages(forContact:contact.contactJid, forAccount:contact.accountID, beforeMsgHistoryID:beforeId) as! [MLMessage]).map {ChatViewMessage($0)}
+        //insert what we got from the db
+        if !oldMessages.isEmpty {
+            messages.insert(contentsOf: oldMessages, at:0)
+        }
+
+        // if we didn't get enough (or any) messages from the DB, try to get some from MAM
+        if oldMessages.count < kMonalBackscrollingMsgCount && !contact.hasReachedMamArchiveTop {
+            self.isLoadingMamHistory = true
+
+            guard let oldestStanzaId = self.findOldestStanzaId() else {
+                DDLogError("Couldn't find a stanzaId to perform a mam query with! (oldestStanzaId is nil)")
+                return
+            }
+            // now try to load more (older) messages from mam
+            DDLogVerbose("Loading more messages from mam before stanzaId \(oldestStanzaId)")
+            firstly {
+                self.account.setMAMQueryMostRecentFor(self.contact.obj, before: oldestStanzaId)
+            }
+            .done { returnedMessages in
+                let returnedMessages = returnedMessages as! [MLMessage]
+                if returnedMessages.isEmpty {
+                    DDLogVerbose("Reached the top of the mam archive for \(self.contact.obj.contactJid)")
+                    // Don't block the main thread while writing to the db
+                    DispatchQueue.global(qos: .default).async {
+                        contact.obj.markReachedMamArchiveTop()
+                    }
+                } else {
+                    DDLogVerbose("Got backscrolling mam response: \(returnedMessages.count) messages")
+                    self.messages.insert(contentsOf: returnedMessages.map {ChatViewMessage($0)}, at: 0)
+                }
+            }
+            .catch { error in
+                alertPrompt = AlertPrompt(
+                    title: Text("Could not fetch messages"),
+                    message: Text(error.localizedDescription),
+                    dismissLabel: Text("Close")
+                )
+            }
+            .finally {
+                self.isLoadingMamHistory = false
+            }
+        }
+    }
+
     var body: some View {
         ExyteChatView(messages: messages, chatType: .conversation, replyMode: .quote) { draft in
             guard let newMLMessage = MLXMPPManager.sharedInstance().sendMessageAndAddToHistory(message: draft.text, havingType: kMessageTypeText, toContact: self.contact.obj, isEncrypted: self.contact.isEncrypted, uploadInfo: nil) else {
@@ -205,9 +284,11 @@ struct ChatView: View {
             MessageView(message: (message as! ChatViewMessage), viewModel: viewModel, positionInUserGroup: positionInUserGroup, positionInMessagesSection: positionInMessagesSection)
         }
         .showNetworkConnectionProblem(false)
-//         .enableLoadMore(pageSize: 3) { message in
-//             print("load more messages before: \(String(describing:message))")
-//         }
+        .enableLoadMore(pageSize: 10) { message in
+            await MainActor.run {
+                loadHistory()
+            }
+        }
 //         .messageUseMarkdown(messageUseMarkdown: true)
         .sheet(item: $selectedContactForContactDetails) { selectedContact in
             AnyView(AddTopLevelNavigation(withDelegate:nil, to:ContactDetails(delegate:nil, contact:selectedContact)))
@@ -270,6 +351,9 @@ struct ChatView: View {
             }
             
             ToolbarItemGroup(placement: .topBarTrailing) {
+                ProgressView()
+                    .opacity(isLoadingMamHistory ? 1 : 0)
+
                 if !(contact.isMuc || contact.isSelfChat) {
                     let activeChats = (UIApplication.shared.delegate as! MonalAppDelegate).activeChats!
                     let voipProcessor = (UIApplication.shared.delegate as! MonalAppDelegate).voipProcessor!
@@ -346,15 +430,7 @@ struct ChatView: View {
             MLNotificationManager.sharedInstance().currentContact = self.contact.obj
 
             checkOmemoSupport(withAlert:false)
-            
-            //TODO: load messages from db
-            // In some scenarios, onAppear can be executed more than once without the state variables being cleared.
-            if messages.isEmpty {
-                let dbMessages = DataLayer.sharedInstance().messages(forContact:contact.contactJid, forAccount:contact.accountID) as! [MLMessage]
-                for msg in dbMessages {
-                    messages.append(ChatViewMessage(msg))
-                }
-            }
+            loadHistory()
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
 //             messages = [
 //                 ExyteChat.Message(
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -477,7 +477,7 @@ struct ChatView: View {
             }
         }
         .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kMonalNewMessageNotice)).receive(on: RunLoop.main)) { notification in
-            DDLogVerbose("chat view got new message notice \(String(describing:notification.userInfo))")
+            DDLogVerbose("ChatView got new message notice \(String(describing:notification.userInfo))")
 
             guard let message = notification.userInfo?["message"] as? MLMessage else {
                 DDLogError("Notification without message");
@@ -492,6 +492,16 @@ struct ChatView: View {
         }
         .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kMonalRefresh)).receive(on: RunLoop.main)) { notification in
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
+        }
+        .onReceive(NotificationCenter.default.publisher(for: NSNotification.Name(kMonalContactHistoryCleared)).receive(on: RunLoop.main)) { notification in
+            DDLogVerbose("ChatView got history cleared notice \(String(describing:notification.userInfo))")
+            guard let contact = notification.userInfo?["contact"] as? MLContact else {
+                MLAssert(false, "Notification without contact")
+                return
+            }
+            if contact.isEqual(to: self.contact.obj) {
+                self.messages = []
+            }
         }
     }
 }
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -136,64 +136,64 @@ struct ChatView: View {
     
     private func checkOmemoSupport(withAlert showWarning: Bool) {
 #if !DISABLE_OMEMO
-    if DataLayer.sharedInstance().isAccountEnabled(contact.accountID) {
-        var omemoDeviceForContactFound = false
-        if !contact.isMuc {
-            omemoDeviceForContactFound = account.omemo.knownDevices(forAddressName:contact.contactJid).count > 0
-        } else {
-            omemoDeviceForContactFound = false
-            for participant in DataLayer.sharedInstance().getMembersAndParticipants(ofMuc:contact.contactJid, forAccountID:contact.accountID) {
-                if let participant_jid = participant["participant_jid"] as? String {
-                    omemoDeviceForContactFound = omemoDeviceForContactFound || account.omemo.knownDevices(forAddressName:participant_jid).count > 0
-                } else if let participant_jid = participant["member_jid"] as? String {
-                    omemoDeviceForContactFound = omemoDeviceForContactFound || account.omemo.knownDevices(forAddressName:participant_jid).count > 0
-                }
-                if omemoDeviceForContactFound {
-                    break
+        if DataLayer.sharedInstance().isAccountEnabled(contact.accountID) {
+            var omemoDeviceForContactFound = false
+            if !contact.isMuc {
+                omemoDeviceForContactFound = account.omemo.knownDevices(forAddressName:contact.contactJid).count > 0
+            } else {
+                omemoDeviceForContactFound = false
+                for participant in DataLayer.sharedInstance().getMembersAndParticipants(ofMuc:contact.contactJid, forAccountID:contact.accountID) {
+                    if let participant_jid = participant["participant_jid"] as? String {
+                        omemoDeviceForContactFound = omemoDeviceForContactFound || account.omemo.knownDevices(forAddressName:participant_jid).count > 0
+                    } else if let participant_jid = participant["member_jid"] as? String {
+                        omemoDeviceForContactFound = omemoDeviceForContactFound || account.omemo.knownDevices(forAddressName:participant_jid).count > 0
+                    }
+                    if omemoDeviceForContactFound {
+                        break
+                    }
                 }
             }
-        }
-        if !omemoDeviceForContactFound && contact.isEncrypted {
-            if HelperTools.isContactBlacklistedForEncryption(contact.obj) {
-                // this contact was blacklisted for encryption
-                // --> disable it
-                contact.isEncrypted = false
-                DataLayer.sharedInstance().disableEncrypt(forJid:contact.contactJid, andAccountID:contact.accountID)
-            } else if contact.isMuc && contact.mucType != kMucTypeGroup {
-                // a channel type muc has OMEMO encryption enabled, but channels don't support encryption
-                // --> disable it
-                contact.isEncrypted = false
-                DataLayer.sharedInstance().disableEncrypt(forJid:contact.contactJid, andAccountID:contact.accountID)
-            } else if !contact.isMuc || (contact.isMuc && contact.mucType == kMucTypeGroup) {
+            if !omemoDeviceForContactFound && contact.isEncrypted {
+                if HelperTools.isContactBlacklistedForEncryption(contact.obj) {
+                    // this contact was blacklisted for encryption
+                    // --> disable it
+                    contact.isEncrypted = false
+                    DataLayer.sharedInstance().disableEncrypt(forJid:contact.contactJid, andAccountID:contact.accountID)
+                } else if contact.isMuc && contact.mucType != kMucTypeGroup {
+                    // a channel type muc has OMEMO encryption enabled, but channels don't support encryption
+                    // --> disable it
+                    contact.isEncrypted = false
+                    DataLayer.sharedInstance().disableEncrypt(forJid:contact.contactJid, andAccountID:contact.accountID)
+                } else if !contact.isMuc || (contact.isMuc && contact.mucType == kMucTypeGroup) {
+                    hideLoadingOverlay(overlay)
+
+                    if showWarning {
+                        DDLogWarn("Showing omemo not supported alert for: \(self.contact)")
+                        alertPrompt = AlertPrompt(
+                            title: Text("No OMEMO keys found"),
+                            message: Text("This contact may not support OMEMO encrypted messages. Please try to enable encryption again in a few seconds, if you think this is wrong."),
+                            dismissLabel: Text("Disable Encryption")
+                        ) {
+                            contact.isEncrypted = false
+                            DataLayer.sharedInstance().disableEncrypt(forJid:contact.contactJid, andAccountID:contact.accountID)
+                        }
+                    } else {
+                        DDLogInfo("Trying to fetch omemo keys for: \(self.contact)")
+
+                        // we won't do this twice, because the user won't be able to change isEncrypted to YES,
+                        // unless we have omemo devices for that contact
+                        showPromisingLoadingOverlay(overlay, headlineView:Text("Loading OMEMO keys"), descriptionView:Text("")).done {
+                            // request omemo devicelist
+                            account.omemo.subscribeAndFetchDevicelistIfNoSessionExists(forJid:contact.contactJid)
+                        }
+                    }
+                }
+            } else {
                 hideLoadingOverlay(overlay)
-                
-                if showWarning {
-                    DDLogWarn("Showing omemo not supported alert for: \(self.contact)")
-                    alertPrompt = AlertPrompt(
-                        title: Text("No OMEMO keys found"),
-                        message: Text("This contact may not support OMEMO encrypted messages. Please try to enable encryption again in a few seconds, if you think this is wrong."),
-                        dismissLabel: Text("Disable Encryption")
-                    ) {
-                        contact.isEncrypted = false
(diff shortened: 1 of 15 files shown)


Monal/Classes/ChatView.swift
```diff
@@ -205,9 +287,11 @@ struct ChatView: View {
             MessageView(message: (message as! ChatViewMessage).message, viewModel: viewModel, positionInUserGroup: positionInUserGroup, positionInMessagesSection: positionInMessagesSection)
         }
         .showNetworkConnectionProblem(false)
-//         .enableLoadMore(pageSize: 3) { message in
-//             print("load more messages before: \(String(describing:message))")
-//         }
+        .enableLoadMore(pageSize: 10) { message in
+            DispatchQueue.main.async {
+                loadHistory()
+            }
```
lissine0, line 292: Is it okay to do database reads on the main thread?

Monal/Classes/ChatView.swift
```diff
@@ -136,65 +137,146 @@ struct ChatView: View {
...
+                                message: Text("Could not fetch (all) old messages for this chat from your server archive. Please try again later. \(errorText!)"),
+                                dismissLabel: Text("Close")
+                            )
+                        } else {
+                            DDLogVerbose("Got backscrolling mam response: \(returnedMessages!.count) messages")
+                            if returnedMessages!.count != 0 {
+                                self.messages.insert(contentsOf: (returnedMessages as! [MLMessage]).map {ChatViewMessage(ObservableKVOWrapper($0))}, at: 0)
+                            }
+                            else {
+                                contact.obj.markReachedMamArchiveTop()
```
lissine0: This method call does a database write, but it's only updating a one boolean. Should I do it in a background thread instead?

Monal/Classes/MLContact.h
```diff
@@ -97,6 +97,10 @@ FOUNDATION_EXPORT NSString* const kAskSubscribe;
 @property (nonatomic, readonly) NSString* contactDisplayName;
 @property (nonatomic, readonly) NSString* contactDisplayNameWithoutSelfnotesPrefix;
 
+// This property is used to avoid querying MAM if the top of the archive
+// was already reached in a previous query
+@property (nonatomic, readonly) BOOL hasReachedMamArchiveTop;
```
lissine0, line 102: Is the property's name okay? or do you prefer something like `MamQueriesDisabled` or `fullHistoryFetched` or `fullHistorySynced`?

Monal/Classes/ChatView.swift
tmolitor-stud-tu: yes, sure, we use multithreaded/multiconnection sqlite3 in wal mode: every thread gets its own db connection.
--> reads should never be blocked by db writes on other threads (nor, obviously, by other threads' reads)

Monal/Classes/ChatView.swift
tmolitor-stud-tu: yes please. sqlite3 in wal mode can only have one writing thread and in catchup/heavy traffic phases, that's most likely the receive queue and that would block the main thread.

tmolitor-stud-tu: most, if not all writes to the db should not be done in the main thread but in a bg thread (the only thing I'd write in the main thread would be new sent messages because we want them in our db the instance we show them on screen [otherwise users could think the message was written, force close the app and loose the message if the db was heavily congested])

Monal/Classes/MLContact.h
tmolitor-stud-tu: the name is okay imho :) it is even more descriptive than your other suggestions :)

tmolitor-stud-tu:
> I want your opinion on also disabling the alert / MAM queries when we get a element-not-found error.

yes, that's a good idea.

Monal/Classes/ChatView.swift
lissine0: > most, if not all writes to the db should not be done in the main thread but in a bg thread

Even in cases like this, where the db write is tied to a toggle / button's state? https://github.com/monal-im/Monal/blob/3ffed6f5f5831/Monal/Classes/ContactDetails.swift#L366

Monal/Classes/DataLayerMigrations.m
```diff
@@ -963,6 +963,10 @@ FOREIGN KEY('account_id') REFERENCES 'account'('account_id') ON DELETE CASCADE \
             );"];
         }];
 
+        // Keep track of when the full MAM archive has been fetched, so we don't query it again
+        [self updateDB:db withDataLayer:dataLayer toVersion:6.413 withBlock:^{
```
tmolitor-stud-tu: this should be 7.000 because it will be part of Monal 7.0

Monal/Classes/ChatView.swift
tmolitor-stud-tu: no, that's similar to the sending message case I mentioned. rule of thumb: no bg if directly visible change in ui is tied to that db write, bg for all other cases...

Monal/Classes/MLContact.m
```diff
@@ -719,7 +719,9 @@ -(void) addToRoster
 -(void) clearHistory
 {
     [[DataLayer sharedInstance] clearMessagesWithBuddy:self.contactJid onAccount:self.accountID];
-    [[MLNotificationQueue currentQueue] postNotificationName:kMonalRefresh object:nil userInfo:nil];
+    [[MLNotificationQueue currentQueue] postNotificationName:kMonalContactHistoryCleared object:nil userInfo:@{
```
tmolitor-stud-tu: it would be nice to set `object` to the `xmpp*` instance, the contact belongs to.

Monal/Classes/ChatView.swift
```diff
@@ -136,65 +137,146 @@ struct ChatView: View {
...
+                return
+            }
+            // now try to load more (older) messages from mam
+            DDLogVerbose("Loading more messages from mam before stanzaId \(oldestStanzaId)")
+            self.account.setMAMQueryMostRecentFor(
+                self.contact.obj,
+                before: oldestStanzaId,
+                withCompletion: { (returnedMessages: [Any]?, error: String?) in
+                    DispatchQueue.main.async {
+                        if returnedMessages == nil && error == nil {
```
tmolitor-stud-tu: we should take the opportunity to rewrite the history mam query to not use ephemeral iq handling blocks, but MLHandler (in MLIQProcessor) and matthew's the new promise framework to signal success to the ui (rather than a completion callback like its currently done).

That would prevent the "The connection to the server was interrupted and no old messages could be fetched for this chat. Please try again later." error (which afaik sometimes spuriously pops up even though there was no disconnect).

Monal/Classes/MLContact.m
```diff
@@ -713,7 +719,15 @@ -(void) addToRoster
...
+        @"contact": self,
+    }];
+}
+
+-(void) markReachedMamArchiveTop
```
tmolitor-stud-tu, line 739: why use a dedicated method instead of a property setter (that would be named `setHasReachedMamArchiveTop:(BOOL) value`)?
that would achieve the same, but with a much nicer syntax: `contact.hasReachedMamArchiveTop=YES;`

I'm not even sure if I like my toggle methods either. Back then when I started with swiftui I saw a few tutorials using such toggle methods, but currently that toggle leads to ugly swiftui code like this:
```
                Toggle(isOn: Binding(get: {
                    contact.isPinned
                }, set: {
                    contact.obj.togglePinnedChat($0)
                })) {
                    Text("Pin Chat")
                }
```
Whereas the KVO wrapper could already provide us with a simple `contact.isPinned=true` / `contact.isPinned=false` which would look much cleaner:
```
                Toggle(isOn: contact.isPinned) {
                    Text("Pin Chat")
                }
```
Or this, I'm not sure currently:
```
                Toggle(isOn: $contact.isPinned) {
                    Text("Pin Chat")
                }
```

tmolitor-stud-tu: Okay, well the mark has one advantage: it isn't possible to reset it to NO once it was set. So `markReachedMamArchiveTop` is better than a setter, but I still don't like my toggle methods...

tmolitor-stud-tu (changes requested):

tmolitor-stud-tu:
That's a full review now

Monal/Classes/ChatView.swift
lissine0: But this is a user-interactive action.
If the user doesn't have an internet connection, for example, shouldn't we display an error rather than an infinitely spinning ProgressView?
Or we could prevent the MAM query from being sent in the first place

tmolitor-stud-tu: I didn't say that we should ignore every error, but the `returnedMessages=nil && error=nil` case comes from the fact that we use the ephemeral iq handling rather than the `MLHandler` one. And that ephemeral handling doesn't survive any reconnect while the `MLHandler` does (in fact it even survives background switches to the appex and then back to the main app).

Matthew's new promise framework is precisely for this kind of thing: triggering an action in the UI, showing a spinner or whatever in the UI while this action is taking place, but allowing the app to go into background while performing the action (in the appex) and still show the result of the action once the app comes into foreground again. And even if the app was force closed, the action will be completed in the background (but obviously no UI result/error will be shown apart from - in this case - new messages in the history) .

lissine0: What I meant is, from a UX perspective, an infinitely-spinning ProgressView doesn't make a lot of sense when you're scrolling up while being offline.
Note that we don't save the scroll position when quitting the ChatView.
So, if the user regains connectivity later (and the app was killed, or the user swiped back to ActiveChats), the user will have to scroll up anyway for the history. At which point, why not send a MAM query at that time if being online?

tmolitor-stud-tu: Well, `sendIq:withResponseHandler:andErrorHandler:` and the `MLHandler` based `sendIq:withHandler:` behave exactly the same when being offline: they are queued. Also both are being invalidated after non-smacks reconnect, the `MLHandler` one by calling its invalidation handler, if present, and the other one by calling its error handler with `nil`.

If you want to show a message when offline and not show the spinner and queue the mam load, you can do this. But that's orthogonal to my suggested change here.

lissine0:
I updated the PR with a refactor for `setMAMQueryMostRecentForContact:` to use MLHandler and MLPromise.
But I have a problem with the Handler. It has an optional `NSArray*` parameter, but it is always received by the handler as `nil`. And if I make it a mandatory parameter, I get `Dynamic unpacking exception triggered`.

For context: we may receive too few messages containing bodies. In that case, we need to send another MAM query in order to get more messages with bodies.
I pass the promise to the new query, and I want to pass an array of historyIds (corresponding to messages processed in the previous query(ies)) to the new query; so that it can resolve all the messages retrieved in all the queries.
But for some reason that array is always received as nil by the handler of the new query.

I'm probably overlooking something related to handlers.

Monal/Classes/ChatView.swift
```diff
@@ -136,65 +137,132 @@ struct ChatView: View {
...
+            self.isLoadingMamHistory = true        //don't allow multiple parallel mam fetches
+
+            guard let oldestStanzaId = self.findOldestStanzaId() else {
+                DDLogError("Couldn't find a stanzaId to perform a mam query with! (oldestStanzaId is nil)")
+                return
+            }
+            // now try to load more (older) messages from mam
+            DDLogVerbose("Loading more messages from mam before stanzaId \(oldestStanzaId)")
+            self.account.setMAMQueryMostRecentFor(self.contact.obj, before: oldestStanzaId)
+            .done { returnedMessages in
```
tmolitor-stud-tu, line 253: You can use `async xxx.toTypedPromise().asyncOnMainActor()` instead of `.done` and `.catch` if you want more linear looking code (use try-catch to catch errors you would normally catch in `.catch`)

Example: https://github.com/monal-im/Monal/blob/develop/Monal/Classes/MediaViewer.swift#L119 (it uses `.toTypedGuarantee()` instead of `.toTypedPromise()` because the promise can never be rejected)

Monal/Classes/MLIQProcessor.m
```diff
@@ -232,6 +233,143 @@ +(void) processErrorIq:(XMPPIQ*) iqNode forAccount:(xmpp*) account
...
+
+$$class_handler(handleMAMBackscrollingResult, $$ID(xmpp*, account), $$ID(XMPPIQ*, iqNode), $$ID(MLContact*, contact), $_ID(NSArray*, historyIdsOfAlreadyRetreivedMessages), $$ID(MLPromise*, promise))
+    //the promise will be rejected if an error prevented us to get any messages.
+    //it will resolve an empty array, if the upper end of our archive was reached
+    //and it will resolve an array of newly loaded mlmessages in all other cases
+
+    //if we get less than half kMonalBackscrollingMsgCount message bodies,
+    //recursively send another query, passing it the promise and the historyIds of processed messages,
+    //and let the new query resolve the promise.
+    unsigned int retrievedBodiesOverall = 0;
```
tmolitor-stud-tu: why not use either `NSUInteger` or `uint32_t`?

Monal/Classes/MLIQProcessor.m
```diff
@@ -232,6 +233,143 @@ +(void) processErrorIq:(XMPPIQ*) iqNode forAccount:(xmpp*) account
...
+            //process received message stanzas and manipulate the db accordingly
+            //if a new message got added to the history db, the message processor will return a MLMessage instance containing the history id of the newly created entry
+            DDLogDebug(@"Handling %@ entries in mam page...", @([mamPage count]));
+            uint32_t entryNo = 0;
+            for(NSDictionary* data in mamPage)
+            {
+                //don't write data to our tcp stream while inside this db transaction
+                //(all effects to the outside world should be transactional, too)
+                [account freezeSendQueue];
+                //process all queued mam stanzas in a dedicated db write transaction
```
tmolitor-stud-tu: This needs a comment, that an app crash will leave holes in the history, because we process our queued message stanzas forward in time (which is needed for displyed, received markers etc. to work properly).
--> crashing after processing some messages will make the next mam fetch query *above* these messages while we still have a permanent hole *below* these messages (between them and our old top).

tmolitor-stud-tu: Unfortunately I don't know a good way to prevent this. Doing everything in one db transaction could block the db for write transactions for several seconds (and that could potentially block the main thread if it tries to write to the db).

Monal/Classes/MLIQProcessor.m
```diff
@@ -232,6 +233,143 @@ +(void) processErrorIq:(XMPPIQ*) iqNode forAccount:(xmpp*) account
...
+        {
+            if(
+                ![[iqNode findFirst:@"{urn:xmpp:mam:2}fin@complete|bool"] boolValue] &&
+                [iqNode check:@"{urn:xmpp:mam:2}fin/{http://jabber.org/protocol/rsm}set/first#"]
+            )
+            {
+                //page through to get more messages, and pass the current promise and overallHistoryIdList
+                //to the new query's handler
+                DDLogVerbose(@"Going to send another mam backscrolling query because we didn't get enough message bodies. We got %u bodies in this query and %u bodies in all the queries since the user scrolled up to fetch more.", retrievedBodiesInThisQuery, retrievedBodiesOverall);
+                XMPPIQ* newQuery = [[XMPPIQ alloc] initWithType:kiqSetType];
```
tmolitor-stud-tu: Don't duplicate the following code from `xmpp.m` but rather introduce a new method in `xmpp.m` to be called by `setMAMQueryMostRecentForContact:` and from here.

Monal/Classes/MLIQProcessor.m
```diff
@@ -232,6 +233,143 @@ +(void) processErrorIq:(XMPPIQ*) iqNode forAccount:(xmpp*) account
...
+            //if a new message got added to the history db, the message processor will return a MLMessage instance containing the history id of the newly created entry
+            DDLogDebug(@"Handling %@ entries in mam page...", @([mamPage count]));
+            uint32_t entryNo = 0;
+            for(NSDictionary* data in mamPage)
+            {
+                //don't write data to our tcp stream while inside this db transaction
+                //(all effects to the outside world should be transactional, too)
+                [account freezeSendQueue];
+                //process all queued mam stanzas in a dedicated db write transaction
+                [[DataLayer sharedInstance] createTransaction:^{
```
tmolitor-stud-tu: This is meant to use a dedicated db transaction per processed stanza to not block the db in a big write transaction of ~50 message stanzas. But the whole handler will already be called inside a db write transaction, so this nested transaction is a noop.

--> You need to move the handling of collected stanzas to a background thread to free up the receive queue and its db transaction. You don't need to switch to the main thread when fullfilling the promise later on, the promise lib will already ensure that the promise handlers are scheduled on the main thread.

tmolitor-stud-tu: Yes, I know my original code did not do this either. That's a bug :/

tmolitor-stud-tu (changes requested):
Some more comments and needed changes, but overall this looks pretty good now :)

Monal/Classes/ChatView.swift
lissine0: I actually prefer the promise syntax of `.done` and `.catch`

Monal/Classes/MLIQProcessor.m
lissine0: > why not use either NSUInteger or uint32_t?

I don't know, I was following the old code: https://github.com/monal-im/Monal/blob/66e362863e8/Monal/Classes/xmpp.m#L4397
If I was writing the method from scratch I would've used a normal `int`, as the variable's value is unlikely to ever exceed 2^31 i.e. 2 billion

Monal/Classes/MLIQProcessor.m
lissine0: >  the whole handler will already be called inside a db write transaction

Where is this transaction created? I don't see it in the code.
Also, does this negate the other comments? (https://github.com/monal-im/Monal/pull/1465#discussion_r2299680692)

I tried removing the transaction and wrapping the `for` statement inside `dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{ })`
But the UI was became weird when fetching messages from the server.
Before:

https://github.com/user-attachments/assets/da9f872a-a994-44f7-8d77-fc255cd8a2be

After:

https://github.com/user-attachments/assets/a86b5daa-a176-4437-a7b5-60ae9d9cea8a


I don't really understand the rationale for the changes you're suggesting. So if you still want them, and if you don't mind, can you do the changes yourself? :)

Monal/Classes/ChatView.swift
lissine0: I could indent `.done` , `.catch` and `.finally` to the right if you think that improves readability.
Or I could use [firstly](https://github.com/mxcl/PromiseKit/blob/master/Documentation/GettingStarted.md#firstly)

tmolitor-stud-tu: No, your code is just fine, I just wanted to point that out, not request any changes :)

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: ah yes, could you change that to use `NSUInteger` or `int` then? My original code is quite old and back then I did not yet have a clear sense of which datatype to use when in objc.

Monal/Classes/MLIQProcessor.m
tmolitor-stud-tu: Well, the rationale is, that adding ~50 messages in one db transaction will block the db for writes. I don't know how much time it takes to parse and write all these message stanzas, but if it takes more than 500ms or 1s, then user might perceive this as a ui hang if the main thread tries to write something to db in this timeframe, too.

What's the actual timing for ~50 stanzas (text, geo, images)?

lissine0: In my testing, it takes about 2 seconds when fetching history from public channels

lissine0: Note that one of the commits in this PR changes the backscrolling mam queries to use a `max` of `kMonalBackscrollingMsgCount` instead of hardcoding `50`: https://github.com/lissine0/Monal/commit/0b1d56af82f3e5ec5d946c307795f2a741e57d91

You may want to drop that commit if `75` stanzas would block the UI for too long on macOS

tmolitor-stud-tu: Imho the weird ui behavior should not happen. The ui should not detect the new messages because all notifications emitted by the message processor should be blocked by https://github.com/monal-im/Monal/pull/1465/files#diff-66b7eebc413e2bae577e26d73798ea9b90376b6bb105c3c6b17a185d50e8ab99R288 and thrown away without being delivered by https://github.com/monal-im/Monal/pull/1465/files#diff-66b7eebc413e2bae577e26d73798ea9b90376b6bb105c3c6b17a185d50e8ab99R316

Maybe there is some bug in notification queue?

Monal/Classes/MLIQProcessor.m
lissine0: The Handler is already executed in a DB transaction.
And after some benchmarking, it turns out that processing 50 stanzas (on the Simulator) takes 58ms on average.
So, no substantial risk of blocking the main thread :)

Monal/Classes/MLIQProcessor.m
lissine0: > Maybe there is some bug in notification queue?

There is no bug in the notification queue. I had made a mistake in my (local) code when I was testing your suggestion.

> This is meant to use a dedicated db transaction per processed stanza to not block the db in a big write transaction of ~50 message stanzas. But the whole handler will already be called inside a db write transaction, so this nested transaction is a noop.

> --> You need to move the handling of collected stanzas to a background thread to free up the receive queue and its db transaction. You don't need to switch to the main thread when fullfilling the promise later on, the promise lib will already ensure that the promise handlers are scheduled on the main thread.

Since processing 50 stanzas takes ~58ms on average, it was decided to keep processing the stanzas on the receive queue thread (i.e. inside the Handler's transaction, without dispatching to another thread and processing each message in its own transaction). This has the benefit of preventing the potential MAM holes described in https://github.com/monal-im/Monal/pull/1465#discussion_r2299679462

tmolitor-stud-tu:
Is something in here still unresolved / wip or can this be merged now?

lissine0:
> Is something in here still unresolved / wip or can this be merged now?

This can be merged :)

lissine0:
About the CI, I don't know why it gets stuck. It used to work before the last time I force-pushed. And the only change in the last force-push was signing the commits so they have a "verified" badge.

Monal/Classes/ChatView.swift
```diff
@@ -410,13 +493,23 @@ struct ChatView: View {
             if message.isEqual(to: self.contact.obj) {
                 // Do not insert based on delay timestamp because that
                 // would make it possible to fake history entries.
-                messages.append(ChatViewMessage(message))
+                queuedNewMessages.append(ChatViewMessage(message))
```
tmolitor-stud-tu, line 492: You need locking for `queuedNewMessages` to not get race conditions between line 439 and here.

Monal/Classes/ChatView.swift
```diff
@@ -346,16 +431,14 @@ struct ChatView: View {
...
-                for msg in dbMessages {
-                    messages.append(ChatViewMessage(msg))
-                }
-            }
+            loadHistory()
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
+            // Insert new messages in batches to work around https://github.com/exyte/Chat/issues/223
+            let timer = Timer.scheduledTimer(withTimeInterval: 0.2, repeats: true) { _ in
+                messages.append(contentsOf: queuedNewMessages)
+                queuedNewMessages.removeAll(keepingCapacity: true)
```
tmolitor-stud-tu: You need locking for `queuedNewMessages` to not get race conditions between line 496 and here. **Unless**, of course, the timer runs in the same thread as line 496 (e.g. the main thread).

Monal/Classes/ChatView.swift
```diff
@@ -346,16 +431,14 @@ struct ChatView: View {
...
-            if messages.isEmpty {
-                let dbMessages = DataLayer.sharedInstance().messages(forContact:contact.contactJid, forAccount:contact.accountID) as! [MLMessage]
-                for msg in dbMessages {
-                    messages.append(ChatViewMessage(msg))
-                }
-            }
+            loadHistory()
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
+            // Insert new messages in batches to work around https://github.com/exyte/Chat/issues/223
+            let timer = Timer.scheduledTimer(withTimeInterval: 0.2, repeats: true) { _ in
```
tmolitor-stud-tu: First thought: You don't want to fire the timer every 0.2 seconds even if nothing happened.

You want a timer to be started once a new message notification came in. That timer should empty everything buffered so far and don't repeat. If then another new message notice comes in, a new timer would be started.

Second thought: I don't know what happens if the timer triggers after closing the ui, I hope that won't cause a crash.

tmolitor-stud-tu (changes requested):
Only the timer stuff and possible race condition left, everything else looks fine now :)

Monal/Classes/ChatView.swift
lissine0: There's no risk of race conditions, because all accesses to queuedMessages happen on `RunLoop.main`
The notification handling is scheduled on it explicitly, and the timer is scheduled on it implicitly because it's the current RunLoop when the timer is defined.

Tangential note:
The difference between scheduling tasks on `RunLoop.main` vs `DispatchQueue.main`, is that `RunLoop.main` doesn't run when the user is interacting with the UI e.g. when scrolling.

Monal/Classes/ChatView.swift
lissine0: > Second thought: I don't know what happens if the timer triggers after closing the ui, I hope that won't cause a crash.

I tested and it doesn't cause a crash.
But for peace of mind, I now invalidate the timer on view disappear.

tmolitor-stud-tu: What about the first thought?

You don't want to fire the timer every 0.2 seconds even if nothing happened.
You want a timer to be started once a new message notification came in. That timer should empty everything buffered so far and don't repeat. If then another new message notice comes in, a new timer would be started.

lissine0: I have already addressed that :)
See https://github.com/monal-im/Monal/compare/97ad54c750fdf7c0637c420a89a775102f14523c..706b818a4d096c89652e3b0b8bcb54d676e0b10a

tmolitor-stud-tu: Ah, my bad. I didn't check the code again.

tmolitor-stud-tu:
Why did you remove these two `kMonalContactRefresh` in https://github.com/monal-im/Monal/compare/706b818a4d096c89652e3b0b8bcb54d676e0b10a..13243a28e995be28295dd0139ebba48f4975d247

lissine0:
> Why did you remove these two `kMonalContactRefresh` in https://github.com/monal-im/Monal/compare/706b818a4d096c89652e3b0b8bcb54d676e0b10a..13243a28e995be28295dd0139ebba48f4975d247

I removed them because they seemingly cause a bug after your XEP-0492 implementation: sometimes updating the notification settings doesn't work (IIRC the button state would flip back to its previous state) 
I wanted to investigate the issue and not delay this PR in the meantime.

lissine0:
My final force push changed the db migration version, because I noticed that existing migrations never have a version that ends with .000
They always start with .001 for a new Monal version

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Interesting distinction between RunLoop.main and DispatchQueue.main :)

tmolitor-stud-tu:
> > Why did you remove these two `kMonalContactRefresh` in https://github.com/monal-im/Monal/compare/706b818a4d096c89652e3b0b8bcb54d676e0b10a..13243a28e995be28295dd0139ebba48f4975d247
> 
> I removed them because they seemingly cause a bug after your XEP-0492 implementation: sometimes updating the notification settings doesn't work (IIRC the button state would flip back to its previous state) I wanted to investigate the issue and not delay this PR in the meantime.

Well, thats interesting. But we need these notifications to update the symbols in the active chats view...

lissine0:
> Well, thats interesting. But we need these notifications to update the symbols in the active chats view...

I know, and I'm planning to re-add the commit in a new PR

PR #1470, opened 2025-08-28
Bubble up MLMessage changes to ChatViewMessage and from there to the view

### LINKED ISSUE #1462: Change ExyteChat `Message` to be a protocol
We want to be able to do something like this:
1. A view using an `@StateObject` to a class named `ChatViewMessage` and using its `text` property in the view body
2. The `ChatViewMessage` class being a child of the ExyteChat `open class Message: ObservableObject, Identifiable`
3. The `ChatViewMessage` class having a property `innerMessage` of type `ObservableKVOWrapper<MLMessage>`
4.  The `ChatViewMessage` class having an `@Published` computed property of type `String` named `text`, that always returns the `innerMessage.messageText` value. This property overwrites the `@Published public var text: String` property of the base class

Unfortunately, step 4 is a problem here, apparently you cannot overwrite a stored property with a computed one :(

**Solution:**
1. Check if the proof of concept above works while not deriving the `ChatViewMessage` class from anything. (Only check if the computed property properly forwards the change event of the `ObservableKVOWrapper` wrapped `MLMessage` object (use some timer to periodically change the text).
2. If that works, then:
    1. Update the ExyteChat to define a `protocol` named `MessageProtocol` that defines all the properties of the ExyteChat `Message` class/struct
    2. Let the ExyteChat `Message` derive from that protocol and change all mentions of `Message` in the ExyteChat codebase to be `MessageProtocol`  ones
    3. Then derive our implementation from that protocol, too, instead of deriving it from the `Message` class.

_I think introducing a new protocol is the way to go here, because it makes the original ExyteChat stuff more extensible but doesn't change any behavior. I think we can even upstream that change (together with making `Message` a class rather than a struct)._

lissine0:
Quoting #1462:

> We want to be able to do something like this:
> 
>1. A view using an `@StateObject` to a class named `ChatViewMessage` and using its `text` property in the view body
> 
>2. The `ChatViewMessage` class being a child of the ExyteChat `open class Message: ObservableObject, Identifiable`
> 
>3. The `ChatViewMessage` class having a property `innerMessage` of type `ObservableKVOWrapper<MLMessage>`
> 
>4. The `ChatViewMessage` class having an `@Published` computed property of type `String` named `text`, that always returns the `innerMessage.messageText` value. This property overwrites the `@Published public var text: String` property of the base class
> 
> 
> Unfortunately, step 4 is a problem here, apparently you cannot overwrite a stored property with a computed one :(

Actually, it turns out you can override a stored property with a computed one: you need to provide both a getter and a setter when overriding, and the setter can be empty.

So, I did the steps quoted above. But I ran into [The Nested Observables Problem in SwiftUI](https://developer.apple.com/forums/thread/761399):
> Basically, when a parent object holds a child object that is also an ObservableObject SwiftUI does not automatically update the view to reflect changes in the child object.

The solution was to either use the iOS17+ `@Observable` macro (both in Monal and ExyteChat), or to manually listen for changes to `MLMessage` in `ChatViewMessage` and call `objectWillChange.send()`

I chose the latter solution, as it's only a few lines of code, and we can easily migrate to `@Observable` later.

This fixes #1462 (without introducing a new protocol)

### DIFF
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -200,7 +200,7 @@ struct ChatView: View {
             guard let newMLMessage = MLXMPPManager.sharedInstance().sendMessageAndAddToHistory(message: draft.text, havingType: kMessageTypeText, toContact: self.contact.obj, isEncrypted: self.contact.isEncrypted, uploadInfo: nil) else {
                 return
             }
-            messages.append(ChatViewMessage(ObservableKVOWrapper(newMLMessage)))
+            messages.append(ChatViewMessage(newMLMessage))
         } messageBuilder: { message, viewModel, positionInUserGroup, positionInMessagesSection, positionInCommentsGroup, showContextMenuClosure, messageActionClosure, showAttachmentClosure in
             MessageView(message: (message as! ChatViewMessage), viewModel: viewModel, positionInUserGroup: positionInUserGroup, positionInMessagesSection: positionInMessagesSection)
         }
@@ -352,7 +352,7 @@ struct ChatView: View {
             if messages.isEmpty {
                 let dbMessages = DataLayer.sharedInstance().messages(forContact:contact.contactJid, forAccount:contact.accountID) as! [MLMessage]
                 for msg in dbMessages {
-                    messages.append(ChatViewMessage(ObservableKVOWrapper(msg)))
+                    messages.append(ChatViewMessage(msg))
                 }
             }
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
@@ -410,7 +410,7 @@ struct ChatView: View {
             if message.isEqual(to: self.contact.obj) {
                 // Do not insert based on delay timestamp because that
                 // would make it possible to fake history entries.
-                messages.append(ChatViewMessage(ObservableKVOWrapper(message)))
+                messages.append(ChatViewMessage(message))
             }
             ChatViewHelpers.refreshCounter(for: self.contact.obj)
         }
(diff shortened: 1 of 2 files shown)


Monal/Classes/ChatView.swift
```diff
@@ -421,13 +421,26 @@ struct ChatView: View {
...
+    private var subscriptions: Set<AnyCancellable> = Set()
+    override var text: String {
+        get {
+            return innerMessage.retracted ? "This message got retracted" : innerMessage.messageText
+        }
+        set {}
+    }
+    init(_ message: MLMessage) {
+        self.innerMessage = ObservableKVOWrapper(message)
         let user = ExyteChat.User(id: message.senderID, name: message.contactDisplayName, avatarURL: nil, isCurrentUser: !message.inbound)
```
tmolitor-stud-tu, line 434: Nothing to be changed in this PR, but we need `ExyteChat.User` to be dynamic, too. The `contactDisplayName` and avatar should also auto-update if the underlying MLContact changes.

We should add a `getContact` method to `MLMessage` if we don't have this already.

tmolitor-stud-tu (approved):
Looks good, sad that nested stuff doesn't work without manually subscribing, but well, it works good enough I guess :)

Monal/Classes/ChatView.swift
lissine0: > We should add a getContact method to MLMessage if we don't have this already.

We have a `contact` property already: https://github.com/monal-im/Monal/blob/66e362863e8/Monal/Classes/MLMessage.m#L315

tmolitor-stud-tu: Great :) then this is the one to be used to get the `MLContact` for `User` :)

PR #1471, opened 2025-08-30
Implement message actions and message status in the new ChatView

lissine0:
Support for replies or quotes will come later in the future, as part of another PR.


lissine0:
Continuing the discussion from this comment thread https://github.com/monal-im/Monal/pull/1465#discussion_r2286386368:

> the only thing I'd write in the main thread would be new sent messages because we want them in our db the instance we show them on screen [otherwise users could think the message was written, force close the app and loose the message if the db was heavily congested]

I agree that we want the newly sent message in the DB the moment it's shown on screen. But, we can delay displaying the new message until _after_ the DB write is finished.

In fact, that is the case already for message sending.

So, by running the DB write on a separate thread, we avoid blocking the UI: the user can scroll up and down; and they know their message wasn't sent yet, as it still didn't show up in a chat bubble (and it probably is still visible in the input view)

I think this same reasoning applies to other message actions like correction, retraction and local deletion, so I adhered to it in this PR: run the DB write on a separate thread, and only update the UI after the write is finished.

lissine0:
I ended up choosing an empty icon for `.sent` message status, like the UIKit ChatView, as it looks more natural in public channels, or when the contact has disabled sending read receipts and delivery receipts.

Monal/Classes/ChatView.swift
```diff
@@ -203,6 +280,81 @@ struct ChatView: View {
...
         } messageBuilder: { message, viewModel, positionInUserGroup, positionInMessagesSection, positionInCommentsGroup, showContextMenuClosure, messageActionClosure, showAttachmentClosure in
             MessageView(message: (message as! ChatViewMessage), viewModel: viewModel, positionInUserGroup: positionInUserGroup, positionInMessagesSection: positionInMessagesSection)
+        } messageMenuAction: { (action: MessageAction, defaultActionClosure, message) in
+            Task { @MainActor in
+                let mlMessage = (message as! ChatViewMessage).innerMessage.obj
+                switch action {
+                    case .copy:
+                        defaultActionClosure(message, .copy)
+                    case .edit:
+                        defaultActionClosure(message, .edit { editedText in
```
lissine0: This `.edit` closure is [defined](https://github.com/exyte/Chat/blob/114f42ab9e1ff65/Sources/ExyteChat/Views/MessageView/MessageMenu/MessageMenu%2BAction.swift#L25) as `Sendable` in ExyteChat.

This results in the following warning on the `sendMessage` call a couple of lines below :
```
Capture of 'mlMessage' with non-sendable type 'MLMessage' in a `@Sendable` closure; this is an error in the Swift 6 language mode
Class 'MLMessage' does not conform to the 'Sendable' protocol (monalxmpp.MLMessage)
```

We can't make the MLMessage class `Sendable` (that would require it to contain only stored properties that are immutable and sendable), so I silenced the warning by declaring it `@unchecked Sendable` in this source file.

lissine0:
Note that for messages with `.error` status, if the user clicks the status icon, a new message with the same contents is sent. (i.e. the failed message remains visible)
This is built-in to ExyteChat.

On the other hand, the "Resend" message action I added resends the same message (it reuses the same IDs)

This means we have two, slightly different ways to resend a failed message.
I don't think it's that big of a problem, but if you want, we can comment out these couple of lines in ExyteChat to disable the built-in way https://github.com/exyte/Chat/blob/114f42ab9e1ff6/Sources/ExyteChat/Views/MessageView/MessageView.swift#L129,L131

tmolitor-stud-tu:
> I agree that we want the newly sent message in the DB the moment it's shown on screen. But, we can delay displaying the new message until _after_ the DB write is finished.

Oh yes, good idea!

> So, by running the DB write on a separate thread, we avoid blocking the UI: the user can scroll up and down; and they know their message wasn't sent yet, as it still didn't show up in a chat bubble (and it probably is still visible in the input view)
> I think this same reasoning applies to other message actions like correction, retraction and local deletion, so I adhered to it in this PR: run the DB write on a separate thread, and only update the UI after the write is finished.

Yes, great! :)

tmolitor-stud-tu:
> Note that for messages with `.error` status, if the user clicks the status icon, a new message with the same contents is sent. (i.e. the failed message remains visible) This is built-in to ExyteChat.
> 
> On the other hand, the "Resend" message action I added resends the same message (it reuses the same IDs)
> 
> This means we have two, slightly different ways to resend a failed message. I don't think it's that big of a problem, but if you want, we can comment out these couple of lines in ExyteChat to disable the built-in way https://github.com/exyte/Chat/blob/114f42ab9e1ff6/Sources/ExyteChat/Views/MessageView/MessageView.swift#L129,L131

Why did you add a "resend" action in the first place?

I think removing the old message having the error when resending using the status icon would be better, if that's easily achievable.

Monal/Classes/ChatView.swift
tmolitor-stud-tu: One of the reasons why I din't like the direction Swift is currently moving to, it makes easy things much much harder or even impossible, even if the programmer knows what they are doing :(
Same goes for the stupid actor stuff that should ensure that ui things run on the main thread, but not detecting that we are already on the main thread.

That said, I'm not sure why the closure was defined as `Sendable` in the first place. @matthewrfennell do you have any insights here?

tmolitor-stud-tu:
> I ended up choosing an empty icon for `.sent` message status, like the UIKit ChatView, as it looks more natural in public channels, or when the contact has disabled sending read receipts and delivery receipts.

Sounds sensible :)

lissine0:
> Why did you add a "resend" action in the first place?

I did so because I thought the error status icon was too small, and that it's not obvious that it's a clickable button. (Note however that some of the area neighboring the icon to the top, right, and top-right is also clickable)

Here's a screenshot
screenshot

lissine0:
> I think removing the old message having the error when resending using the status icon would be better, if that's easily achievable.

That may be easily achievable on the ExyteChat side, but the message won't be removed from the DB, meaning it'll be shown the next time the ChatView is entered.

A proper solution is to modify ExyteChat, allowing library users to provide a custom closure that is triggered when the error status icon is clicked.

If you prefer clicking the error status icon over using the message actions, I can try to contribute this change to ExyteChat. (A custom make it possible to also remove the "Show Error" message action.)

tmolitor-stud-tu:
> A proper solution is to modify ExyteChat, allowing library users to provide a custom closure that is triggered when the error status icon is clicked.
> 
> If you prefer clicking the error status icon over using the message actions, I can try to contribute this change to ExyteChat. (A custom make it possible to also remove the "Show Error" message action.)

yes please :)

tmolitor-stud-tu:
Rethinking this. Given that the red status icon is that small, I think I prefer to just have the message action to display the error and be asked for resending (the confirmation prompt way we already discussed in chat) and let the icon only be a simple non-tappable icon.

I hope you didn't invest too much time into adding this closure calling stuff to exytechat :/

lissine0:
Here's how the UI for resending messages is currently implemented:

https://github.com/user-attachments/assets/edec35b7-476a-4d13-9d05-66a2d978c654


Also, note that I tested the localization extraction script, and the following string is extracted successfully:
https://github.com/lissine0/Monal/blob/4ee957b4cc81c64c78c01433fb77c6f03fa2ed43/Monal/Classes/ChatView.swift#L334

tmolitor-stud-tu:
Sure, I've updated the extraction script a few months ago to really build the project, so everything taking a `LocalizedStringKey` gets properly extracted (even when used as parameter for our own functions etc.) :)

Monal/Classes/ChatView.swift
```diff
@@ -80,6 +80,8 @@ extension MonalView {
 }
 */
 
+// This silences a Swift concurrency warning (in the .edit defaultActionClosure)
+extension MLMessage: @unchecked @retroactive Sendable {}
```
tmolitor-stud-tu: This belongs into `SwiftHelpers.swift` (we have some more `@retroactive` stuff at the top of that file)

Monal/Classes/ChatView.swift
tmolitor-stud-tu: @matthewrfennell Do you know why this closure was defined as sendable?
Would it be possible to remove that sendable requirement upstream? That would spare us this dirty `MLMessage` as `Sendable` hack

Monal/Classes/ChatView.swift
```diff
@@ -284,13 +275,96 @@ struct ChatView: View {
...
+                        defaultActionClosure(message, .edit { editedText in
+                            Task { @MainActor in
+                                self.account.sendMessage(editedText, to: self.contact.obj, isEncrypted: self.contact.isEncrypted || mlMessage.encrypted, isUpload: false, andMessageId: UUID().uuidString, withLMCId: mlMessage.messageId)
+
+                                // Don't block the main thread while writing to the DB
+                                await Task.detached(priority: .userInitiated) {
+                                    DataLayer.sharedInstance().updateMessageHistory(mlMessage.messageDBId, withText: editedText)
+                                }.value
+
+                                (MLNotificationQueue.currentQueue() as AnyObject).post(
```
tmolitor-stud-tu: Why is this explicit cast to `AnyObject` needed?

Would it be possible to abstract this explicit cast to `AnyObject` away by somehow write a swift extension for `MLNotificationQueue` overwriting / redeclaring the `currentQueue` method so that no explicit cast is needed anymore?

Monal/Classes/ChatView.swift
```diff
@@ -284,13 +275,96 @@ struct ChatView: View {
...
+                                               "LMCReplaced": true,
+                                               "correctedText": editedText,
+                                              ]
+                                )
+                            }
+                        })
+                    case .retract:
+                        self.account.retractMessage(mlMessage)
+                        await Task.detached(priority: .userInitiated) {
+                            DataLayer.sharedInstance().retractMessageHistory(mlMessage.messageDBId)
```
tmolitor-stud-tu: I think this should become a part of the `retractMessage` method in out `xmpp` class. It should be wrapped in a async receive queue block doing the call to `[self send:messageNode]` and the call to our DataLayer's `retractMessageHistory`.
That way it will become (more) transactional (not fully, because we don't freeze the send queue until the block completes, but we could make this fully transactional later on by introducing some `sendWithAdditionalBlock:stanza:andSmacks:` method doing everything atomically in a single db transaction AND frezing the send queue until the transaction has been committed to disk).

Monal/Classes/ChatView.swift
```diff
@@ -284,13 +275,96 @@ struct ChatView: View {
...
+                        }.value
+
+                        (MLNotificationQueue.currentQueue() as AnyObject).post(
+                            name: Notification.Name(kMonalDeletedMessageNotice),
+                            object: self.account,
+                            userInfo: ["message": mlMessage, "contact": self.contact.obj]
+                        )
+                    case .moderate:
+                        messageToModerate = mlMessage
+                        showModeratingMessageAlert = true
```
tmolitor-stud-tu: Use the `Binding.optionalMappedToBool` to get rid of the `showModeratingMessageAlert` and show the alert if the `messageToModerate` is not `nil`.

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Maybe we can also introduce a new `postToCurrentMLNotificationQueue` function wrapping all of this, if the above is not possible. But I'd avoid this indirection if possible.

Monal/Classes/ChatView.swift
```diff
@@ -284,13 +275,96 @@ struct ChatView: View {
...
+                                    Text("Retry"),
+                                    action: {
+                                        Task { @MainActor in
+                                            await Task.detached(priority: .userInitiated) {
+                                                DataLayer.sharedInstance().clearError(ofMessageId: mlMessage.messageId)
+                                            }.value
+
+                                            mlMessage.errorType = ""
+                                            mlMessage.errorReason = ""
+                                            self.account.sendMessage(mlMessage.messageText, to: self.contact.obj, isEncrypted: self.contact.isEncrypted, isUpload: false, andMessageId: mlMessage.messageId)
```
tmolitor-stud-tu: What about resending a fileupload? Shouldn't `isUpload` be dynamic and mirror the value of the original `MLMessage`?

tmolitor-stud-tu: Same for the `isEncrypted`, shouldn't that be mirrored, too?

Monal/Classes/ChatView.swift
```diff
@@ -311,6 +385,20 @@ struct ChatView: View {
                 }
             }))
         }
+        .alert("Moderating message", isPresented: $showModeratingMessageAlert, actions: {
```
tmolitor-stud-tu: Maybe `richAlert` would be a better fit than `alert`? I don't know how both would look and what would look better.

Monal/Classes/ChatView.swift
```diff
@@ -522,7 +596,30 @@ class ChatViewMessage: ExyteChat.Message {
     private var subscriptions: Set<AnyCancellable> = Set()
     override var text: String {
         get {
-            return innerMessage.retracted ? "This message got retracted" : innerMessage.messageText
+            return innerMessage.retracted ? NSLocalizedString("This message got retracted", comment: "") : innerMessage.messageText
```
tmolitor-stud-tu, line 598: Good catch! :)

tmolitor-stud-tu (commented):
Only small comments

Monal/Classes/ChatView.swift
lissine0: > Do you know why this closure was defined as sendable?
Would it be possible to remove that sendable requirement upstream? That would spare us this dirty MLMessage as Sendable hack

With some code changes in the PR, we don't get the warning anymore

<details>
<summary>
details
</summary>

Since we removed a lot of the `await Task.detached` usages, only two remain now.
So, rather than wrapping the entire switch statement inside a `Task { @MainActor in`, I'm wrapping only the two switch cases (separately).
This removes most of the `Sendable` warnings. Only one remains.

The remaining warning is fixed by passing the `MLMessage` property we need (`messageDBId`, which is `Sendable` because it's a `NSNumber`) instead of the entire `MLMessage` object:
```
let messageDBId = mlMessage.messageDBId
/* ... */
                await Task.detached(priority: .userInitiated) {
                    DataLayer.sharedInstance().updateMessageHistory(messageDBId, withText: editedText)
                }.value
```
</details>

Monal/Classes/ChatView.swift
lissine0: It turns out, when I was casting to `AnyObject`, I was using the `post` method defined on `NotificationCenter`: https://developer.apple.com/documentation/foundation/notificationcenter/post(name:object:userinfo:)

I fixed this by changing the return type of `MLNotificationQueue`'s `currentQueue` method from `id` to `MLNotificationQueue*`
I also used custom swift names for the `post` methods so they match the swift names for `post` methods defined on `NotificationCenter`

Monal/Classes/ChatView.swift
lissine0: Yes, you're right. These are minor bugs in the old chatView that I copied to the new one.

For the `isEncrypted`, we could alternatively do like message corrections: 
```
encrypted: mlMessage.encrypted || contact.isEncrypted
```
i.e. pick the stronger encryption setting between the originally used setting and the one used now.

Or, we could always use the original setting. That's fine too.

Monal/Classes/ChatView.swift
lissine0: I didn't try the `richAlert` yet

Monal/Classes/ChatView.swift
tmolitor-stud-tu: The proper type would be `instancetype`. That automatically resolves to whatever type the class has (in this case `MLNotificationQueue*` and even gets properly resolved in derived classes etc.

tmolitor-stud-tu: I like the swift renaming :)

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Since this is half-transactional now, the notification MUST be delayed until the transaction is committed.
So this belongs into the `xmpp*`'s `retract` method, too:

```
MLNotificationQueue.current().post(
    name: Notification.Name(kMonalDeletedMessageNotice),
    object: self.account,
    userInfo: ["message": mlMessage, "contact": self.contact.obj]
)
```

Monal/Classes/ChatView.swift
tmolitor-stud-tu: I think `encrypted: mlMessage.encrypted || contact.isEncrypted` is better/safer :)

Monal/Classes/ChatView.swift
lissine0: This is how it looks like with `alert`:
moderation alert

Though I admit it might look a bit ugly on macOS:
alert on macOS (1)
alert on macOS (2)

I didn't try `richAlert` because it will require a bit of refactoring to use `.optionalMappedToBool()` with it:
`richAlert` expects `isPresented` to be a `Binding<T?>` but `optionalMappedToBool` returns a `Binding<Bool>`, and the Swift compiler can't convert automatically.

The error is:
```
Cannot convert value of type 'Binding<Bool>' to expected argument type 'Binding<()?>'
Arguments to generic parameter 'Value' ('Bool' and '()?') are expected to be equal
```

You can check the invite creation view to get a feel for `richAlert`: its background is blurred, it has a small shadow (try it in dark mode!), and it would look the same on iOS and macOS.

Monal/Classes/ChatView.swift
```diff
@@ -203,6 +272,93 @@ struct ChatView: View {
...
+
+                                        mlMessage.errorType = ""
+                                        mlMessage.errorReason = ""
+                                        let isUpload = mlMessage.messageType == kMessageTypeFiletransfer
+                                        let isEncrypted = mlMessage.encrypted || self.contact.isEncrypted
+                                        self.account.sendMessage(mlMessage.messageText,
+                                                                 to: self.contact.obj,
+                                                                 isEncrypted: isEncrypted,
+                                                                 isUpload: isUpload,
+                                                                 andMessageId: mlMessage.messageId)
```
lissine0, line 346: This is how Xcode automatically indents multiline function arguments:
```swift
self.account.sendMessage(mlMessage.messageText,
                         to: self.contact.obj,
                         isEncrypted: isEncrypted,
                         isUpload: isUpload,
                         andMessageId: mlMessage.messageId)
```
But if you prefer we could use 4-space indentation for the argument list instead:
```
self.account.sendMessage(editedText,
    to: self.contact.obj,
    isEncrypted: self.contact.isEncrypted || mlMessage.encrypted,
    isUpload: false,
    andMessageId: UUID().uuidString,
    withLMCId: mlMessage.messageId
)
```

Monal/Classes/ChatView.swift
tmolitor-stud-tu: No, you don't need the `optionalMappedToBool` at all. The rich alert will do this internally already :) You can just use the value directly and the alert will pop up if it is not `nil`.

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Well, I somewhat lean to the 4-space version. But we for sure have the XCode version in our codebase, too.

tmolitor-stud-tu: Nearly all code written by me uses the 4-space version, though.

tmolitor-stud-tu (approved):

Issue #1423, opened 2025-05-04
[Bug]: repeated notifications that Monal failed to delete bookmark for deleted group/channel


Jamagrammatic:
### What Monal Release channel are you using?

Beta

### iOS system version (if applicable)

18.1.1

### Monal version

6.4.11 (1038)

### Used XMPP server (domain)

Snikket.org

### Which XMPP-Server software are you using?

other

### How many accounts are you using in Monal?

1

### What happened?

Opening Monal results in repeated notifications for a deleted group/channel saying “_Failed to remove bookmark for group/channel xxxx item not found_”

### Anything else?

Does not happen in other clients using same account including Snikket iOS, Siskin, Conversations, Dino. Tried deleting and re-adding account as well as deleting and reinstalling app but issue persists. 

### FAQ

- [x] I have [checked the FAQ](https://github.com/monal-im/Monal/wiki/FAQ---Frequently-Asked-Questions) for my issue

### Considerations for XMPP users

- [x] I have checked if my issue can be solved with [Considerations for XMPP users](https://github.com/monal-im/Monal/wiki/Considerations-for-XMPP-users) and [Considerations for XMPP server admins](https://github.com/monal-im/Monal/wiki/Considerations-for-XMPP-server-admins)

### XEP-Check

- [x] I have checked that at least XEP-198, XEP-0280, XEP-0352, XEP-0357, XEP-0313 and XEP-0163 are activated on my server and shown as 'green' under Settings --> Account--> (i) in advanced settings

### Notifications-Menu

- [x] I have checked that all checkmarks are present under Settings --> Notifications

tmolitor-stud-tu:
To debug this, I need a logfile sent via mail to info@monal-im.org
See here on how to accomplish this: https://github.com/monal-im/Monal/wiki/Introduction-to-Monal-Logging

PR #1474, opened 2025-09-14
Backport the smacks rework to Stable 6.4

lissine0:
And while at it, also cherry-pick the fix for #1423


tmolitor-stud-tu:
Well, for #1423 it would have been cool to get a logfile because repeated tries to delete a bookmark should never happen.

The only case it happened several times in a row was when receiving several prosody errors then reflected by an ejabberd muc. Maybe the issue reported in #1423 was the same, but without a logfile, we'll never know.

PR #1480, opened 2025-09-30
Fetch MUC participant avatars

### LINKED ISSUE #699: Show users' avatars in groupchats
Most of clients (like Conversations, Psi, Dino, Pidgin, etc) shows recepients avatars directly in messages list (or in a separate members list, which is not applicable for Monal). It will be very useful for distinguishing between chat participants

lissine0:
Add support for fetching MUC participant avatars, using vcard-temp (XEP-0153)
And display the avatars in the ChatView, and in ChannelMemberList.

Fixes #699

### DIFF
--- a/Monal/Classes/MLImageManager.m
+++ b/Monal/Classes/MLImageManager.m
@@ -194,6 +194,26 @@
         DDLogError(@"Got error while trying to delete all contact avatar files: %@", error);
 }
 
+-(void) deleteAvatarsOfRoom:(MLContact*) roomContact
+{
+    NSFileManager* fileManager = [NSFileManager defaultManager];
+
+    // delete the room avatar
+    //documents directory/buddyicons/accountID/contact
+    NSString* filename = [self fileNameforContact:roomContact];
+    NSString* writablePath = [self.documentsDirectory stringByAppendingPathComponent:@"buddyicons"];
+    writablePath = [writablePath stringByAppendingPathComponent:roomContact.accountID.stringValue];
+    writablePath = [writablePath stringByAppendingPathComponent:filename];
+    [fileManager removeItemAtPath:writablePath error:nil];
+
+    // delete participant avatars
+    //documents directory/muc_participant_avatars/accountID/room/
+    writablePath = [self.documentsDirectory stringByAppendingPathComponent:@"muc_participant_avatars"];
+    writablePath = [writablePath stringByAppendingPathComponent:roomContact.accountID.stringValue];
+    writablePath = [writablePath stringByAppendingPathComponent:roomContact.contactJid];
+    [fileManager removeItemAtPath:writablePath error:nil];
+}
+
 #pragma mark chat bubbles
 
 -(UIImage*) inboundImage
--- a/Monal/Classes/MLImageManager.m
+++ b/Monal/Classes/MLImageManager.m
@@ -214,6 +214,23 @@
     [fileManager removeItemAtPath:writablePath error:nil];
 }
 
+-(void) deleteAvatarsOfAccount:(NSNumber*) accountID
+{
+    NSFileManager* fileManager = [NSFileManager defaultManager];
+
+    // delete the contact avatar
+    //documents directory/buddyicons/accountID/
+    NSString* writablePath = [self.documentsDirectory stringByAppendingPathComponent:@"buddyicons"];
+    writablePath = [writablePath stringByAppendingPathComponent:accountID.stringValue];
+    [fileManager removeItemAtPath:writablePath error:nil];
+
+    // delete the room participant avatars
+    //documents directory/muc_participant_avatars/accountID/
+    writablePath = [self.documentsDirectory stringByAppendingPathComponent:@"muc_participant_avatars"];
+    writablePath = [writablePath stringByAppendingPathComponent:accountID.stringValue];
+    [fileManager removeItemAtPath:writablePath error:nil];
+}
+
 #pragma mark chat bubbles
 
 -(UIImage*) inboundImage
--- a/Monal/Classes/MLImageManager.m
+++ b/Monal/Classes/MLImageManager.m
@@ -113,6 +113,11 @@
     [self resetCachedBackgroundImageForContact:[MLContact createContactFromJid:contact andAccountID:accountID]];
 }
 
+-(void) purgeCacheForOccupant:(NSString*) occupantId inRoom:(NSString*) room
+{
+    [self.iconCache removeObjectForKey:[NSString stringWithFormat:@"%@_%@", room, occupantId]];
+}
+
 -(void) cleanupHashes
 {
     NSFileManager* fileManager = [NSFileManager defaultManager];
@@ -141,17 +146,52 @@
             if(error)
                 DDLogError(@"Error deleting orphan avatar file: %@", error);
         }
+
+        //clean up orphan hashes and orphan avatar files of muc participants
+        if(contact.isMuc)
+        {
+            NSArray* participants = [[DataLayer sharedInstance] getMembersAndParticipantsOfMuc:contact.contactJid forAccountID:contact.accountID];
+            //documents directory/muc_participant_avatars/accountID/room/occupantId
+            NSString* rootPath = [self.documentsDirectory stringByAppendingPathComponent:@"muc_participant_avatars"];
+            rootPath = [rootPath stringByAppendingPathComponent:contact.accountID.stringValue];
+            rootPath = [rootPath stringByAppendingPathComponent:contact.contactJid];
+            for(NSDictionary* participant in participants)
+            {
+                NSString* occupantId = participant[@"occupant_id"];
+                if(occupantId == nil || [occupantId isEqualToString:@""])
+                    continue;
+
+                writablePath = [rootPath stringByAppendingPathComponent:[self fileNameForOccupant:occupantId]];
+                hasHash = participant[@"iconhash"] != nil && ![@"" isEqualToString:participant[@"iconhash"]];
+
+                if(hasHash && ![fileManager isReadableFileAtPath:writablePath])
+                {
+                    DDLogDebug(@"Deleting orphan hash '%@' of muc participant: %@/%@", participant[@"iconhash"], participant[@"room"], participant[@"room_nick"]);
+                    //delete avatar hash from db if the file containing our image data vanished
+                    [[DataLayer sharedInstance] clearAvatarHashForOccupant:occupantId inRoom:participant[@"room"] forAccount:contact.accountID];
+                }
+
+                if(!hasHash && [fileManager isReadableFileAtPath:writablePath])
+                {
+                    DDLogDebug(@"Deleting orphan avatar file '%@' of muc participant: %@/%@", writablePath, participant[@"room"], participant[@"room_nick"]);
+                    NSError* error;
+                    [fileManager removeItemAtPath:writablePath error:&error];
+                    if(error)
+                        DDLogError(@"Error deleting orphan avatar file: %@", error);
+                }
+            }
+        }
     }
 }
 
--(void) removeAllIcons
+-(void) removeAllContactIcons
 {
     NSError* error;
     NSFileManager* fileManager = [NSFileManager defaultManager];
     NSString* writablePath = [self.documentsDirectory stringByAppendingPathComponent:@"buddyicons"];
     [fileManager removeItemAtPath:writablePath error:&error];
     if(error)
-        DDLogError(@"Got error while trying to delete all avatar files: %@", error);
+        DDLogError(@"Got error while trying to delete all contact avatar files: %@", error);
 }
 
 #pragma mark chat bubbles
@@ -175,19 +215,11 @@
 
 #pragma mark user icons
 
--(UIImage*) generateDummyIconForContact:(MLContact*) contact
+-(UIImage*) generatePlaceholderAvatarForString:(NSString*) inputString withInitial:(NSString*) initialCharacter
 {
-    NSString* contactLetter;
+    MLAssert(initialCharacter != nil && initialCharacter.length == 1, @"initialCharacter string does not represent a character");
 
-    if(contact.isSelf)
-    {
-        xmpp* account = contact.account;
-        contactLetter = [[[MLContact ownDisplayNameForAccount:account] substringToIndex:1] uppercaseString];
-    }
-    else
-        contactLetter = [[[contact contactDisplayName] substringToIndex:1] uppercaseString];
-
-    UIColor* background = [HelperTools generateColorFromJid:contact.contactJid];
+    UIColor* background = [HelperTools generateColorFromString:inputString];
     UIColor* foreground = [UIColor blackColor];
     if(![background isLightColor])
         foreground = [UIColor whiteColor];
@@ -210,18 +242,43 @@
             NSForegroundColorAttributeName: foreground,
             NSParagraphStyleAttributeName: paragraphStyle
         };
-        CGSize textSize = [contactLetter sizeWithAttributes:attributes];
+        CGSize textSize = [initialCharacter sizeWithAttributes:attributes];
         CGRect textRect = CGRectMake(floorf((float)(renderer.format.bounds.size.width - textSize.width) / 2),
                                     floorf((float)(renderer.format.bounds.size.height - textSize.height) / 2),
                                     textSize.width,
                                     textSize.height);
-        [contactLetter drawInRect:textRect withAttributes:attributes];
+        [initialCharacter drawInRect:textRect withAttributes:attributes];
     }];
+}
+
+-(UIImage*) generateDummyIconForContact:(MLContact*) contact
+{
+    NSString* initial = [[contact.contactDisplayNameWithoutSelfnotesPrefix substringToIndex:1] uppercaseString];
+    return [self generatePlaceholderAvatarForString:contact.contactJid withInitial:initial];
+}
+
+-(UIImage*) generatePlaceholderAvatarForNick:(NSString*) nick
+{
+    NSString* initial = [[nick substringToIndex:1] uppercaseString];
+    return [self generatePlaceholderAvatarForString:nick withInitial:initial];
+}
+
+-(UIImage*) generatePlaceholderAvatarForOccupantId:(NSString*) occupantId andNick:(NSString*) nick
+{
+    NSString* initial = [[nick substringToIndex:1] uppercaseString];
+    return [self generatePlaceholderAvatarForString:occupantId withInitial:initial];
 }
 
 -(NSString*) fileNameforContact:(MLContact*) contact
 {
     return [NSString stringWithFormat:@"%@_%@.png", contact.accountID.stringValue, [contact.contactJid lowercaseString]];;
+}
+
+-(NSString*) fileNameForOccupant:(NSString*) occupantId
+{
+    // Replace forward slashes with "__" because a forward slash is a path separator
+    NSString* escapedOccupantId = [occupantId stringByReplacingOccurrencesOfString:@"/" withString:@"__"];
+    return [NSString stringWithFormat:@"%@.png", escapedOccupantId];
 }
 
 -(NSString*) fileNameforThumbnailOfMessage:(MLMessage*) message
@@ -293,7 +350,41 @@
     
     //remove from cache if its there
     [self.iconCache removeObjectForKey:[NSString stringWithFormat:@"%@_%@", contact.accountID, contact]];
-    
+}
+
+-(void) setAvatarForOccupant:(NSString*) occupantId inRoom:(NSString*) room forAccount:(NSNumber*) accountID WithData:(NSData* _Nullable) data
+{
+    MLAssert(occupantId != nil && ![occupantId isEqualToString:@""], @"The occupantId can't be empty when saving MUC participant avatars");
+
+    //documents directory/muc_participant_avatars/accountID/room/occupantId
+    NSString* filename = [self fileNameForOccupant:occupantId];
+    NSFileManager* fileManager = [NSFileManager defaultManager];
+
+    NSString* writablePath = [self.documentsDirectory stringByAppendingPathComponent:@"muc_participant_avatars"];
+    writablePath = [writablePath stringByAppendingPathComponent:accountID.stringValue];
+    writablePath = [writablePath stringByAppendingPathComponent:room];
+
+    NSError* error;
+    [fileManager createDirectoryAtPath:writablePath withIntermediateDirectories:YES attributes:nil error:&error];
+    [HelperTools configureFileProtectionFor:writablePath];
+    writablePath = [writablePath stringByAppendingPathComponent:filename];
+
+    if([fileManager fileExistsAtPath:writablePath])
+        [fileManager removeItemAtPath:writablePath error:nil];
+
+    if(data)
+    {
+        if([data writeToFile:writablePath atomically:NO])
+        {
+            [HelperTools configureFileProtectionFor:writablePath];
+            DDLogVerbose(@"wrote image to file: %@", writablePath);
+        }
+        else
+            DDLogError(@"failed to write image to file: %@", writablePath);
+    }
+
+    //remove from cache if its there
+    [self purgeCacheForOccupant:occupantId inRoom:room];
 }
 
 -(BOOL) hasIconForContact:(MLContact*) contact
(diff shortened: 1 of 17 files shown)


lissine0:
Here are some points for discussion:
- When first joining a room, mam queries don't work until one of the following two things happens:
  - the avatars finish being fetched
  - 60 seconds have passed. At which point the following [error](https://github.com/monal-im/Monal/blob/413b788e8a6d3a7e1dfc6ffa8c40c269975b913f/Monal/Classes/MLMucProcessor.m#L1555) is shown: `Failed to query new messages for Group/ Channel (stanzaid) room@conference.domain.tld: remote-server-timeout (No response in 60 seconds)`.
 Avatar fetching also stops at this point, or slows down significantly.
- Currently, if some participants are present in multiple rooms, their avatars are fetched once for every room.
It would be nice to optimize this case: before sending the iq, we could check if the avatar hash is known in other rooms. If so, we simply copy the file. otherwise we proceed with sending the iq.
- The avatar file storage on disk does not distinguish between different accounts.
i.e. if two accounts are in the same room, they share the same avatar files.
(the avatars are stored in `documents directory/muc_participant_avatars/room/nick`)
I think this can be extended to the database as well, by dropping the `account_id` column from the `muc_participants` table.

- Minor: there's a [comment](https://github.com/monal-im/Monal/blob/413b788e8a6d3a7e1dfc6ffa8c40c269975b913f/Monal/Classes/DataLayer.m#L876) that says `//hashes should never be nil`, should we enforce that on the DB side? (by making the column NOT NULL, and setting an empty string as a default value)
- ejabberd has a custom muc configuration form field that lets you disable iqs in the room:
```xml
<field var="allow_query_users" type="boolean" label="Allow users to query other users">
    <value>1</value>
</field>
 ```
Should we skip fetching avatars if we detect that this config option is set to `false`? If we don't, then we'll be sending a pointless iq every time we get a presence with a hash in it, which will return an error every time.
But this configuration is rarely used I think.

File: Monal/Classes/MLImageManager.m [hunk 1]
```diff
@@ -113,6 +113,11 @@ -(void) purgeCacheForContact:(NSString*) contact andAccount:(NSNumber*) accountI
     [self resetCachedBackgroundImageForContact:[MLContact createContactFromJid:contact andAccountID:accountID]];
 }
 
+-(void) purgeCacheForNick:(NSString*) nick inRoom:(NSString*) room
```
tmolitor-stud-tu: I think we should never use the nick to identify `channel` (anon muc) users, but always use the occupant-id. And use the jid for `groups` (non-anon mucs).

tmolitor-stud-tu:
See my answers inline.

> * When first joining a room, mam queries don't work until one of the following two things happens:
>   
>   * the avatars finish being fetched
>   * 60 seconds have passed. At which point the following [error](https://github.com/monal-im/Monal/blob/413b788e8a6d3a7e1dfc6ffa8c40c269975b913f/Monal/Classes/MLMucProcessor.m#L1555) is shown: `Failed to query new messages for Group/ Channel (stanzaid) room@conference.domain.tld: remote-server-timeout (No response in 60 seconds)`.
>     Avatar fetching also stops at this point, or slows down significantly.

Why do avatar fetches block mam? I don't think this is something we want.

> * Currently, if some participants are present in multiple rooms, their avatars are fetched once for every room.
>   It would be nice to optimize this case: before sending the iq, we could check if the avatar hash is known in other rooms. If so, we simply copy the file. otherwise we proceed with sending the iq.

Well, since different mucs could have different policies for direct communication, I think that optimization would cause more confusion than do good (its hard to understand for ordinary users, why a muc shows only avatars of some users and not others).

> * The avatar file storage on disk does not distinguish between different accounts.
>   i.e. if two accounts are in the same room, they share the same avatar files.
>   (the avatars are stored in `documents directory/muc_participant_avatars/room/nick`)
>   I think this can be extended to the database as well, by dropping the `account_id` column from the `muc_participants` table.

No, it should even be the other way round: we need to differentiate by account. My own affiliation might influence, if I'm able to retrieve the avatar or even see the jid of participants. Since the avatar hash of every participant is stored in the db, using something like this: `documents directory/muc_participant_avatars/accountId/roomJid/avatarHash`

> * Minor: there's a [comment](https://github.com/monal-im/Monal/blob/413b788e8a6d3a7e1dfc6ffa8c40c269975b913f/Monal/Classes/DataLayer.m#L876) that says `//hashes should never be nil`, should we enforce that on the DB side? (by making the column NOT NULL, and setting an empty string as a default value)

No, no. That simply means that we should return an empty string rather than nil to indicate a missing hash. Having `NULL` in the database it still beneficial, I think. Only the `getAvatarHashForContact:andAccount:` should always return a string to ease the code using that return value (e.g. calling `isEqualToString:` on it etc.).

> * ejabberd has a custom muc configuration form field that lets you disable iqs in the room:
> 
> ```
> <field var="allow_query_users" type="boolean" label="Allow users to query other users">
>     <value>1</value>
> </field>
> ```
> 
> Should we skip fetching avatars if we detect that this config option is set to `false`? If we don't, then we'll be sending a pointless iq every time we get a presence with a hash in it, which will return an error every time. But this configuration is rarely used I think.

Yes, I think that would be beneficial, even if that is rarely used.

tmolitor-stud-tu (changes requested):
I did only a quick scroll-through

lissine0:
> See my answers inline.
> 
> > * When first joining a room, mam queries don't work until one of the following two things happens:
> >   
> >   * the avatars finish being fetched
> >   * 60 seconds have passed. At which point the following [error](https://github.com/monal-im/Monal/blob/413b788e8a6d3a7e1dfc6ffa8c40c269975b913f/Monal/Classes/MLMucProcessor.m#L1555) is shown: `Failed to query new messages for Group/ Channel (stanzaid) room@conference.domain.tld: remote-server-timeout (No response in 60 seconds)`.
> >     Avatar fetching also stops at this point, or slows down significantly.
> 
> Why do avatar fetches block mam? I don't think this is something we want.

This is a bug, that I was hoping you'd have some insights about. my guess is it's somehow related to the sendQueue having too many iqs, because the mam query seems to be added to the sendQueue after the avatar queries. I'll investigate the issue more.


> > * Currently, if some participants are present in multiple rooms, their avatars are fetched once for every room.
> >   It would be nice to optimize this case: before sending the iq, we could check if the avatar hash is known in other rooms. If so, we simply copy the file. otherwise we proceed with sending the iq.
> 
> Well, since different mucs could have different policies for direct communication, I think that optimization would cause more confusion than do good (its hard to understand for ordinary users, why a muc shows only avatars of some users and not others).

If by policies for direct communication you mean blocking iqs, my last suggestion already covers this point. we can detect that policy and avoid copying the avatar files for MUCs that block iqs

> > * The avatar file storage on disk does not distinguish between different accounts.
> >   i.e. if two accounts are in the same room, they share the same avatar files.
> >   (the avatars are stored in `documents directory/muc_participant_avatars/room/nick`)
> >   I think this can be extended to the database as well, by dropping the `account_id` column from the `muc_participants` table.
> 
> No, it should even be the other way round: we need to differentiate by account. My own affiliation might influence, if I'm able to retrieve the avatar or even see the jid of participants. Since the avatar hash of every participant is stored in the db, using something like this: `documents directory/muc_participant_avatars/accountId/roomJid/avatarHash`

I get that my affiliation influences if I can see the jid of participants. But I don't get how it influences if I'm able to retrieve avatars.

tmolitor-stud-tu:
> If by policies for direct communication you mean blocking iqs, my last suggestion already covers this point. we can detect that policy and avoid copying the avatar files for MUCs that block iqs

yes and no. allowing iqs between participants isn't mandated by the spec and that ejabberd exposes the fact that it blocks them is entirely an implementation detail. other servers might block iqs but not expose it as such. especially bridges to other chat networks like biboumi.

--> you can't rely on this field to detect this.

tmolitor-stud-tu:
> I get that my affiliation influences if I can see the jid of participants. But I don't get how it influences if I'm able to retrieve avatars.

Because servers could choose to allow moderators/admins to send iqs to participants, but not ordinary users. Or allow it for members, but not for participants having no affiliation.

Monal/Classes/MLImageManager.m
lissine0: > And use the jid for groups (non-anon mucs).

If we're going to use occupant-id as identifier, it can work for both groups and channels.

> I think we should never use the nick to identify channel (anon muc) users, but always use the occupant-id.

The problem is, some MUC servers don't support occupant-ids. And this will always remain a possibility, because on Prosody occupant-id support depends on loading the muc_mam module.

So, if we use the occupant-id as identifier, we won't be displaying avatars in rooms without occupant-id support.

Furthermore, in those cases, the cache key for the avatar placeholders can't be based on the occupantID. So it'll have to be based on the nick.

Now, since it's based on the nick, we'll need to pass the nick to the methods reading and writing avatar files (`MLImageManager getAvatarForOccupant` and `MLImageManager setAvatarForOccupant` and their callers). At which point, why not use the nick as identifier instead of the occupant-id?

(Moreover, the nick is already needed in`MLImageManager getAvatarForOccupant` and its callers, in order to generate the eventual placeholder avatar from the first character of the nick)

I have an draft of this PR using occupantIDs. I can push it somewhere if you want to take a look

lissine0: We could avoid caching when a room doesn't support occupant-ids, but I'm not very fond of that idea.

lissine0:
> When first joining a room, mam queries don't work until ...

This is due to server rate limits. I limited the avatar query sending to have only one in flight at a time.
My implementation does the limit on the whole account though. So, if you join a large room and then immediately join a second one, the avatars of the second one only get fetched after all the avatars of the first one got fetched.

lissine0:
Note: to make review easier, I'm adding commits rather than squashing.
When you deem the PR ready for merging, I'll squash the added commits with the old ones.
(the commits that I plan to squash have messages that start in lowercase)

Monal/Classes/MLMucProcessor.m
```diff
@@ -1607,14 +1633,98 @@ -(void) sendJoinPresenceFor:(NSString*) room
...
+        [_pendingAvatarQueries removeObjectAtIndex:0];
+        if(_pendingAvatarQueries.count > 0)
+        {
+            DDLogVerbose(@"Sending avatar query to '%@'. Avatar queue size: %lu", _pendingAvatarQueries[0].to, _pendingAvatarQueries.count);
+            [_account sendIq:_pendingAvatarQueries[0] withHandler:$newHandlerWithInvalidation(self, handleParticipantVcardResponse, handleParticipantVcardResponseInvalidation, $ID(fullJid, _pendingAvatarQueries[0].to))];
+        }
+    }
+}
+
+$$instance_handler(handleParticipantVcardResponseInvalidation, account.mucProcessor, $$ID(xmpp*, account), $$ID(NSString*, fullJid))
```
lissine0: Is it correct that we need this invalidation handler?

Monal/Classes/MLImageManager.m
tmolitor-stud-tu: Yes, we SHOULD avoid caching if the room doesn't support occupant-ids because its a security issue otherwise: someone could take the same nick while you are not joined and still get **your** avatar image displayed for all Monal users. That will confuse them at best and allow for easy impersonation attacks at worst.

In the long term we should display a status message saying something along the line `A new user took the nick 'xyz'` to fully prevent such attacks.

As for occupant-ids in general: we already use them to authenticate last message correction and message retraction (using the nick would be insecure as per the impersonation attack described above).

Monal/Classes/MLMucProcessor.m
tmolitor-stud-tu: You need an invalidation handler if you want to be able to clean up things/state if the account could not resume a XEP-0198 smacks session. I believe you need it for cleaning up your requests queue and triggering the next request.

tmolitor-stud-tu:
I think we should throw in some moderate extra effort and make the queueing of vcard iqs a general mechanism with an api, probably exposed by xmpp.m or a new class.

It should have the following properties:

1. Have different queues identified by names, all queues running in parallel.
2. Every queue should always have a per-queue-configurable amount of requests in flight (or MLHandler executed), never more.
3. Each queue entry should be either an IQ to be sent or an arbitrary MLHandler.
4. Each queue should have an associated dictionary for auxiliary data passed as argument to all handlers (the IQ handlers handling the result or IQ invalidation, in the IQ case AND the "raw" MLHandler in the MLHandler case).
5. The api should provide a method to invalidate a queue, removing all items from it and calling their invalidation handlers (if provided).

Using such queues for all sorts of low-priority explicit PEP/PubSub fetches, omemo key fetches etc. would allow us to speed up various catchup scenarios etc.

@matthewrfennell do you have any feedback? Did I forget something? I'd value your insights as well :)

Monal/Classes/MLImageManager.m
lissine0: > Yes, we SHOULD avoid caching if the room doesn't support occupant-ids because its a security issue otherwise: someone could take the same nick while you are not joined and still get your avatar image displayed for all Monal users. That will confuse them at best and allow for easy impersonation attacks at worst.

Note that this security issue is not possible, unless the impersonator has the same avatar hash in their presence.
(in `MLMucProcessor processStanza:`, if we detect that the hash changed, we fetch new avatar; or we just delete the old one if there's no hash in the presence)

lissine0: And by the way, when I said:
> We could avoid caching when a room doesn't support occupant-ids

I meant the avatar placeholders saved in `MLImageManager`'s `NSCache` property.
Because if we're going to use occupant-ids as identifiers, avatars won't be displayed anyways in rooms without occupant-id support.

tmolitor-stud-tu: Hmmm, well, I get the avatar hash in presence argument. But it still doesn't feel right to use the unstable/spoofable nickname as identifier, especially since occupant-ids are widely supported now.

(Btw: you'd have to sanitize the nickname to not contain characters that allow a directory traversal attack)

tmolitor-stud-tu: And yes, I know that we'll show the placeholder avatars for mucs without occupant-ids, but imho that's fine.

matthewrfennell:
I'll first verify my understanding, then my opinion, please let me know if I misunderstood anything!

# Understanding/assumptions

At present, we don't differentiate between low and high priority IQs. We send everything straight away, via the `sendQueue`, which doesn't block waiting for handlers to be resolved.

The idea is to introduce queues. High-prio IQs will be unchanged, but low-prio IQs will be sent through these queues. We'll have a separate queue per task, for instance, when joining this MUC, a queue will be created for "fetch me all the avatars in the MUC". It gets destroyed once all handlers are resolved. If I join 3 MUCs in quick succession then there would be 3 queues at once.

XMPP does not have a way for clients to discover the server's rate limits. So, we can't guarantee we stay under the rate limit. All we can do is prevent extreme bursts of IQs, which these queues prevent especially when the number of IQs to send is out of our control (as in this PR).

# Opinion

The design makes sense to me; it's simple, logically groups IQs so they can be invalidated together, and has rate limiting per queue to prevent extreme bursts.

One doubt: I think it's still technically possible for a user to generate a lot of low-prio IQs at once, e.g. by joining lots of MUCs in quick succession.

Should we separate rate limiting from grouping IQs? So, we have a global (or per account/server combination) limit of in-flight requests, instead of per queue?

That said, it's not a likely scenario, and if a user is doing that, maybe they should understand that they are hitting the server's rate limit because of their behaviour, rather than Monal fetching everything of its own volition. What do you think?

tmolitor-stud-tu:
Thanks Matthew!! Your understanding is correct :)

Well, having a global ratelimiting queue and assigning priorities to every IQ would solve the problem, too (assuming we somehow get the ratelimiting values from the server, I asked Holger (ejabberd) about this).
But that's somewhat orthogonal to my design above: I think having both, would be good, since my queue above can be used for other things, too (hence the MLHandler OR iq).

But one thing you mentioned is crucial: joining multiple MUCs in rapid succession is common for non-smacks (XEP-0198) resumes. That means we should create one "global" avatar queue rather than one per MUC. That said, having, for example, 10 MUCs and thus 10 avatar IQs in flight isn't that much either.

The default limits for prosody are:
```
limits = {
        c2s = {
                rate = "10kb/s";
        };
        s2sin = {
                rate = "30kb/s";
        };      
}
```

Ejabberd doesn't have a default, but the example config snippet in the [Share documentation](https://docs.ejabberd.im/admin/configuration/basic/#shapers) says:
```
shaper:
  normal: 1000
  fast: 50000
```

lissine0:
> XMPP does not have a way for clients to discover the server's rate limits.


> (assuming we somehow get the ratelimiting values from the server, I asked Holger (ejabberd) about this)

Note that there is an XEP for advertising the stream rate limits: [XEP-0478](https://xmpp.org/extensions/xep-0478.html)
It's currently only supported by Prosody and Openfire, but it seems easy to implement in other servers.

tmolitor-stud-tu:
[XEP-0478](https://xmpp.org/extensions/xep-0478.html) doesn't advertise ratelimit settings, but the maximum size of a stanza (and the connection idle time allowed).

tmolitor-stud-tu:
Holger said, that the ratelimit could differ from node to node in an ejabberd cluster and generally could be some other value on the s2s side (which may involved on the path to the muc) etc.

He also said, that something must be wrong if we are hitting the ratelimit of an xmpp server and I tend to agree:
Could it be, that it isn't the server's ratelimit but parsing all incoming avatar stanzas that delay the MAM response? You can check the time between logging the `RECV:` carrying the MAM response and the `RECV Stanza:` handling the MAM response.

tmolitor-stud-tu:
That now means --> drop the idea of a global ratelimiting queue and implement my "orignal design" described here: https://github.com/monal-im/Monal/pull/1480#issuecomment-3524234805

lissine0:
> He also said, that something must be wrong if we are hitting the ratelimit of an xmpp server and I tend to agree:
> Could it be, that it isn't the server's ratelimit but parsing all incoming avatar stanzas that delay the MAM response? You can check the time between logging the `RECV:` carrying the MAM response and the `RECV Stanza:` handling the MAM response.

It is neither the server's rate limit nor the parsing of incoming stanzas.

The problem is: we send a bunch of vcard iqs, and then send a MAM iq.
The server answers iqs sequentially. So, it doesn't answer the mam query until it has answered all the vcard queries.

And answering vcard queries takes a lot of time because it involves contacting many other servers.
In a big room (400 participants) it can take 2-3 minutes.

lissine0:
I updated the PR.
Notable changes since the inital version:
- Use the room + occupantId as identifier, instead of room + nick
- Use the occupantId for the avatar placeholder color
- Save the avatars per account (So if a room is joined by two accounts in Monal, they would both download the avatars separately)
- Have only one avatar query in flight at a time (for a given account). In the previous comment I explained why this is needed (not rate limits)


I did not implement a generic abstraction for this. Because I don't think this is needed in other situations. Fetching channel participant avatars is the only time where you can send dozens or hunderds of iqs at the same time. (plus answering them takes several minutes)
So I kept things simple and just implemented it in MLMucProcessor



I'd like to optimize things to reduce the number of avatar fetches:
- The way things are implemented currently, when a participant leaves, its avatar is deleted. (this is because unavailable presences don't contain the avatar hash) 
Ideally, we'd only delete the avatar if it's unpublished, or if the participant is banned

- The second potential optimization was already discussed in https://github.com/monal-im/Monal/pull/1480#issuecomment-3368160460
>> Currently, if some participants are present in multiple rooms, their avatars are fetched once for every room.
>> It would be nice to optimize this case: before sending the iq, we could check if the avatar hash is known in other rooms. If so, we simply copy the file. otherwise we proceed with sending the iq.
>
>Well, since different mucs could have different policies for direct communication, I think that optimization would cause more confusion than do good (its hard to understand for ordinary users, why a muc shows only avatars of some users and not others).

What changed: since version 26.01, ejabberd no longer blocks avatar IQs, even when `muc#roomconfig_allow_query_users` is set to false: https://github.com/processone/ejabberd/commit/46737b115deb0dbef9da479279a8ea97660e18ba
The rationale is that avatar queries are answered by the server on behalf of the client. So no direct communication is involved. See https://github.com/processone/ejabberd/issues/4489 for more details

lissine0:
Notes:
- I didn't implement automatic view updates when the avatar changes. This PR is too big already and I wanted to limit its scope.
- I don't display the avatars of participants of private groups that aren't in our roster, even though their avatars are downloaded. Displaying them would complicate the code a bit and I don't have plans to implement it.

Monal/Classes/MLMucProcessor.m
```diff
@@ -1689,14 +1722,118 @@ -(void) sendJoinPresenceFor:(NSString*) room
...
+    [self sendTheNextPendingAvatarQuery];
+$$
+
+$$instance_handler(handleParticipantVcardResponse, account.mucProcessor, $$ID(xmpp*, account), $$ID(XMPPIQ*, iqNode), $$ID(NSString*, occupantId))
+    [self sendTheNextPendingAvatarQuery];
+
+    // Ignore the vcard response if we're no longer joined to the room
+    // We send the next pending avatar query even in this case because the next query
+    // might be for a participant in another room
+    if(![self isJoined:iqNode.fromUser] && ![self isJoining:iqNode.fromUser])
```
lissine0, line 1779: Is this the correct way to check that we're no longer joined to the room?

Monal/Classes/MLImageManager.m
```diff
@@ -141,17 +146,52 @@ -(void) cleanupHashes
             if(error)
                 DDLogError(@"Error deleting orphan avatar file: %@", error);
         }
+
+        //clean up orphan hashes and orphan avatar files of muc participants
```
lissine0, line 150: I think the entire `cleanupHashes` method can / should be removed.
I initially followed the existing code for contact avatars.

These are the only situations that I'm aware of where a hash does not have a corresponding avatar file:
- The PEP node of the participant avatar does not have `open` access model. as per [XEP-0398](https://xmpp.org/extensions/xep-0398.html#presence), the server would still inject the hash to the presence in this case:
> services MUST include the hash on behalf of their users in every available presence that does not contain an empty photo element wrapped in an x element qualified by the 'vcard-temp:x:update' namespace.

- The participant avatar is in a file format that Monal doesn't understand. For example I know a user who has an AVIF avatar. At least as of now, Monal tries to load the avatar image (in order to resize it, make it circular and convert it to PNG) and if it can't do that, [it doesn't save the file](https://github.com/monal-im/Monal/pull/1480/commits/89c1379d325efe72437796a2fcc04f49962b3d64#diff-88fdaba60a41ca7722798305b134c137a37cec17f127f9c1293443bca5dd33fbR1807-R1812)

In both of these cases, if `cleanupHashes` removes the hash, the next time we receive a presence from the participant, the hash will get saved again, and the avatar will try to be fetched again, which is wasteful since we know it will fail.


As for situations where we have the avatar file but don't have a corresponding hash, I added separate commits that address them:
- When removing a contact
- When leaving a room
- When removing / deleting an account

To conclude, I think we should remove `cleanupHashes`, even the part concerning contact avatars, unless you have an objection.

lissine0:
I investigated the avatar queries delaying MAM issue a bit more:
- While fetching participant avatars, MAM queries are delayed only for rooms on the same MUC server. MAM queries to rooms on other servers get answered in a timely manner.
- The rate limits of the home server can make fetching avatars at once much slower, but they're not the cause of the MAM delays, since MAM queries to other MUC servers are still fast. (In fact, using an account on my prosody server, fetching 200 avatars takes 2m45s with the default rate limits and 35s using increased rate limits)


There was a bit of discussion about avatar queries delaying MAM [here](https://logs.xmpp.org/xsf/2026-05-06)
One suggestion was to send the MAM query at the same time as the join presence, but that is not ideal for two reasons:
- The user can scroll up to fetch more history, and will notice the delay then
- Currently in Monal we send a first iq to get the most recent stanzaId before sending the avatar query. If we want to query MAM while sending the join presence, we'll need to change this.


We may want to keep the approach that's currently implemented in this PR: have only one avatar query in flight at a time. The avatars will take a very long time to get fetched, but if we optimize things to remove unnecessary queries (e.g. avoiding deleting the avatar when receiving unavailable presence, unless the user was banned or something, see a previous comment of mine), then users won't notice except the first time joining the room.

tmolitor-stud-tu:
We could delay the deletion of avatars not used anymore by simply changing the mtime of the avatar files any time we see that a user advertises that avatar and simply run a background delete for avatars older than, say, 3 or 7 days, or min(message autodelete, 3 days)...

tmolitor-stud-tu:
but I'd really like a generic framework for: queue iqs in queue A with at most X items in flight, allowig for multiple queues having different settings for the parallel in-flight items.

tmolitor-stud-tu:
we could even increase and decrease the parallel iqs in a avatar fetch queue depending on the mam state (no mam versus mam running), so, for example, only 1 in-flight query if mam is running and 4 queries in parallel if no mam is running for that muc

PR #1519, opened 2025-12-17
Support sending media attachments in the new ChatView

lissine0:


Monal/Classes/ChatView.swift
tmolitor-stud-tu: use `??` to specify a default value for both these optionals

tmolitor-stud-tu (changes requested):
Small change needed to silence the compiler warning and make the UX *slightly* better

PR #1527, opened 2025-12-22
Improvements and bug fixes

lissine0:
This PR:
- Fixes the unstable order of reactions from the same user
- Addresses the review feedback given in #1424
- Renames a few properties, and adds a new one

I can squash the #1424 fixes in one commit if you want.

P.S. we have to use `NSMutableOrderedSet` in Swift, because it has no bridged Swift-native equivalent. Yes there is `OrderedSet` in `swift-collections`, but it's not bridged to `NSOrderedSet`. (meaning you can't use it when calling methods defined in objective-C that expect `NSOrderedSet`)

### DIFF
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -882,15 +882,15 @@ class ChatViewMessage: ExyteChat.Message {
                 MLFiletransfer.checkMimeTypeAndSize(forHistoryID: innerMessage.messageDBId)
                 return []
             }
-            let cacheFile = fileInfo.cacheFile as String
-            let attachmentUUID = HelperTools.stringToUUID(cacheFile).uuidString
+            let cacheFilePath = fileInfo.cacheFilePath as String
+            let attachmentUUID = HelperTools.stringToUUID(cacheFilePath).uuidString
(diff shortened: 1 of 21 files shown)


Monal/Classes/ChatView.swift
```diff
@@ -883,7 +883,8 @@ class ChatViewMessage: ExyteChat.Message {
                 return []
             }
             let cacheFilePath = fileInfo.cacheFilePath as String
-            let attachmentUUID = HelperTools.stringToUUID(cacheFilePath).uuidString
+            let cacheId = fileInfo.cacheId as String
+            let attachmentUUID = HelperTools.stringToUUID(cacheId).uuidString
```
lissine0, line 887: Basing the UUID on the cacheId might be better (the cacheId is the filename of the cache file, whereas cacheFilePath is the absolute path of the file)
For example, with the cacheId, the UUID would remain stable after a device migration.
What do you think?

tmolitor-stud-tu: yes, good idea. although I don't think that ExyteChat needs a stable id across app restarts here.

tmolitor-stud-tu (approved):

PR #1531, opened 2025-12-24
Bump SwiftPM dependencies

lissine0:
This updates our ExyteChat fork with two things:
- It fixes a crash in the audio recorder that affects small-screen devices. I already opened a PR in the upstream ExyteChat with this fix.
- It fixes the displayed time in the audio recorder not getting updated unless the user presses one of the buttons.
It turns out this is caused by our switch from struct to class for `Recording`.
And since we're not extending `Recording` in Monal, I switched it back to a struct.
See https://github.com/monal-im/ExyteChat/commit/9852f899cea836194c67597fe5e2a1d9e72797c7
I don't think we'll need to extend it in the future (it's a very simple struct), but should the need ever arise, we can then reverse this change and look for a fix that keeps `Recording` a class

A demonstration of the second fixed issue:
Before:

https://github.com/user-attachments/assets/b0178c0d-7635-43c5-9878-ee33642cd5e2

After:

https://github.com/user-attachments/assets/0116621f-e5e3-48cd-822d-43174e04cfb9

### DIFF
--- a/Monal/Classes/ChannelMemberList.swift
+++ b/Monal/Classes/ChannelMemberList.swift
@@ -14,6 +14,7 @@ struct ChannelMemberList: View {
     @State private var ownRole: String;
     @StateObject var channel: ObservableKVOWrapper<MLContact>
     @State private var participants: OrderedDictionary<String, [String:String]>
+    @StateObject private var overlay = LoadingOverlayState()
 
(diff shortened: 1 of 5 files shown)


tmolitor-stud-tu:
I force pushed, please rebase

tmolitor-stud-tu:
never mind, I'll do a cherry-pick :)

PR #1533, opened 2025-12-25
Fix compilation errors, and backport some unit tests (oldstable/6.4-line branch)

lissine0:


Monal/Classes/MemberList.swift
```diff
@@ -292,21 +286,6 @@ struct MemberList: View {
                                 view
                             }
                         }
-                        .swipeActions(allowsFullSwipe: false) {
```
lissine0, line 295: This is redundant with the `.onDelete` down below.
Moreover, `.swipteActions` is only available on iOS15.0+

tmolitor-stud-tu: Must have been a leftover of my conflict solving attempt.

PR #1534, opened 2025-12-25
1.17.0

lissine0:
- Fix random black video feed on video calls
- Fix accidental reordering of chats when opening a chat
- Fix automatic speaker activation in video calls
- Don't mirror local camera view in video calls
- Don't play call error sounds indefinitely
- Fix crash on incoming non-spec-conformant LMC
- Improve OMEMO discovery for chats with users not in our contact list
- Update ssdp implementation to version 0.4
- Properly open links in external browser, if configured to do so
- Fix the discovery of jid type (muc/account) for some bridges


tmolitor-stud-tu:
it is changelog only in the description and version number in the title. also make sure the quicksy build workflow in this branch matches the one in the develop branch or the one in the current stable branch. not sure if they differ and what workflow would be best.

PR #1536, opened 2025-12-26
Fix the handling of stanzaId updates in MLMessage

lissine0:
Update the `stanzaId` property only if a `stanzaId` was provided in the notification.
This prevents the `stanzaId` property from wrongly becoming `nil` while handling message update notifications that don't contain a `stanzaId` update.

### DIFF
--- a/Monal/Classes/MLMessage.m
+++ b/Monal/Classes/MLMessage.m
@@ -180,15 +180,20 @@ static NSMutableDictionary* _singletonCache;
         MLAssert(correctedText != nil, @"Message correction notification without the corrected text");
         self.messageText = correctedText;
     }
-    else if(reactionsUpdate.boolValue)
+
+    if(reactionsUpdate.boolValue)
     {
+        // Reactions update
         MLAssert(data[@"reactions"] != nil, @"Reactions updates MUST contain a reactions dictionary!");
         self.reactions = data[@"reactions"];
     }
-    
-    // MUC reflection and everything else
-    DDLogDebug(@"Updating stanzaid of %p: %@", self, data[@"stanzaId"]);
-    self.stanzaId = data[@"stanzaId"];
+
+    if(data[@"stanzaId"])
+    {
+        // MUC reflection
+        DDLogDebug(@"Updating stanzaid of %p: %@", self, data[@"stanzaId"]);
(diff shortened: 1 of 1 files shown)


Monal/Classes/MLMessage.m
```diff
@@ -180,15 +180,21 @@ -(void) handleMessageUpdate:(NSNotification*) notification
         MLAssert(correctedText != nil, @"Message correction notification without the corrected text");
         self.messageText = correctedText;
     }
-    else if(reactionsUpdate.boolValue)
+    else
```
tmolitor-stud-tu: why this nested `if` instead of using an `else if`?

tmolitor-stud-tu (changes requested):

Monal/Classes/MLMessage.m
lissine0: Hmm. It was originally written like that: https://github.com/monal-im/Monal/commit/024a801e16e0beeb9120baf9498838b77b523d9f#diff-a7c63cd3939168b512e4c538a1d8cbaece6cf5624db9735dede5a5d2d7bf92c8
Why did you change it?
I assumed that you changed it because you wanted to allow a single notification to update both the stanzaId and the reactions.
And your comment "MUC reflection _and everything else_" made me think you intended it with a nested if like I did here.

Now, checking `MLMessageProcessor`, I verified that the notification updates only one thing at a time. Either it's a LMC, or a reaction, or a MUC reflection.
So, the original `if`-`else if`-`else` works.
But if you want to make it work without any assumptions of how `MLMessageProcessor` is written, you can do it like this:

```objc
if(LMCReplaced.boolValue)
{

}

if(reactionsUpdate.boolValue)
{

}

if(data[@"stanzaId"])
{

}
```
Anyway, if you want to, feel free to close this PR and fix the issue how you see fit

tmolitor-stud-tu: Thanks for this long explanation! Now I understand what you were thinking :)

I guess the three independent ifs are the best, no assumptions about the message processor sounds good :)

I'll merge your PR and then change it to be three ifs :)

lissine0: No need, I'll change it to three ifs now

lissine0: Done

PR #1539, opened 2025-12-28
Make links clickable in the new ChatView

lissine0:
Also fix a couple of bugs in `ExternalLinkViewModifier`.

### DIFF
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -1008,8 +1008,8 @@ struct MessageView: View {
             chatType: .conversation,
             avatarSize: 32,
             tapAvatarClosure: nil,
-            messageStyler: AttributedString.init,
-            shouldShowLinkPreview: { _ in true },
+            messageStyler: { $0.linkify() },
+            shouldShowLinkPreview: { _ in false },  //disabled for now due to https://github.com/exyte/Chat/issues/208
             isDisplayingMessageMenu: false,
             showMessageTimeView: true,
             messageLinkPreviewLimit: 8,
--- a/Monal/Classes/SwiftuiHelpers.swift
+++ b/Monal/Classes/SwiftuiHelpers.swift
@@ -541,23 +541,11 @@ struct NumberlessBadge: View {
 
 struct LinkedText: View {
     let attributed: AttributedString
-    
+
     init(_ text: String) {
-        //make links in text clickable
-        var attributed = AttributedString(text)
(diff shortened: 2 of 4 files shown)


Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -540,26 +540,12 @@ struct NumberlessBadge: View {
...
-                }
-            }
-        }
-        self.attributed = attributed
+        self.text = text
     }
-    
     var body: some View {
-        Text(attributed)
+        Text(text.linkify())
```
tmolitor-stud-tu: This will enumerate the string on every swiftui redraw which is not very efficient, whereas the original design did the enumeration only once.

tmolitor-stud-tu (changes requested):

Monal/Classes/ChatView.swift
```diff
@@ -1008,8 +1008,8 @@ struct MessageView: View {
             chatType: .conversation,
             avatarSize: 32,
             tapAvatarClosure: nil,
-            messageStyler: AttributedString.init,
-            shouldShowLinkPreview: { _ in true },
+            messageStyler: { $0.linkify() },
+            shouldShowLinkPreview: { _ in false },  //disable link previews for now
```
tmolitor-stud-tu: Why did you disable the link preview?

lissine0: Because of a flickering bug in ExyteChat. I clarified this in the comment now.

lissine0:
Note: as implemented in https://github.com/monal-im/Monal/commit/6628ff17f6d521a20fecba6608ea1160483bb23f, the inline Safari view opens in a sheet, which doesn't feel nice for web browsing IMHO:

https://github.com/user-attachments/assets/694871ce-e5f1-4613-ab86-2563231ec9f3

So, in my last commit I used fullScreenCover instead. I had to also ignore the safe area, otherwise it would look like this:
(Notice the black bars at the top and bottom)
simulator_screenshot_9719DB8D-6482-47C6-A22D-1F2DA46B485F

The problem is, with fullscreen cover + ignoring the safer area, there's a bug when you rotate the phone twice:
the toolbar becomes inside the safe area, and you can't click the Done button:

https://github.com/user-attachments/assets/9750aec7-9899-416e-9900-add04d14b278

Luckily, this bug has been fixed in iOS 26:

https://github.com/user-attachments/assets/987c61c0-3ce2-4ef6-b5b7-06ed4faf1553

What do you think?

tmolitor-stud-tu:
Perfect now, indeed fullscreen cover looks way better :)

PR #1560, opened 2026-01-29
Fix a couple of bugs

lissine0:
- Set the `accountID` and `buddyName` MLMessage properties for message drafts, because some parts of the code base expect them to be not nil e.g. `MLContact createContactFromJid:andAccountID`
- Fix a mistake in the reactions fallback handling commit


lissine0:
Closing as the bugs are now fixed

PR #1568, opened 2026-02-09
Generate image thumbnails and use thumbnails in the media gallery

lissine0:
- Generate thumbnails for image attachments, including for SVGs
- Use the image thumbnails in the ChatView. This should make the loading of images much faster. I also used HEIC or JPEG instead of PNG for the thumbnails, to make the size smaller and reduce the loading times even more.
- Use the generated thumbnails of videos and images in the media gallery. This has a noticeable improvement on the loading time, and also fixes some pre-existing issues (like thumbnails for videos and for SVGs which previously didn't work)

A few notes:
- ExyteChat doesn't already have support displaying SVGs, so we don't have to use their SVG library (the original argument for using it being that it's already in our dependency tree.)
We may want to consider this as the Exyte library has poor SVG support. In fact doesn't even render the recently added Monal SVG logos correctly.
- GIFs (and other animated image formats like WebP I presume) are displayed as still images.
The removed Giphy library isn't useful in this context, as I'd imagine it works by just sending an image link from one phone to another, and the receiving end would download the image from the Giphy server. i.e. it wouldn't handle animated images sent from a non-giphy client like Gajim.
Therefore, if we want to display animated images as animated we'll have to patch our exyteChat fork and add another library. Note that FLAnimatedImage (used in the old chatView for GIF displaying) doesn't support WebP.
(Animated image support is out of scope of this PR, as it needs to be done in the exytechat fork anyway. I just wanted to discuss it.)

### DIFF
--- a/Monal/Classes/HelperTools.m
+++ b/Monal/Classes/HelperTools.m
@@ -51,6 +51,7 @@ extern int64_t kscrs_getNextCrashReport(char* crashReportPathBuffer);
 #import <monalxmpp/MLContact.h>
 #import <monalxmpp/MLMessage.h>
 #import <monalxmpp/MLFileTransfer.h>
+#import <monalxmpp/MLFiletransferInfo.h>
 #import <monalxmpp/DataLayer.h>
 #import <monalxmpp/OmemoState.h>
 #import <monalxmpp/MLUDPLogger.h>
@@ -1169,7 +1170,55 @@ static void notification_center_logging(CFNotificationCenterRef center, void* ob
     }];
 }
 
-+(AnyPromise*) generateVideoThumbnailFromFile:(NSString*) file havingMimeType:(NSString*) mimeType andFileExtension:(NSString* _Nullable) fileExtension
++(AnyPromise*) generateThumbnailFromFile:(MLFiletransferInfo*) fileInfo
+{
+    if(fileInfo.isImage)
+        return [HelperTools generateThumbnailFromImageFile:fileInfo];
+    else if(fileInfo.isVideo)
+        return [HelperTools generateThumbnailFromVideoFile:fileInfo.cacheFilePath havingMimeType:fileInfo.mimeType andFileExtension:fileInfo.fileExtension];
+    else
+        unreachable(@"Trying to generate a thumbnail for a file that is neither an image nor a video");
+}
+
++(AnyPromise*) generateThumbnailFromImageFile:(MLFiletransferInfo*) fileInfo
+{
+    return [AnyPromise promiseWithResolverBlock:^(PMKResolver resolve) {
+        AnyPromise* imagePromise = nil;
+        if(fileInfo.isSVGImage)
+            imagePromise = [HelperTools renderUIImageFromSVGURL:fileInfo.fileURL];
+        else
+            imagePromise = [AnyPromise promiseWithValue:[[UIImage alloc] initWithContentsOfFile:fileInfo.cacheFilePath]];
+        imagePromise.then(^(UIImage* image) {
+             if(image == nil && fileInfo.isSVGImage)
+                return resolve([NSError errorWithDomain:@"Monal" code:0 userInfo:@{NSLocalizedDescriptionKey: @"Could not load SVG image to generate its thumbnail."}]);
+
+            // The size passed to prepareThumbnailOfSize: is the maximum dimensions, because the method
+            // preserves the original aspect ratio.
+            // For example, if we pass it 200x200 size and a 2:1 image, the output will be a 200x100 thumbnail
+            // So, we calculate the size that should be passed to the method, such that its output has 200x200 as
+            // a minimum size. i.e. in the previous example that would be 400x200
+            // That ensures that the thumbnail will look as good as possible in the ChatView
+            // (which currently crops the thumbnail to 204x200, and in some cases (using SFS) 204x100 or 100x100)
+            CGFloat scale = [UIScreen mainScreen].scale;
+            CGFloat minSize = 200 * scale; // minimum width and height are both 200
+            CGFloat aspectRatio = image.size.width / image.size.height;
+            CGSize targetSize;
+            if(aspectRatio > 1.0)
+                targetSize = CGSizeMake(minSize * aspectRatio, minSize);
+            else
+                targetSize = CGSizeMake(minSize, minSize / aspectRatio);
+
+            [image prepareThumbnailOfSize:targetSize completionHandler:^(UIImage* thumbnail) {
+                if(thumbnail == nil)
+                    return resolve([NSError errorWithDomain:@"Monal" code:0 userInfo:@{NSLocalizedDescriptionKey: @"Could not generate thumbnail of image file"}]);
+                else
+                    return resolve(thumbnail);
+            }];
+        });
+    }];
+}
+
++(AnyPromise*) generateThumbnailFromVideoFile:(NSString*) file havingMimeType:(NSString*) mimeType andFileExtension:(NSString* _Nullable) fileExtension;
 {
     return [AnyPromise promiseWithResolverBlock:^(PMKResolver resolve) {
         [self createAVURLAssetFromFile:file havingMimeType:mimeType andFileExtension:fileExtension withCompletionHandler:^(AVURLAsset* asset) {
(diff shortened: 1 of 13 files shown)


Monal/Classes/HelperTools.m
```diff
@@ -1167,7 +1168,55 @@ +(AnyPromise*) computeMediaDurationFromFile:(NSString*) file havingMimeType:(NSS
...
+    else
+        unreachable(@"Trying to generate a thumbnail for a file that is neither an image nor a video");
+}
+
++(AnyPromise*) generateThumbnailFromImageFile:(MLFiletransferInfo*) fileInfo
+{
+    return [AnyPromise promiseWithResolverBlock:^(PMKResolver resolve) {
+        AnyPromise* imagePromise = nil;
+        if(fileInfo.isSVGImage)
+            imagePromise = [HelperTools renderUIImageFromSVGURL:fileInfo.fileURL];
```
lissine0, line 1188: The existing Monal code often uses `PMKHang` when loading SVGs. What does that do and should I use it here?

tmolitor-stud-tu: Since ObjC doesn't have async/await, I'm using `PMKHang` to hang the current thread until the given promise is resolved. It should obviously only be used in a thread that is allowed to hang (never in the main thread for example).

See the implementation here: https://github.com/mxcl/PromiseKit/blob/2bc44395edb4f8391902a9ff7c220471882a4d07/Sources/hang.m#L10

tmolitor-stud-tu: Ah wait, it can even be used in the main thread, because it spins the runloop while "hanging".

tmolitor-stud-tu: It is the ObjC pendant of the async/await implementation for promises over here: https://github.com/mxcl/PromiseKit/blob/2bc44395edb4f8391902a9ff7c220471882a4d07/Sources/Async.swift

tmolitor-stud-tu: Okay, another note: the swift version of "hang" says in the comments:
```
// hang doesn't make sense on threads that aren't the main thread.
// use `.wait()` on those threads.
```

So the swift version and the ObjC versions seem to be slightly different.

tmolitor-stud-tu: To answer your question now: No, you don't need to use `PMKHang` here. It should only be used if you need the result of the svg promise in the following synchronous code (like in the notification manager) and it is okay to hang the current thread while waiting for the svg promise (the notification manager will hang the receive queue which is okay).

Monal/Classes/HelperTools.m
```diff
@@ -1169,7 +1170,55 @@ +(AnyPromise*) computeMediaDurationFromFile:(NSString*) file havingMimeType:(NSS
...
+{
+    if(fileInfo.isImage)
+        return [HelperTools generateThumbnailFromImageFile:fileInfo];
+    else if(fileInfo.isVideo)
+        return [HelperTools generateThumbnailFromVideoFile:fileInfo.cacheFilePath havingMimeType:fileInfo.mimeType andFileExtension:fileInfo.fileExtension];
+    else
+        unreachable(@"Trying to generate a thumbnail for a file that is neither an image nor a video");
+}
+
++(AnyPromise*) generateThumbnailFromImageFile:(MLFiletransferInfo*) fileInfo
```
tmolitor-stud-tu, line 1183: Wouldn't it be better to move this method to `MLFiletransferInfo`? Maybe even make it a readonly property (hanging for the result rather than returning a promise).

Monal/Classes/HelperTools.m
```diff
@@ -1169,7 +1170,55 @@ +(AnyPromise*) computeMediaDurationFromFile:(NSString*) file havingMimeType:(NSS
...
 
-+(AnyPromise*) generateVideoThumbnailFromFile:(NSString*) file havingMimeType:(NSString*) mimeType andFileExtension:(NSString* _Nullable) fileExtension
++(AnyPromise*) generateThumbnailFromFile:(MLFiletransferInfo*) fileInfo
+{
+    if(fileInfo.isImage)
+        return [HelperTools generateThumbnailFromImageFile:fileInfo];
+    else if(fileInfo.isVideo)
+        return [HelperTools generateThumbnailFromVideoFile:fileInfo.cacheFilePath havingMimeType:fileInfo.mimeType andFileExtension:fileInfo.fileExtension];
+    else
+        unreachable(@"Trying to generate a thumbnail for a file that is neither an image nor a video");
```
tmolitor-stud-tu, line 1180: Maybe just return a SFSymbol image (document/file image) and use this method for every filetransfer.

The `HelperTools` method for handling a file shared with Monal does even have some thumbnail generating code that could be reused here, so every location showing some thumbnail would show the same thumbnail. That would be the best solution.

Monal/Classes/HelperTools.m
```diff
@@ -1169,7 +1170,55 @@ +(AnyPromise*) computeMediaDurationFromFile:(NSString*) file havingMimeType:(NSS
     }];
 }
 
-+(AnyPromise*) generateVideoThumbnailFromFile:(NSString*) file havingMimeType:(NSString*) mimeType andFileExtension:(NSString* _Nullable) fileExtension
++(AnyPromise*) generateThumbnailFromFile:(MLFiletransferInfo*) fileInfo
```
tmolitor-stud-tu, line 1173: Wouldn't it be better to move this method to MLFiletransferInfo? Maybe even make it a readonly property (hanging for the result rather than returning a promise).

Monal/Classes/HelperTools.m
```diff
@@ -1169,7 +1170,55 @@ +(AnyPromise*) computeMediaDurationFromFile:(NSString*) file havingMimeType:(NSS
...
+                if(thumbnail == nil)
+                    return resolve([NSError errorWithDomain:@"Monal" code:0 userInfo:@{NSLocalizedDescriptionKey: @"Could not generate thumbnail of image file"}]);
+                else
+                    return resolve(thumbnail);
+            }];
+        });
+    }];
+}
+
++(AnyPromise*) generateThumbnailFromVideoFile:(NSString*) file havingMimeType:(NSString*) mimeType andFileExtension:(NSString* _Nullable) fileExtension;
```
tmolitor-stud-tu: Shouldn't this be modernized to also use the `MLFiletransferInfo*` argument?
Also: Wouldn't it be better to move this method to MLFiletransferInfo? Maybe even make it a readonly property (hanging for the result rather than returning a promise).

Monal/Classes/MLFiletransferInfo.m
```diff
@@ -162,6 +162,9 @@ -(double) mediaDuration {
 }
 
 -(NSURL*) thumbnailURL {
+    if(!self.isImage && !self.isVideo)
+        return nil;
```
tmolitor-stud-tu, line 166: Maybe always return a thumbnail and reuse the thumbnail generating code from the `HelperTools` method `addUploadItemPreviewForItem:` (see my other comment above)?

Monal/Classes/HelperTools.m
tmolitor-stud-tu: Or even better: remove it and use the `HelperTools` method `addUploadItemPreviewForItem:` as stated in a comment further down.

Monal/Classes/HelperTools.m
tmolitor-stud-tu: That would be the `HelperTools` method `addUploadItemPreviewForItem:`, see my other comments.

tmolitor-stud-tu (changes requested):

tmolitor-stud-tu:
Regarding the SVG library choice: sure, lets switch to a better lib. Do you know of any (should ideally be lightweight).

tmolitor-stud-tu:
Yeah, Giphy is just a library of gif images / (sometimes) animated images, no rendering library.

Monal/Classes/HelperTools.m
lissine0: I'm aware of `addUploadItemPreviewForItem`. I intentionally didn't use it, for future-proofing purposes:
If we use it now to generate generic thumbnails for all kinds of files, in the future we may start to generate actual thumbnails for PDFs for example. But then pre-existing PDFs would keep showing the generic thumbnails that were already generated in the past. (because we cache the thumbnails on disk)

Monal/Classes/HelperTools.m
lissine0: I can move this to MLFiletransferInfo, but I don't see the point in making it a readonly property.
The `thumbnailURL` property is already readonly, with automatic updates in the UI, and disk / memory cache. Its setter calls this method (`generateThumbnailFromFile`) if a thumbnail wasn't generated and cached previously.

Monal/Classes/HelperTools.m
lissine0: My plan was actually to get rid of `addUploadItemPreviewForItem:` after the old chat view is deleted.
That method for example generates video thumbnails correctly on macOS, but returns a generic one when selecting a video for upload on iOS.

Monal/Classes/HelperTools.m
tmolitor-stud-tu: Okay, makes sense, I guess.

Monal/Classes/HelperTools.m
tmolitor-stud-tu: We could easily clear the disk cache once we add new thumbnail types (e.g. PDF previews like in your example above).

Returning only a generic icon for videos on iOS could be fixed, I guess?

tmolitor-stud-tu:
this needs rebasing, also: the svg library was now changed.

PR #1591, opened 2026-02-28
Fix filetransfer-related crashes

lissine0:


Monal/Classes/ChatView.swift
```diff
@@ -458,6 +458,7 @@ struct ChatView: View {
                             DataLayer.sharedInstance().deleteMessageHistoryLocally(mlMessage.messageDBId)
                         }.value
 
+                        mlMessage.deletedLocally = true
```
tmolitor-stud-tu: I don't like this: having to delete the message in the db and then manually update the model class is a bad pattern. Please add a "deleteLocally" method to the MLMessage class and let that method update its state and the database.

Additionally: are we sure the same bug can't be triggered by sending an image and then retracting it while the headers fetch or download is running? If yes, we should fix this as well. If no, maybe the fix for the retraction case and this fix could maybe be merged somehow?

Monal/Classes/ChatView.swift
```diff
@@ -944,7 +945,9 @@ class ChatViewMessage: ExyteChat.Message {
 
             guard (fileInfo.downloadState as DownloadState.RawValue) == DownloadState.complete.rawValue else {
                 // TODO: show a proper button for mime type checks instead of doing this automatically
-                MLFiletransfer.checkMimeTypeAndSize(forHistoryID: innerMessage.messageDBId)
+                if !innerMessage.deletedLocally {
```
tmolitor-stud-tu: The MLFiletransfer class should do this sanity check on its own, rather than relying on something outide like the chat view or other things only using the MLFiletransfer api.

Monal/Classes/chatViewController.m
```diff
@@ -2553,6 +2553,7 @@ -(UISwipeActionsConfiguration*) tableView:(UITableView*) tableView trailingSwipe
 
     UIContextualAction* localDeleteAction = [UIContextualAction contextualActionWithStyle:UIContextualActionStyleDestructive title:NSLocalizedString(@"Delete", @"Chat msg action") handler:^(UIContextualAction* action, UIView* sourceView, void (^completionHandler)(BOOL actionPerformed)) {
         [[DataLayer sharedInstance] deleteMessageHistoryLocally:message.messageDBId];
+        message.deletedLocally = YES;
```
tmolitor-stud-tu: Same as my first comment: belongs into the model class together with the db delete.

Monal/Classes/MLFiletransferInfo.m
```diff
@@ -114,27 +114,19 @@ -(void) refresh
...
-    // return already cached value.
-    if(_downloadState != DownloadStateUndefined)
-        return _downloadState;
-
-    DownloadState state;
-    if(self.cacheFilePath)
-        state = DownloadStateComplete;
-    else if(self.size && self.mimeType && ![self.mimeType isEqualToString:@""])
-        state = DownloadStateHeaders;
+    if(self.mimeType && self.mimeType.length > 0 && self.cacheFilePath)
```
tmolitor-stud-tu: Couldn't that be reduced to a check for `self.cacheFilePath != nil`? Also: please always do an explicit `== nil` or `!= nil` check rather than the implicit `if(something)`. I know our codebase isn't strictly following this pattern everywhere, but Friedrich and I agreed to follow this pattern at least for new code.

Monal/Classes/MLMessage.h
```diff
@@ -79,6 +79,8 @@ The of the message in the DB , should be int
 
 @property (nonatomic, assign) BOOL retracted;
 
+@property (nonatomic, assign) BOOL deletedLocally;
```
tmolitor-stud-tu, line 88: Just for my understanding: This isn't backed by a db field because it is only used to track if the message is gone while the model class is still instantiated, right? So once this instance is gone, everything is fine, right?

tmolitor-stud-tu (changes requested):

Monal/Classes/MLMessage.h
lissine0: Yes, exactly.

Monal/Classes/ChatView.swift
lissine0: > Additionally: are we sure the same bug can't be triggered by sending an image and then retracting it while the headers fetch or download is running?

Yes, the same bug happens for retraction. But the PR already fixes the bug for both retraction and local deletion: in MLFiletransfer.m the condition is `if(msg.retracted || msg.deletedLocally)`

tmolitor-stud-tu: Do you see any chance of merging the two things (retracted + deleted locally) into one?

Monal/Classes/MLFiletransfer.m
```diff
@@ -61,6 +61,11 @@ +(void) checkMimeTypeAndSizeForHistoryID:(NSNumber*) historyId
         DDLogError(@"historyId %@ does not yield an MLMessage object, aborting", historyId);
         return;
     }
+    if(msg.deletedLocally)
```
tmolitor-stud-tu: Why don't you also check `retracted` here?

lissine0: Because the method `checkMimeTypeAndSizeForHistoryID:` is called in two places:
[here](https://github.com/monal-im/Monal/blob/21d8da09f90fdae89e8d0a994369d6958848665e/Monal/Classes/MLMessageProcessor.m#L836) and [here](https://github.com/monal-im/Monal/blob/21d8da09f90fdae89e8d0a994369d6958848665e/Monal/Classes/ChatView.swift#L947)

And each time, there's a check that `messageType == kMessageTypeFiletransfer` before calling the method (a retracted message has a messageType equal to kMessageTypeText)

I can add it as a sanity check if you want, but there's no bug if we don't. (whereas there is a bug if we don't check that the file was locally deleted, due to the automatic mimeType check in each access to `ChatViewMessage.attachments`)

Monal/Classes/ChatView.swift
lissine0: I think the code may be easier to understand if we keep them separate.

`deletedLocally` would concern only the case of the `MLMessage` instance remaining in memory after the message deletion from the db.

Whereas `retracted` is used many contexts.

Moreover, if we set the `retracted` property when deleting locally, the messageText will briefly be "This message got retracted" while the deletion animation is being displayed. This could be confusing to users.

Monal/Classes/MLFiletransfer.m
tmolitor-stud-tu: yes please, the more sanity the better :D

Monal/Classes/ChatView.swift
tmolitor-stud-tu: Hmmm well, I couldn't think of a good solution combining these either.

PR #1599, opened 2026-03-06
Avoid message duplication when getting MAM backscrolling results for 1-1 chats

lissine0:
This fixes the following fatalError: `Messages can not have duplicate ids, please make sure every message gets a unique id`

Duplicate ids can happen when backscrolling in a 1-1 chat because the oldestStanzaId doesn't necessarily correspond to the earliest message we have. This is because messages we send don't have a stanzaId.
This means the results of backscrolling in a 1-1 chat risk having messages that are already displayed in the ChatView, which triggers the crash.
=> we need to deduplicate the results in the case of 1-1 chats.

###### Note:
Rather than storing the messages in an OrderedSet, I kept them in an array, and used the OrderedSet only for deduplication.
I didn't benchmark, but doing this should be a little bit faster than storing in an OrderedSet, especially as the user loads more messages in the ChatView.

### DIFF
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -351,7 +351,10 @@ struct ChatView: View {
                     }
                 } else {
                     DDLogVerbose("Got backscrolling mam response: \(returnedMessages.count) messages: \(String(describing:returnedMessages))")
-                    self.messages.insert(contentsOf: returnedMessages.map {ChatViewMessage($0)}, at: 0)
+                    // When backscrolling in 1-1 chats, the oldest stanzaId may not correspond to the earliest
+                    // message we have, because messages we send don't have a stanzaId => There's a risk of
+                    // getting duplicate messages in the messages array => we need to deduplicate.
+                    self.messages = OrderedSet(returnedMessages.map {ChatViewMessage($0)} + self.messages).elements
                 }
             }
             .catch { error in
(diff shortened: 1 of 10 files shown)


Monal/Classes/SwiftuiHelpers.swift
```diff
@@ -14,9 +14,9 @@
 @_exported import SwiftUI
 @_exported import monalxmpp
 @_exported import Combine
+@_exported import OrderedCollections
```
tmolitor-stud-tu: I'd import it in SwiftHelpers rather than SwiftuiHelpers (ordered collections aren't ui specific. But other than that, I like that you opted to globalize this :)

Monal/Classes/ChatView.swift
```diff
@@ -351,7 +351,14 @@ struct ChatView: View {
                     }
                 } else {
                     DDLogVerbose("Got backscrolling mam response: \(returnedMessages.count) messages: \(String(describing:returnedMessages))")
-                    self.messages.insert(contentsOf: returnedMessages.map {ChatViewMessage($0)}, at: 0)
+                    if self.contact.isMuc {
```
tmolitor-stud-tu: I think we should not distinguish between MUC and 1:1 here. Sure, we *currently* don't need to deduplicate things for MUC, but doing it indiscriminately will save us headaches if that underlying assumption changes in the future. And, like you, I don't think it hurts much performance wise.

tmolitor-stud-tu (changes requested):
2 small changes

PR #1616, opened 2026-04-08
Fix a few bugs

lissine0:


Monal/Classes/ChatView.swift
tmolitor-stud-tu, line 587: I don't think this localization comment is right ;)

tmolitor-stud-tu (approved):
Thanks! :)

PR #1618, opened 2026-04-10
Fix a couple of bugs

lissine0:
The first commit fixes crashes in https://github.com/monal-im/Monal/blob/dd38f5e6ed7485621448b74e95ca56b92e9c8c09/Monal/Classes/ChatView.swift#L869 and https://github.com/monal-im/Monal/blob/dd38f5e6ed7485621448b74e95ca56b92e9c8c09/Monal/Classes/ChatView.swift#L959 when joining the same room with multiple accounts

The 5th commit fixes the occupantId being displayed as nickname of users who left a public channel.

The commit messages contain more details.

### DIFF
--- a/Monal/Classes/MLChannelContact.m
+++ b/Monal/Classes/MLChannelContact.m
@@ -115,8 +115,6 @@ static NSMutableDictionary* _singletonCache;
         NSDictionary* participantInfo = [[DataLayer sharedInstance] getParticipantForOccupant:self.occupantId inRoom:self.mucContact.contactJid forAccountID:self.mucContact.accountID];
         if(participantInfo != nil)
             self.nick = participantInfo[@"room_nick"];
-        else
-            self.nick = self.occupantId;        //fallback (should never happen when receiving this notification)
     }
 }
 
--- a/Monal/Classes/MLFiletransfer.m
+++ b/Monal/Classes/MLFiletransfer.m
@@ -703,7 +703,7 @@ $$
         DDLogWarn(@"Failed to get url components, returning empty url!");
         return @"";
     }
-    urlComponents.fragment = @"";       //make sure we don't leak urlfragments to upload server
+    urlComponents.fragment = nil;       //make sure we don't leak urlfragments to upload server
     return urlComponents.string;
 }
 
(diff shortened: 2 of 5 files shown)


Monal/Classes/MLChannelContact.m
lissine0, line 100: This probably causes a similar bug to the one fixed in https://github.com/lissine0/Monal/commit/d9ea2d559988406207e3e33294960ec7f2f1a724

We should encode and decode the nick. Or if you don't like that, we should read the nick from the message_history table:
```sql
SELECT actual_from FROM message_history WHERE occupant_id=? AND buddy_name=? AND account_id=? ORDER BY message_history_id DESC LIMIT 1
```
But this second approach would be in the else case, because we still want to support the case of a participant that is online but didn't write any message.

tmolitor-stud-tu: Why encode/decode the nick?

I'm not sure if this case here also causes a bug: why should an instance of `MLChannelContact` be created in the first place, if the corresponding participant isn't part of the muc anymore?

Monal/Classes/MLFiletransfer.m
```diff
@@ -674,14 +674,7 @@ +(NSString*) retrieveCacheFilePathForUrl:(NSString*) url andMimeType:(NSString*)
         if([_fileManager fileExistsAtPath:cacheFilePath])
             return cacheFilePath;
     }
-    
-    //check for files having a different mime type but the same base url
```
tmolitor-stud-tu, line 678: It *think* this check was in there because some servers out there reported one mime-type when doing a HEAD request and another mime-type when doing the actual GET request.

tmolitor-stud-tu (approved):

Monal/Classes/MLChannelContact.m
lissine0: In my testing, this didn't also cause a bug.

I don't know when exactly the encode / decode methods run. But it feels wrong to read the nick from the `muc_participants` when decoding, because when creating an `MLChannelContact` object, you pass it the nick.
For example, here the `message.actualFrom` is passed to the MLChannelContact initializer as nick:
https://github.com/monal-im/Monal/blob/8d2ceb2d54d02154c0180d2d7282bc0cc8d76319/Monal/Classes/MLMessage.m#L361
And that gets called when displaying a message, even if the participant has left:
https://github.com/monal-im/Monal/blob/8d2ceb2d54d02154c0180d2d7282bc0cc8d76319/Monal/Classes/ChatView.swift#L999

PR #1620, opened 2026-04-17
6.4.20-b1

tmolitor-stud-tu:
- Now forcing SASL2+SCRAM for yax.im (already forced for conversations.im)
- Fixed display of messages containing a list of urls
- Fixed crash when receiving files with too long filenames
- Fixed rare case not connecting to a server after bringing the app from foreground to background too fast
- Fixed rare OMEMO devicelist publishing loop (thanks renegade)
- Fixed list of contacts in share sheet to not contain contacts only known from groupchats (thanks Kati)
- Fixed very rare OMEMO identity regeneration when receiving a push after device reboot but before entering the passcode for the first time
- Notes about new/deleted OMEMO devices of chat partners aren't displayed anymore for TOFU devices (no device explicitly trusted/untrusted by the user)


lissine0:
> Bump the minimum macOS version to 13.1 to match the minimum iOS version

That commit was only meant for the alpha branch, because there the minimum iOS version is 16.1
(feel free to ignore, as I don't think anyone is running old macOS versions)

tmolitor-stud-tu:
Oh you are right, well, I'll have to do a force push anyways, I'll remove it then.

PR #1625, opened 2026-04-21
Fix a bug + update SwiftPM dependencies

lissine0:
Note: among the ExyteChat changes, I removed this commit: https://github.com/monal-im/ExyteChat/commit/8c9ee4bb71b76ae03dbc44d12507840d95ece7ff

I benchmarked on the Simulator, and now opening the ChatView takes ~130ms before removing the commit and ~150ms after removing it. (My guess is the ~20ms is due to this change https://github.com/monal-im/ExyteChat/commit/cff61f6f66182ee0daf783f55b9272253f4e547c#diff-d124b557e3f767674d3d4d618a1e5fe4905f86bec60ede3acd1d59452c07f3cf but I'm going to keep it)

This means we can use the ExyteChat attachment upload status implementation.

Another note: now that I copied ZoomableContainer.swift to the ExyteChat fork, I made it public in ExyteChat and deleted it from Monal. That way if we ever need to modify it we'd do so in one place


lissine0:
Closing as the commits were cherry-picked

PR #1636, opened 2026-05-08
Add support for downloading, opening and uploading files in the new ChatView

lissine0:
Depends on https://github.com/monal-im/ExyteChat/pull/2

It adds a "Check File Size" button for `.none` downloadState, a "Download File" button for `.headers` downloadState, and a "Open File" button for `.complete` download state (but only for files that can't be displayed inline)

**A few implementation notes:**
- When opening files, the Apple framework relies on the file extension to determine the file type. Since our cached files have a hash as file extension, this unfortunately means that we have to make a temporary copy of a file when opening it.
The old chatView also does this same workaround: https://github.com/monal-im/Monal/blob/8cc133bd2a337818bdb99975fac40244eb2db81f/Monal/Classes/MLFileTransferFileViewController.m#L70-L71
- ~When a file is in `.none` and `.headers` downloadState, I want to show an icon representing the file. The QuickLookThumbnailing framework allows us to do so. But it requires a file URL. And since the file type detection happens entirely using the file extension (see the previous point), I decided to create empty files having the real file extension.
(I also do the same for .complete files that can't be displayed inline. We could generate a proper thumbnail for those. But except for the case of PDFs, the icon looks better than a thumbnail)~
_Update: decided against the icon idea when I redesigned the UI._

~**Note**: Since I'm displaying an icon representing the file before even doing a mimeType check, we may want to update MLContactCell of Active Chats so it wouldn't display a generic "📁 A File" when the file extension is that of an image, video or audio.~

**Potential improvements: (in the ExyteChat fork)**
- Maybe consider making the message bubbles that display files wider. Currently I use the same width as messages that contain a single image / video. The width can be changed from here: https://github.com/monal-im/ExyteChat/blob/10038d8b3239a1c1c4cc3a61527694e63de568cd/Sources/ExyteChat/Views/MessageView/MessageView.swift#L369-L372
- ~Maybe consider changing the way the checking / downloading `ProgressView()` is displayed in the FileView: by using an `if` instead of changing the opacity, the button would be perfectly centered by default, and would slide to the left to leave space for the ProgressView() when that's to be displayed. (currently there's reserved space for the progress view, so the "Check File Size" / "Download File" button is not perfectly centered)~
_Update: I redesigned the UI and this no longer applies._
- When improving the accessibility, we can create a UTType from the file extension or mimeType, and use the localized description for voice over: [UTType.localizedDescription](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype-swift.struct/localizeddescription)

### DIFF
--- a/Monal/Classes/ChatView.swift
+++ b/Monal/Classes/ChatView.swift
@@ -410,7 +410,7 @@ struct ChatView: View {
 
     var body: some View {
         ExyteChatView(messages: messages, chatType: .conversation, replyMode: .quote) { draft in
-            if !draft.medias.isEmpty || draft.recording != nil {
+            if !draft.medias.isEmpty || draft.recording != nil || !draft.files.isEmpty {
                 Task {
                     if draft.recording != nil {
                         await uploadAndSendFile(draft.recording!.url)
@@ -418,6 +418,9 @@ struct ChatView: View {
                     for media in draft.medias {
                         let localFileURL = await media.getURL()
                         await uploadAndSendFile(localFileURL)
+                    }
+                    for file in draft.files {
+                        await uploadAndSendFile(file.localURL)
                     }
                 }
             }
@@ -565,6 +568,12 @@ struct ChatView: View {
         })
         .showUsername(contact.isMuc)
         .linkPreviewsEnabled(false) //disabled for now due to https://github.com/exyte/Chat/issues/208
+        .checkFileSizeClosure { file in
+            MLFiletransfer.checkMimeTypeAndSize(forHistoryID: Int(file.id)! as NSNumber)
+        }
+        .downloadFileClosure { file in
+            MLFiletransfer.downloadFile(forHistoryID: Int(file.id)! as NSNumber)
+        }
         .enableLoadMore(offset: 10) {
             loadHistory()
         }
@@ -885,32 +894,8 @@ class ChatViewMessage: ExyteChat.Message {
     let fileInfo: ObservableKVOWrapper<MLFiletransferInfo>?
     private var subscriptions: Set<AnyCancellable> = Set()
     var text: String {
-        if innerMessage.messageType == kMessageTypeFiletransfer, let fileInfo = fileInfo {
-            switch(fileInfo.downloadState as DownloadState.RawValue) {
-                case DownloadState.complete.rawValue:
-                    let mimeType = fileInfo.mimeType as String
-                    if mimeType.starts(with: "audio/") || mimeType.starts(with: "image/") || mimeType.starts(with: "video/") {
-                        return ""
-                    } else {
-                        return """
-                            [File transfer with a file type that's not yet supported for displaying]
-                            \(innerMessage.obj.encrypted ? "" : "Link: \(innerMessage.messageText as String)")
-                        """
-                    }
-                case DownloadState.headers.rawValue:
-                    let humanReadableSize = (fileInfo.size as NSNumber).int64Value.formatted(.byteCount(style: .file))
-                    return """
-                    [File transfer; auto-downloading if the settings allow it...]
-                    Size: \(humanReadableSize)
-                    MimeType: \(fileInfo.mimeType as String)
-                    \(innerMessage.encrypted ? "" : "Link: \(fileInfo.downloadURL as String)")
-                    """
-                case DownloadState.none.rawValue:
-                    return "[File transfer; checking the size...]"
-                default:
-                    // .invalid case
-                    unreachable()
-            }
+        if innerMessage.messageType == kMessageTypeFiletransfer && fileInfo != nil {
+            return ""
         }
         return innerMessage.retracted ? NSLocalizedString("This message got retracted", comment: "") : innerMessage.messageText
     }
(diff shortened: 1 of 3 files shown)


tmolitor-stud-tu:
> When opening files, the Apple framework relies on the file extension to determine the file type. 

I'd always create a UTType from the decoded mime-type, at least on ios 17:

```
import QuickLook
import UniformTypeIdentifiers

let item = PreviewItem(
    url: fileURL,
    displayName: "Report",
    editingMode: .disabled,
    contentType: .pdf
)

let session = PreviewApplication.open(items: [item])
```

And for ios 16 I'd create a hardlink in the `MLFiletransfer calculateCacheFilePathForNewUrl:` directory prefixed with `tmp.` and having the correct file extension. Once the preview gets closed, remove the hardlink. In case that doesn't happen, our cleanup code will remove that temp file automatically after 3 dayes.

And once we go ios 17+, we can get rid of that whole branch of code entirely and solely use the modern quicklook interface :)

lissine0:
The `PreviewItem` in your code snippet is only available on visionOS, and neither it nor the iOS alternative (QLPreviewItem) have any initializer that takes a UTType.

The QuickLook framework relies entirely on the extension of the file name to detect the file type. You can't force it to use a given file type even if you use UIKit to customize the underlying view. (doing so only allows you to customize the display name, which by default is derived from the file name)

tmolitor-stud-tu:
Oh, my bad! then the ios 16 workaround is actually the any ios version solution.

lissine0:
~TODO: in the ExyteChat fork:
Display a generic icon for `.none` downloadState
And for `.headers` / `.complete`, generate a UTType from the mimeType, then use the UTType to get a file extension:~
```swift
if let utType = UTType(mimeType: file.mimeType) {
    fileExtension = utType.preferredFilenameExtension
}
```
~(we need the fileExtension in order to generate an icon, because QuickLook relies entirely on the fileExtension to determine the type of the file we gave it)~
_edit_: I decided against the icon idea when I redesigned the UI.

lissine0:
>> When opening files, the Apple framework relies on the file extension to determine the file type.

> I'd create a hardlink in the `MLFiletransfer calculateCacheFilePathForNewUrl:` directory prefixed with `tmp.` and having the correct file extension. Once the preview gets closed, remove the hardlink.

Why the requirement to use that specific directory?
I made the hardlink in the default tmp directory, and having the same name as the file (so the preview displays the correct file name):
https://github.com/monal-im/ExyteChat/blob/cbafa9268522afe80e0465f8d9a80eb3c1a9d897/Sources/ExyteChat/Views/Attachments/FileView.swift#L48-L63

Is the worry that the app might crash / get killed while the preview is open, and that the hardlink would remain and prevent the file from being deleted (assuming the user deletes the message containing the file)? ~Won't a symbolic link instead of hard link fix that issue?~
_edit_: I tested and previewing symbolic links doesn't work
_edit 2_: see the next comment

lissine0:
Interestingly, previewing hardlinks doesn't work on macOS (but it works on iOS).
That's the reason why the old chatview was copying instead of hardlinking.
We can either copy in all cases, or use preprocessor `#if targetEnvironment(macCatalyst)` to copy on macOS and hardlink on iOS.
The difference is only one line of code ([linkItem](https://developer.apple.com/documentation/foundation/filemanager/linkitem(at:to:)) vs [copyItem](https://developer.apple.com/documentation/foundation/filemanager/copyitem(at:to:)))

tmolitor-stud-tu:
I'm unsure why previewing hardlinks shouldn't work. Maybe it is some permission thing and creating the hardlink in the shared app group directory fixes things.

The `tmp.xxx` name is meant to be automatically cleaned up by `doStartupCleanup` in `MLFiletransfer.m`.
Of course you can also create a directory following the `tmp.xxx` scheme in that cache dir (`[[HelperTools getContainerURLForPathComponents:@[@"documentCache"]] path]`) and then put the file with the correct name inside (make sure to not create directory traversal security issues).

lissine0:
I implemented your suggested change. Here is the patch for you to review: https://github.com/monal-im/ExyteChat/commit/87c98ab31776449f583f184804477e0190b5c1df
I'm putting the temporary hardlinks inside a directory called `tmp.filePreviews` under GroupContainer/`documentCache`

Note that in the normal case, my code in ExyteChat deletes the hardlink when the user closes the preview.
The automatic cleanup is useful in case the app crashes / gets force-closed while the preview is open.

Moreover, note that the `doStartupCleanup` method in Monal checks the modified date and compares it to 1 day ago before deleting.
And since `tmp.filePreviews` is a directory, its modified date will keep updating each time the user previews a file.
This means, if the user force-closes the app while previewing the file, depending on how regularly the user previews files it may take a long time for the hardlink to get deleted.


Furthermore, I dislike the way I'm detecting the `appGroupID` in ExyteChat. It works, but we could do better: store the hardlinks in a specific directory under /tmp, and update Monal's `doStartupCleanup` method to always delete that directory on start up. That would remove the need to detect `appGroupID` and solve the 1-day modified date problem at the same time. What do you think?

<details>
  <summary>side-notes</summary>

> I'm unsure why previewing hardlinks shouldn't work. Maybe it is some permission thing and creating the hardlink in the shared app group directory fixes things.

Yes, indeed. Previews work on macOS when the hardlink is in the app's group container.

Another side-note: I chose hardlinking, but there's not much of a difference between copying and hardlinking in our case, due to the copy-on-write feature in APFS: https://developer.apple.com/forums/thread/717836#733068022:~:text=copy%2Don%2Dwrite
</details>

PR #1641, opened 2026-05-15
Backports to prepare for v6.4.21

lissine0:
- Show last usage timestamp of OMEMO keys
IOS_ONLY - Use the iOS 18 UI in the share sheet
IOS_ONLY - Allow using any language in the Display Name field in account settings
- Try to improve location sharing accuracy
- Improve contact last activity timestamps
- Fix several bugs


lissine0:
RE: "Remove yax.im as one of the default servers for new accounts", maybe we should ask the yax.im server operator to switch to the mod_csi_battery_saver module instead?

PR #1682, opened 2026-08-11
Delete unused bytesToHuman

matthewrfennell:
This was added in 7f96bbba7c69560f7808b92c2b4c44364b9c5663 to print user-friendly file sizes in the logs. It only ever had one caller, which was removed in 3081ada7c617341902609fe303ed8e58d522226e.

As well as being dead code, I want to delete this to remove the suspicious implicit cast when converting bytes from `int64_t` to `double`.


matthewrfennell:
I was too eager, I missed a Swift call `bytes(toHuman: ...)` :(
I didn't catch it because compilation fails anyway on my xcode because of `CLANG_WARN_SUSPICIOUS_IMPLICIT_CONVERSION`. I should have compiled with that flag off too.

Issue #1685, opened 2026-08-14
[Feature]: Switch to SwiftPM for all dependencies that we currently get using CocoaPods


lissine0:
The main CocoaPods spec repo will become read-only starting from December 2026: https://blog.cocoapods.org/CocoaPods-Specs-Repo/

We should switch to using SwiftPM for the dependencies that we currently get via CocoaPods, in order to keep receiving updates.

See https://github.com/monal-im/Monal/blob/develop/Monal/Podfile

PR #1687, opened 2026-08-14
Start migrating dependencies to SwiftPM

matthewrfennell:
This PR is to start addressing #1685

I'll gradually add more commits to the branch as I migrate more dependencies. But if the branch looks good, it can be merged at any time.

### Old-UI-only

* SDWebImage: not needed with new UI
* DZNEmptyDataSet: not needed with new UI
* MBProgressHUD: not needed with new UI

## Testing

These screenshots/recordings are taken on a clean build after migrating the respective packages to SwiftPM.

### NotificationBannerSwift

NotificationBannerSwift-1
NotificationBannerSwift-2

### FLAnimatedImage

https://github.com/user-attachments/assets/168bc60b-099c-4add-b18b-c9c5501cd02a

### ASN1Decoder

On both `develop` and this branch, I changed `channelBindingDataForType:` to always return `channelBindingData_TLSServerEndPoint` (to force this function to get called without messing with my server's SASL configuration), saw that an `ASN1 decoded cert:` log line was printed, logged `channelBindingData` and saw the hash when connecting to my server was the same on both branches.

### KSCrash

KSCrash-1

### sqlite3

I first did a smoke test to check that the database connection worked at all. This screenshot is taken after sending the messages, closing the app and reopening:

sqlite3-1

I also dumped the configuration from CocoaPods vs SwiftPM, by iterating over `sqlite3_compileoption_get()`. This showed this diff:

```diff
--- CocoaPods
+++ SwiftPM
-ENABLE_LOCKING_STYLE=0
+HAVE_ISNAN
```

All other compile options were the same. I did some digging and found that `ENABLE_LOCKING_STYLE` was added in the CocoaPod to work around a bug in a specific version, which then got reported upstream (clemensg/sqlite3pod#1). They were seeing a compile error at the time, which we don't get when removing the override now.

Meanwhile, `HAVE_ISNAN` means that we use the system `isnan` now instead of SQLite's one. I'm neither expecting iOS to be such a bespoke platform that they do anything wacky here, or that we rely on obscure NaN behaviour, so this should be fine.

Finally, I tested on both iOS 26.3 and 16.1, and saw that all the compile options remain the same (in other words: we are certainly using the SwiftPM version, it's not falling through to the system SQLite somehow).


tmolitor-stud-tu:
Afaik the `ASN1Decoder` is only used for tls-server-end-point channel-binding, not for omemo.

tmolitor-stud-tu:
> PromiseKit: AnyPromise https://github.com/mxcl/PromiseKit/pull/1193

does this mean as soon as we switch to SPM for promise kit, we'll loose promises for ObjC?

matthewrfennell:
> Afaik the ASN1Decoder is only used for tls-server-end-point channel-binding, not for omemo.

Quite right! I kind of assumed it was used throughout `MLCrypto`. I just updated the verification to test that.

> does this mean as soon as we switch to SPM for promise kit, we'll loose promises for ObjC?

Yeah, that's right. `AnyPromise` becomes unavailable, which is our only option in ObjC.

Thinking aloud, some options are:

* Use native swift concurrency features in SwiftUI, passing callbacks to ObjC
* Use a different library like [Google Promises](https://github.com/google/promises)
* Switch the cocoapod to resolve via git (if PromiseKit even continue to support cocoapods afterwards)
* Rebase/maintain an [unmaintained branch that added SwiftPM support](https://github.com/mxcl/PromiseKit/pull/1193) (I tried to test with this fork, and couldn't get it working)
* Vendor the parts we use into Monal
* Work with upstream to find a less intrusive way to support it there

Personally, I prefer to use swift's native concurrency features, if we can get it to work. On a quick glance, there seems to be [quite good interoperability with ObjC callbacks now](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0297-concurrency-objc.md), so if we can get it to work while dropping a dependency and avoiding too much custom code, it'd be ideal.

tmolitor-stud-tu:
To be honest, it was a huge step forward to migrate off of the ObjC callbacks and use promises in our codebase. Not just for ObjC <--> Swiftui interactions, but from a clean readable objc code perspective.

So using native swift concurrency features isn't the way forward imho. And I'm quite sure not every ObjC promise usage is solely a Swiftui interaction.

lissine0:
> Yeah, that's right. `AnyPromise` becomes unavailable, which is our only option in ObjC.
> 
> Thinking aloud, some options are:
> 

Another option is to use a different package manager such as Carthage. In fact it is one of the recommended installation methods of PromiseKit: https://github.com/mxcl/PromiseKit#quick-start

tmolitor-stud-tu:
Honestly I don't understand why SwiftPM support for ObjC promises *and* Swift promises at the same time is that complicated given that cocapods allows this already (showing that its doable somehow). Maybe splitting up the codebase into a Swift and ObjC one and packaging both as separate SwiftPM packages would be enough?

lissine0:
That's more or less the idea of the aforementioned unmerged PromiseKit PR: https://github.com/mxcl/PromiseKit/pull/1193/changes/1e2e1177a3d2d3ff4502ca5cbb760aeadd4514dc

> Split up the PromiseKit package into two targets. One for pure Swift and another for Objective-C
>
>Targets must be Swift or Objective-C only. Mixing languages in one target is not supported by SPM.
Both targets get integrated into the PromiseKit library product

tmolitor-stud-tu:
Hmmm okay, I understand the problem: the codebase is too mixed for this split to work cleanly, right?
Then maybe carthage is the better choice (or accio?). what about sqlite, does that support carthage/accio, too? that would probably solve both problems at once :)

tmolitor-stud-tu:
@matthewrfennell If you migrated KSCrash to SPM, you can try to update it to KSCrash 2.0 by merging my branch and updating it to SPM.

lissine0:
> Then maybe carthage is the better choice (or accio?)

Accio is based on SwiftPM, so it should have the same limitation.
We could either switch to Carthage, or keep using CocoaPods but make it resolve via git, like Matthew suggested earlier, assuming PromiseKit keeps maintaining their podspec.

> what about sqlite, does that support carthage/accio, too? that would probably solve both problems at once :)

Actually we can use sqlite with SwiftPM: https://github.com/clemensg/sqlite3pod/issues/25#issuecomment-3094630899

tmolitor-stud-tu:
Well, since we don't know if promisekit will continue to support cocoapods, I think we'd better use this opportunity to switch to carthage now.

tmolitor-stud-tu:
Why are the test failing?

matthewrfennell:
I haven't had time to investigate yet unfortunately. I'll take a proper look in the next couple of days, re-running with/without various commits.

It was since the most recent change, which bumped the cocoapods version and migrated the sqlite pod. It passes on my laptop, but consistently fails on CI.

tmolitor-stud-tu:
maybe we just have to bump the macos/xcode version of the ci :)

matthewrfennell:
Directly above is a pure rebase

matthewrfennell:
The issue was that the `swift-tools-version` in `sqlite3-package` was too high. When I created the package, my XCode just used the latest `swfit-tools-version` available to it, which was later than the one available to the CI runner.

monal-im/sqlite3-package@4174fb7e23124259f2d6874eb61b301d944421ad has fixed it by dropping the `swift-tools-version` from 6.3 to 6.1 (in sync with the upstream CSQLite package - so anyone who can depend on that one can depend on this one too)

PR #1692, opened 2026-08-17
Unpin the WebRTC version

lissine0:
The issue introduced in WebRTC 141 has since been fixed.
This is a partial revert of commit 4e1c091fe7321c24


tmolitor-stud-tu:
Great! :)

PR #1693, opened 2026-08-17
Persist the "System Sound" notification sound setting properly

### LINKED ISSUE #1691: [Bug]: Notification Sound
**Via email:**

Hello, 
Just wanted to ask if I’m doing something wrong but every time I change notification sound to default system notification sound, it gets reset after a short while. 
What’s interesting, it only happens when I change it to system sound. If I choose one of those provided by Monal, it’s okay. 

I’m using it on iPhone 17e with latest version of iOS 26 and latest stable version of Monal from App Store. 

Sent from my iPhone

lissine0:
The previous approach used nil to represent "System Sound"
However, since this setting is stored in defaultsDB, on next Monal startup it is given the default value (in `MLXMPPManager defaultSettings`)

Fixes #1691

### DIFF
--- a/Monal/Classes/MLConstants.h
+++ b/Monal/Classes/MLConstants.h
@@ -213,6 +213,8 @@ static inline NSString* _Nonnull LocalizationNotNeeded(NSString* _Nonnull s)
 #define kMucActionShowProfile @"profile"
 #define kMucActionReinvite @"reinvite"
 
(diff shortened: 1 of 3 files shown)


tmolitor-stud-tu:
can you resolve the conflicts? I'll merge it afterwards, thanks for fixing this :)

lissine0:
done :)

PR #1703, opened 2026-09-02
Make remaining suspicious implicit casts explicit

matthewrfennell:
I ideally wanted to remove the need for these completely, but these are the more difficult remaining ones which I don't have a good solution to.

For instance, `(double)operationCount` to do calculations is a common pattern, and not a problem at the depths we see on the `parseQueue`. Converting through `NSNumber*` could remove the casts, but I'm not convinced that's actually any safer, and would lead to unnecessarily verbose code.

### DIFF
--- a/Monal/Classes/MLDNSLookup.m
+++ b/Monal/Classes/MLDNSLookup.m
@@ -111,7 +111,7 @@ static NSMutableDictionary* _RRCache;
 
             struct timeval tv;
             tv.tv_sec  = (time_t)remainingTime;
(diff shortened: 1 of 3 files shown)


tmolitor-stud-tu:
Coming from a C background, I don't think casts are inherently bad, but simply a normal tool to get something from int to double or vice versa. So your implicit casts are absolutely right and yes, packing everything into an NSNumber would just make things slower and harder to read, but give us no benefit.

PR #1706, opened 2026-09-06
Bump KSCrash to 2.6.0

matthewrfennell:
This brings us fully up-to-date with upstream.

https://github.com/user-attachments/assets/3d6f9579-ad92-4f0c-9042-da01b9a991eb

I initially only bumped up to `2.5.1` as I thought that the crash handler was broken beyond there. It turns out this was because I was testing it wrong :)

KScrash recently updated to not trigger crash reports on `SIGINT`. When testing initially, I would run from XCode, trigger a crash, see it get suspended in XCode and then re-run from there. It turns out that this sends `SIGINT` under the hood, so in `2.6.0`, the library was deliberately ignoring these signals from XCode.

When testing without XCode, by running the compiled target directly (as in the video), I was able to observe the correct behaviour in clean exit and crash scenarios.

### DIFF
--- a/Monal/Classes/HelperTools.h
+++ b/Monal/Classes/HelperTools.h
@@ -80,7 +80,7 @@ void swizzle(Class c, SEL orig, SEL new);
 +(NSData* _Nullable) convertLogmessageToJsonData:(DDLogMessage*) logMessage counter:(uint64_t*) counter andError:(NSError** _Nullable) error;
 +(void) initSystem;
 +(void) installExceptionHandler;
-+(int) pendingCrashreportCount;
++(NSInteger) pendingCrashreportCount;
 +(void) flushLogsWithTimeout:(double) timeout;
 +(BOOL) isAppSuspended;
 +(void) signalSuspension;
--- a/Monal/Classes/HelperTools.m
+++ b/Monal/Classes/HelperTools.m
@@ -2436,7 +2436,7 @@ static void notification_center_logging(CFNotificationCenterRef center, void* ob
     DDLogVerbose(@"KSCrash installing handler with callback: %p", crash_callback);
     KSCrashConfiguration* config = [KSCrashConfiguration new];
     config.installPath = [[HelperTools getContainerURLForPathComponents:@[@"CrashReports"]] path];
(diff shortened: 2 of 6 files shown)


tmolitor-stud-tu:
Looks good, please do a rebase.

I'm wondering, how did you solve the issue I described here: https://github.com/kstenerud/KSCrash/pull/615 ?

matthewrfennell:
I didn't run into that issue, so I'll take a proper look later today or tomorrow to understand why / what the difference is.

matthewrfennell:
It seems that CocoaPods shares the pod target with the same build settings between the main app and NSE, while SwiftPM supports building it with separate compile flags for the different targets. See this section of this [blog post](https://miqu.me/blog/2016/11/28/app-extensions-xcode-and-cocoapods-omg/):

> The problem comes with the integration with CocoaPods, as they (correctly) deduplicate targets. This means that you’ll compile the library once (say without unavailable API) and link it to your targets. But sometimes you want to compile with usage of unavailable APIs against your main app, and removing unavailable usage on the extension. If you want this, you’re out of luck as it’s not directly supported by Cocoapods.

When compiling code for the app extension, you want `Require Only App-Extension API` to be `Yes` to ensure you don't make any calls to forbidden code that gets you rejected from the app store.

So, CocoaPods sets it by default in the generated KSCrash target. But then, this version with the restrictions in place gets linked to both the app extension and main app. Hence the build failure below:

kscrash-cocoapods-bad

(Since CocoaPods leaves this unset for `Release` builds, I guess that Apple doesn't enforce this build flag in app store review, only that you don't call any forbidden methods from the app extension. So we might have been able to just override that and not use any forbidden methods. But that's just conjecture.)

Anyway, if I force the setting to be `No`, it compiles fine with CocoaPods:

kscrash-cocoapods-good

Meanwhile, in SwiftPM it must compile separately for each, respecting the build settings of each target separately:

kscrash-swiftpm-monal
kscrash-swiftpm-nse

If I set `Require Only App-Extension API` to be `Yes` in the Monal target, it fails again:

kscrash-swiftpm-forced-failure

tmolitor-stud-tu:
Ah, that explains it. Luckyly we are migrating to SPM so I can close that open issue in my in-mind todo list for good, great! :)

Thanks for digging so deep into it!

PR #1708, opened 2026-09-07
Use static properties to access resources

matthewrfennell:
This gives us compile-time checks to ensure we don't misspell any of the `Image`s or `Color`s by accident.

Some spot-checks that everything still looks the same:

assets-1
assets-2
assets-3
assets-4
assets-5
assets-6
assets-7
assets-8

The tests can be re-run, and should pass (as [here](https://github.com/matthewrfennell/Monal/actions/runs/34155803189)) once #1704 is merged.


tmolitor-stud-tu (approved):
Good idea :)

PR #1710, opened 2026-09-12
Move signalDeps to SPM

matthewrfennell:
I tested this more extensively in monal-im/SignalProtocol-ObjC#1, but as a sanity check, made sure we could still send encrypted messages:

signalDeps-1


tmolitor-stud-tu:
can you solve the conflicts?

PR #1711, opened 2026-09-15
Bump the minimum iOS version to 17.0

lissine0:
Maintaining iOS 16 support is starting to become too annoying, as ExyteChat and its dependencies keep using more iOS17+ features.
For example, they recently added support for location sharing / displaying, and it mostly relies on iOS 17+. There's no "simple" iOS 16 alternative in this case. We would have to rewrite the code to maintain iOS 16 support.
So I decided to bump the minimum version instead.


tmolitor-stud-tu:
what about just deactivating features that depend on ios 17 on devices running ios 16?

lissine0:
> what about just deactivating features that depend on ios 17 on devices running ios 16?

Yes we can do that, but we'd still be using older versions of exyte/mediaPicker and exyte/AnchoredPopup.
Most notably, this means we won't have the fix for this bug, which can be triggered often on macOS:
https://github.com/exyte/Chat/pull/278#issuecomment-5326528639

I'm in favor of removing iOS 16 support. We were just delaying the inevitable. Xcode doesn't even have the option to install the iOS 16 simulator runtime anymore, at least as of Xcode 26.6

Furthermore, the most recent device that's stuck on iOS 16 is the iPhone 8. It was released 9 years ago and stopped being sold 6.5 years ago. In 6 months Apple will consider it ["obsolete"](https://support.apple.com/en-us/102772) and will stop providing repairs and battery replacements for it.

tmolitor-stud-tu:
okay, I'm convinced...

PR #1717, opened 2026-09-21
Resolve PromiseKit via git

matthewrfennell:
This will let us easily bump the version if we are still dependent on CocoaPods once CocoaPods trunk becomes read-only.


tmolitor-stud-tu:
Didn't we agree to switch to carthage for promisekit?
https://github.com/monal-im/Monal/pull/1687#issuecomment-5310307421

tmolitor-stud-tu:
Hmm, I just read through the documentation of Google Promises and it seems to provide almist the same features as Promisekit, but has ObjC still as first class citizen.

So when carthage turns out to not work, we should maybe simply switch to Google Promises.

matthewrfennell:
> Didn't we agree to switch to carthage for promisekit?

Ah since then, mxcl (one of the maintainers of PromiseKit) pointed out that the mixed-module problem will be [fixed soon in Swift](https://github.com/mxcl/PromiseKit/issues/1361#issuecomment-5412431451) (the swift enhancement merged a few weeks ago). So, as long as that works, we should be able to swap to SwiftPM and keep using PromiseKit.

tmolitor-stud-tu:
That would be the ideal solution :)
I wasn't even aware you reached out to the project.

